PR #12, opened 2022-05-12
Fix Author UI Padding Issue

eric-ampire:
Before 2
Before

### DIFF
--- a/feature-author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorScreen.kt
+++ b/feature-author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorScreen.kt
@@ -157,8 +157,7 @@ private fun AuthorHeader(author: Author) {
             modifier = Modifier
                 .size(216.dp)
                 .align(Alignment.CenterHorizontally)
-                .clip(CircleShape)
-                .padding(bottom = 12.dp),
+                .clip(CircleShape),
             contentScale = ContentScale.Crop,
             model = author.imageUrl,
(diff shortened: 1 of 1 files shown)


feature-author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorScreen.kt
```diff
@@ -157,8 +157,7 @@ private fun AuthorHeader(author: Author) {
             modifier = Modifier
                 .size(216.dp)
                 .align(Alignment.CenterHorizontally)
-                .clip(CircleShape)
-                .padding(bottom = 12.dp),
```
surajsau, line 161: I think the `.padding(bottom = 12.dp)` is intended. But since with a fixed `.size()`,  `.padding()` will just resize the contents (including any `Modifier` change) following its declaration with -ve padding offsets, placing `.padding()` before `.size()` should solve the issue.

```suggest
AsyncImage(
            modifier = Modifier
                .padding(bottom = 12.dp)
                .size(216.dp)
                .align(Alignment.CenterHorizontally)
                .clip(CircleShape),
            ..
        )
```

<table>
<tr><td>Before</td><td>After</td></tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

eric-ampire: I see thanks, I've applied what you said by changing the modifier placement

JolandaVerhoef (approved):

PR #16, opened 2022-05-12
Reuse the same RoundedCornerShape for NewsResourceCardExpanded and NewsResourceHeaderImage

### LINKED ISSUE #15: [Bug]: `NewsResourceCardExpanded` and `NewsResourceHeaderImage` do not share the same `RoundedCornerShape`

### What happened?

In the following screenshot, we can see the card bacground on the top-left and top-right corners "leaking" behind the header image.

Screenshot_20220512_232114

### Relevant logcat output

_No response_

### Code of Conduct

SimonMarquis:
Fixes #15

| Before | After | Diff |
|---|---|---|
| Screenshot_20220512_232930 | Screenshot_20220512_232950 | 20220512-233019-diff |

### DIFF
--- a/core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
+++ b/core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
@@ -141,7 +141,6 @@ fun NewsResourceHeaderImage(
         },
         modifier = Modifier
             .fillMaxWidth()
-            .clip(RoundedCornerShape(topEnd = 16.dp, topStart = 16.dp))
             .height(180.dp),
         contentScale = ContentScale.Crop,
         model = headerImageUrl,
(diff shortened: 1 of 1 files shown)


core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -141,7 +141,7 @@ fun NewsResourceHeaderImage(
         },
         modifier = Modifier
             .fillMaxWidth()
-            .clip(RoundedCornerShape(topEnd = 24.dp, topStart = 24.dp))
+            .clip(RoundedCornerShape(topEnd = 16.dp, topStart = 16.dp))
```
calren, line 144: what do you think about extracting this value out as a variable (either in this file directly, or in dimens.xml) so that it's less error prone if we decide to update it later?

calren:
thanks for spotting the bug and putting in a fix! added a question

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
SimonMarquis: Clip is no longer necessary on the child element.

calren (approved):
thanks for the changes! lgtm

PR #19, opened 2022-05-12
Move `AuthorsCarousel`'s padding into `AuthorItem`

### LINKED ISSUE #18: [Bug]: AuthorItem pressed effect feels too narrow

### What happened?

Screenshot_20220513_001456

### Relevant logcat output

_No response_

### Code of Conduct

SimonMarquis:
Fixes #18

<details>
<summary>Outdated screenshots diff</summary>

| Before | After | Diff |
|---|---|---|
| Screenshot_20220513_001456 | Screenshot_20220513_003041 | 20220513-003112-diff |

</details>


JolandaVerhoef:
Instead of using padding inside the child elements, can you lift the paddings to the LazyRow?

I checked the original designs, and the padding needs to be 24dp:

```
    LazyRow(
        modifier = modifier,
        contentPadding = PaddingValues(horizontal = 24.dp),
        horizontalArrangement = Arrangement.spacedBy(24.dp)
    ) {
```

While you're at it, could you also add a 24dp top padding to the `AuthorItem` inside the `ForYouScreen`?

SimonMarquis:
> Instead of using padding inside the child elements, can you lift the paddings to the LazyRow?

Moving the padding up to the `LazyRow` will bring back the main issue which is that the bounds of the selection (pressed state) is very narrow. Is this what's really expected?

JolandaVerhoef:
Good point! I hadn't realized that.

Another alternative would be to make the ripple unbounded, as there's no obvious container for the `AuthorItem`. You could do that by adding an `interactionSource` and `indication` to the `toggleable` modifier. 

Currently I'd be open to both options, with a slight preference for the unbounded ripple version. What's your preference? 

I'll make sure to verify with our internal UX designers as well to see what their final verdict is, but in the meantime let's already choose one of the options.

SimonMarquis:
Here is what it looks like with unbounded ripple, which I prefer over the rectangular effect.
I'm still not familiar enough with Compose and was looking for the equivalent of `?attr/selectableItemBackgroundBorderless`.

Can you confirm that the layout dimensions are respecting your UI specs? I'm not sure if we need to restore the `Modifier.padding(8.dp)` back in the `AuthorItem`'s modifier.

feature-foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/AuthorsCarousel.kt
```diff
@@ -88,14 +95,19 @@ fun AuthorItem(
...
             .sizeIn(maxWidth = 48.dp)
             .semantics(mergeDescendants = true) {
                 stateDescription = "$followDescription ${author.name}"
             }
     ) {
-        Box(modifier = Modifier.fillMaxWidth()) {
+        Box(
+            modifier = Modifier.fillMaxWidth(),
+            contentAlignment = Alignment.Center,
+        ) {
```
JolandaVerhoef, line 110: Would you still need this change?

JolandaVerhoef:
This looks good! I added one small comment. Also, for future reference, can you update the PR description with the new "after" screenshot and diff?

feature-foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/AuthorsCarousel.kt
SimonMarquis: No, I'll remove it.

SimonMarquis:
Removing `Column`'s `8.dp` padding shifted the carousel up a bit. Is this alright?

| Before | After | Diff |
|---|---|---|
| Screenshot_20220521_000337 | Screenshot_20220521_000402 | 20220521-000445-diff |

JolandaVerhoef:
Hmm I checked the design and there should be 24dp padding above and below the author items. I think you can change the LazyRow's contentPadding to be both horizontal and vertical and that would match the design.

PR #26, opened 2022-05-13
Update benchmarking & baseline profiles

keyboardsurfer:
Uses newest versions from I/O release
Also updates baseline profile to match latest changes


JoseAlcerreca (changes requested):
Please do not merge yet

dturner (approved):

PR #32, opened 2022-05-14
Replace AndroidManifest's package with AGP's namespace property

SimonMarquis:
AGP 8 will no longer support `AndroidManifest.xml`'s `package` attribute.

Docs: https://developer.android.com/studio/build/configure-app-module#set-namespace


JolandaVerhoef:
Let's hold off with this change for a bit, pending the discussion in https://issuetracker.google.com/u/0/issues/217414933

SimonMarquis:
Sure, I've been bitten by that as well on other projects.
So let's wait for now 👍

JolandaVerhoef:
Closing as discussed

Issue #34, opened 2022-05-14
The `libs` error in the some build.gradle


SaeedNoshadi89:
I see an error with `libs` in some `build.gradle` files.

The error is obvious below files:
`build.grdle.kts (:code-database)`
`build.grdle.kts (:code-datastore)`
`build.grdle.kts (:code-model)`
`build.grdle.kts (:code-navigation)`

![Screen Shot 2022-05-14 at 17 04 39](https://user-images.githubusercontent.com/50018300/168426259-c79d5dbe-d4b8-441d-b7a2-c4b362c7b93e.png)

alexvanyo:
Suppressed this warning in #35

PR #35, opened 2022-05-14
Add suppress annotation to fix libs error in the some build.gradle.kts

SaeedNoshadi89:
This is related to #34.
I used suppress annotation to fix the error red line in the some gradle files `libs` code.

### DIFF
--- a/core-database/build.gradle.kts
+++ b/core-database/build.gradle.kts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@kotlin.Suppress("DSL_SCOPE_VIOLATION")
 plugins {
     id("nowinandroid.android.library")
     id("nowinandroid.android.library.jacoco")
--- a/core-database/build.gradle.kts
+++ b/core-database/build.gradle.kts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+// TODO: Remove once https://youtrack.jetbrains.com/issue/KTIJ-19369 is fixed
 @Suppress("DSL_SCOPE_VIOLATION")
 plugins {
     id("nowinandroid.android.library")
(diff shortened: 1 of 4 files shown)


core-database/build.gradle.kts
```diff
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@kotlin.Suppress("DSL_SCOPE_VIOLATION")
```
alexvanyo, line 16: Small nit: On my side, it looks like the `kotlin.` is unnecessary, I think just the plain `Suppress` is clear enough!
```suggestion
@Suppress("DSL_SCOPE_VIOLATION")
```

core-datastore/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ import com.google.protobuf.gradle.generateProtoTasks
 import com.google.protobuf.gradle.protobuf
 import com.google.protobuf.gradle.protoc
 
+@kotlin.Suppress("DSL_SCOPE_VIOLATION")
```
alexvanyo, line 22: Could we add a TODO with the upstream issue to track this? I believe this is the correct one: https://youtrack.jetbrains.com/issue/KTIJ-19369

Something like
```suggestion
// TODO: Remove once https://youtrack.jetbrains.com/issue/KTIJ-19369 is fixed
@kotlin.Suppress("DSL_SCOPE_VIOLATION")
```

alexvanyo (commented):
Thank you, this looks great! Glad to be able to avoid those annoying squiggly red lines. Just two minor comments from me!

core-datastore/build.gradle.kts
SaeedNoshadi89: It seems great. Thank you

alexvanyo (approved):
Sorry for the delay, this looks great!

PR #44, opened 2022-05-17
Fix themed app icon launcher roundIcon

serbelga:
Add `monochrome` attribute to `ic_launcher_round.xml` in :app and :app-nia-catalog

| **Before** | **After** |
|--|--|
|  |  |


ricknout:
Yes I think we can! I've attached an SVG which is essentially the negative (cutout) of the NiA logo used for the foreground. It will need to be placed in `res/drawable-v24` because it uses `android:fillType="nonZero"` when imported into Android Studio, but that's okay seeing as this only affects API 31+. It will also still need a solid background drawable in `res/drawable-v24` that's the opposite color of the normal one. Please try it out and LMK if it works :)

ic_foreground

serbelga:
It's ok, but It's a bit misaligned

ricknout:
Ah, apologies. Please try this one instead:
ic_foreground

ricknout (approved):
Thanks!

PR #51, opened 2022-05-19
Replaced -Xopt-in to -opt-in because deprecated

NUmeroAndDev:
I want to resolve ↓ warning.

>w: '-Xopt-in' is deprecated and will be removed in a future release, please use -opt-in instead


lmorda (approved):

alexvanyo (approved):
Thanks for catching this and the contribution!

Issue #59, opened 2022-05-23
[Bug]: Align the bottom bar items to center in the landscape mode


SaeedNoshadi89:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

A bug happened when the device is on the landscape mode, the bottom bar items are moved to the top.

![alt text](https://user-images.githubusercontent.com/50018300/169873515-7e9d3df1-bada-4043-b96e-dead352f2ad7.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #60, opened 2022-05-23
fix the bottom bar items alignment

SaeedNoshadi89:
This is related to #59 

device-2022-05-24-013314

### DIFF
--- a/app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
+++ b/app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
@@ -18,6 +18,7 @@ package com.google.samples.apps.nowinandroid.ui
 
 import androidx.compose.foundation.layout.ExperimentalLayoutApi
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.WindowInsetsSides
 import androidx.compose.foundation.layout.consumedWindowInsets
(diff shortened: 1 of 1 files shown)


SaeedNoshadi89:
@JolandaVerhoef can this PR to review?

JolandaVerhoef:
The design defines that the Navigation Bar and Rail should behave as defined by default by the Material 3 component. We do not want to adapt it's behavior, so if there's anything wrong with it we should raise a bug with the library. 

In this case, our design actually tells us that in Landscape mode, the app should still use the Navigation Bar and not switch to the Navigation Rail. So the actual solution to the issue you raised might be to do a better check on landscape / portrait.

RivanParmar:
But the Material 3 documentation even states that the Navigation Bar should not be used for less than 3 destination instead we should use tabs. As shown [here](https://m3.material.io/components/navigation-bar/guidelines).

RivanParmar:
Also I would like to work on it.

JolandaVerhoef:
Good point, and you're right about the fact that the Navigation Bar should not be used for less than 3 destinations. However, we're planning to add more destinations in this bar, which is why we already opted to use a Navigation Bar although that does not follow the recommendations. 

A Navigation Bar is always horizontal. A Navigation Rail is vertical. In our designs, the landscape version of the app uses the same horizontal Navigation Bar as the portrait version. 

Feel free to re-open this Pull Request if you want to continue working on fixing the landscape version of the app, or open a new Pull Request with the changes instead!

RivanParmar:
Ok, Thanks!

RivanParmar:
@JolandaVerhoef I have created a new pull request #68 which fixes this issue and uses the same Navigation Bar in landscape mode. Please review it.

PR #61, opened 2022-05-23
Update baseline profile and generation logic

keyboardsurfer:


app/build.gradle.kts
```diff
@@ -44,18 +44,21 @@ android {
...
             applicationIdSuffix = ".staging"
         }
+        val benchmark by creating {
+            initWith(staging) // Usually should be `initWith(release)`. Connecting to demo backend.
+            matchingFallbacks.add("debug") // Making some settings below to align closer to release.
+            signingConfig = signingConfigs.getByName("debug")
+            proguardFiles("benchmark-rules.pro") // Only use benchmark proguard rules
+            isMinifyEnabled = false //  FIXME enabling minification breaks access to demo backend.
+            isDebuggable = false
+            applicationIdSuffix = "" // reset to base application id suffix
```
mlykotom, line 60: We shouldn't remove the suffix, because it would clash with the main build variant. I'd say we can have `.benchmark`.
```suggestion
            applicationIdSuffix = ".benchmark"
```

app/build.gradle.kts
```diff
@@ -44,18 +44,21 @@ android {
...
             initWith(debug)
             signingConfig = signingConfigs.getByName("debug")
             matchingFallbacks.add("debug")
             applicationIdSuffix = ".staging"
         }
+        val benchmark by creating {
+            initWith(staging) // Usually should be `initWith(release)`. Connecting to demo backend.
+            matchingFallbacks.add("debug") // Making some settings below to align closer to release.
+            signingConfig = signingConfigs.getByName("debug")
+            proguardFiles("benchmark-rules.pro") // Only use benchmark proguard rules
```
mlykotom, line 57: We want to use both app's rules + benchmark rules (that has just -dontobfuscate). That's probably why isMinifyEnabled = true doesn't work

```suggestion
            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro", "benchmark-rules.pro")
```

app/build.gradle.kts
keyboardsurfer: Done

app/build.gradle.kts
keyboardsurfer: Minification isn't the problem, but as soon as `minifyEnabled true` the connection to the data breaks.

benchmark/src/main/java/com/google/samples/apps/nowinandroid/baselineprofile/BaselineProfileGenerator.kt
```diff
@@ -45,6 +46,12 @@ class BaselineProfileGenerator {
                 findObject(By.text("Interests"))
                     .click()
                 waitForIdle()
+                findObject(By.text("Accessibility")).scroll(Direction.DOWN, 2000f)
```
mlykotom, line 49: One more thing, the 2000% look weird. I had troubles with `scroll` because it weirdly scrolls not enough .. I was using `fling` which more reliably scrolls the content.

mlykotom (approved):

benchmark/src/main/java/com/google/samples/apps/nowinandroid/baselineprofile/BaselineProfileGenerator.kt
JolandaVerhoef: @keyboardsurfer can you please resolve this issue?

tikurahul (approved):

PR #69, opened 2022-05-27
Use `ExistingWorkPolicy.KEEP`.

tikurahul:
* You almost never want replace, given `REPLACE` enqueues a
new instance of the WorkRequest again.

### DIFF
--- a/sync/src/main/java/com/google/samples/apps/nowinandroid/sync/initializers/SyncInitializer.kt
+++ b/sync/src/main/java/com/google/samples/apps/nowinandroid/sync/initializers/SyncInitializer.kt
@@ -46,7 +46,7 @@ class SyncInitializer : Initializer<Sync> {
             // Run sync on app startup and ensure only one sync worker runs at any time
             enqueueUniqueWork(
                 SyncWorkName,
-                ExistingWorkPolicy.REPLACE,
+                ExistingWorkPolicy.KEEP,
                 SyncWorker.startUpSyncWork()
             )
         }


keyboardsurfer (approved):

tunjid (dismissed):

tunjid:
@tikurahul I went with replace to force another request to be enqueued so the app stays up to date on each start up.

By moving to `KEEP`, will sync still run on each start up?

tikurahul:
For that, we can just schedule another independent `OneTimeWorkRequest`. That should be done from inside the actual `Activity`. This is because, if you did this in `Application.onCreate()`, you are **always** going to end up cancelling and re-scheduling even when the process is being woken up for executing the actual work that was being scheduled.

tunjid:
@tikurahul Gotcha. The work done in sync unfortunately cannot be performed concurrently, so having two different work requests with different names for the `SyncWorker` is not feasible as described [here](https://github.com/android/nowinandroid/blob/main/core-data/src/main/java/com/google/samples/apps/nowinandroid/core/data/SyncUtilities.kt#L45).

In the case where the app is in the foreground when work is run, I'm assuming there are no issues? The issue with replace being the perpetual canceling and enqueuing of work when the process is being woken up to do work?

If I have the following requirements:

* Sync work should be performed on every app startup
*  The `SyncWorker` cannot be run concurrently; i.e only 1 `SyncWorker` must be guaranteed to run at any time

What is the appropriate configuration to use?

tunjid:
The docs for `KEEP` say
```
If there is existing pending (uncompleted) work with the same unique name, do nothing. 
```

So I'm guessing it will enqueue if no work is present, and if work is, it'll do nothing? If so I think the sync reqs are met.

tikurahul:
The `Worker` can check if a sync session is currently in progress; and no-op with a `Result.success()`. 

You should not be using a `UNIQUE` work policy to solve for concurrency problems. There are much better ways of solving for the latter that don't involve expensive book-keeping (relatively speaking).

tunjid:
When you say much better ways for dealing with concurrency, do you mean in general, or with `WorkManager`? I ask as in an unconstrained system I wouldn't be using `WorkManager` APIs.

> You should not be using a UNIQUE work policy to solve for concurrency problems.

The [docs](https://developer.android.com/topic/libraries/architecture/workmanager/how-to/managing-work#unique-work) suggest this is a valid use case?

i.e, needing to sync with `WorkManager` imposes some constraints that seemed to be easily solved with the use of `UNIQUE` work. Plus we get exponential back off for free amongst other things.

Sync is set up to be completely independent of the `app` or any `feature` modules, enqueuing new `OneTimeWork` from an `Activity` would leak `WorkManager` semantics into other modules and violate that design.

Ultimately, if I want to enqueue `UNIQUE` work to run each time the app starts up, will

``` kotlin
enqueueUniqueWork(
                SyncWorkName,
                ExistingWorkPolicy.KEEP,
                SyncWorker.startUpSyncWork()
            )
```

give me the behavior I want? If so, that'd be awesome. The lexicon suggests as much.

tikurahul:
I meant concurrency problems in general. 

> It will guarantee only one worker can exist at any point in time for the given name.

`UNIQUE` was mainly a way for apps to solve for **idempotency** (ignoring side-effects for cancellation here), and not concurrency. This is in fact a common misconception. 

In fact, there will me **multiple instances** of `SyncWorker` running when you `REPLACE`. One of them will be getting cancelled, while the other will be in `RUNNING`.

**Existing** and **Running** are **not** one and the same. 

> Sync is set up to be completely independent of the app or any feature modules

But with the use-case you described - it needs to know if the app is in the "foreground" or in the background. An `Activity` is the easiest way to do this. You can also register for `Activity` lifecycle callbacks (similar to what ProcessLifecycle does).

While, Sync might be an independent module the app still needs to be aware of sync in some fashion given global resources (jobs et al) are involved. WorkManager abstracts a lot of it away. This is why the `app` module should always be designed in a way where **it is in control**. In fact picking a design where an `app` module has very little say into when some work is being done on the apps behalf - causes impedance mismatches.

tunjid:
@tikurahul just saw you merged this; asking again, will

```
enqueueUniqueWork(
                SyncWorkName,
                ExistingWorkPolicy.KEEP,
                SyncWorker.startUpSyncWork()
            )
```

enqueue UNIQUE work to run each time the app starts up? If not, merging this breaks the app and I'll need to revert.

PR #71, opened 2022-05-28
Rename NiaNetwork & NiaPreferences, add DataSource suffix

madroidmaq:
Rename:
- NiaNetwork -> NiaNetworkDataSource
- NiaPreferences -> NiaPreferencesLocalDataSource

docs: [Data Layer naming-conventions](https://developer.android.com/jetpack/guide/data-layer#naming-conventions)

PS：I'm not sure if `AuthorDao` should be named `AuthorDatabaseDataSource`.

### DIFF
--- a/core-data/src/main/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstAuthorsRepository.kt
+++ b/core-data/src/main/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstAuthorsRepository.kt
@@ -23,7 +23,7 @@ import com.google.samples.apps.nowinandroid.core.database.dao.AuthorDao
 import com.google.samples.apps.nowinandroid.core.database.model.AuthorEntity
 import com.google.samples.apps.nowinandroid.core.database.model.asExternalModel
 import com.google.samples.apps.nowinandroid.core.datastore.ChangeListVersions
-import com.google.samples.apps.nowinandroid.core.datastore.NiaPreferencesLocalDataSource
+import com.google.samples.apps.nowinandroid.core.datastore.NiaPreferencesDataSource
 import com.google.samples.apps.nowinandroid.core.model.data.Author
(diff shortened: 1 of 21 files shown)


tunjid:
@madroidmaq if you could amend `NiaPreferencesLocalDataSource` to `NiaPreferencesDataSource` this will be good to go!

madroidmaq:
> @madroidmaq if you could amend `NiaPreferencesLocalDataSource` to `NiaPreferencesDataSource` this will be good to go!

@tunjid It‘s DONE. Some doubts about the naming I thought you might be able to help me, thanks a lot.

1. Does the name of `DAO` in `Room` need to add the suffix of `DataSource`?
    - if YES: I will submit a patch to rename if needed to add;
    - if NO: Whether the suffix `Dao` can be used as a naming convention for the `Room` data source, it is recommended to follow this naming convention;
2. What naming convention should be used to distinguish different Loacl `DataSource`, for example, if I use both `DataStore` and `File` to store data, should I name it `UserLoaclDataStore` or `UserDiskDataStore`? Or do you have a suggested naming convention here?

tunjid:
@madroidmaq 

1. Does the name of DAO in Room need to add the suffix of DataSource?

- No. A Data Access Object (DAO) is not necessarily the the same as a `DataSource`. A DAO from room may return multiple types for example, where a `DataSource` tends to manage only one type. Specifically, Room's DAOs represent the API for SQLite interactions. You can wrap it with a `DataSource`, or expose it to the `Repository` directly. It depends on what provides the most value for your app.

2. What naming convention should be used to distinguish different Local DataSource?

- The docs say to be [generic](https://developer.android.com/topic/architecture/data-layer#naming-conventions), so `Local` should be enough.

Do note, the docs are a guide; they aren't rigid or set in stone. Ultimately in your own application you should do what makes the most sense.

tunjid (approved):

PR #72, opened 2022-05-30
Add Previews for Feature screens

JolandaVerhoef:
Adding Previews for the different UI states:
- Loading
- Onboarding
- Content
- Empty

Added fake data to the various model classes. Not sure about adding this in production code, but couldn't think of a better approach, and will be removed in release build anyway.

### DIFF
--- a/feature-author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorScreen.kt
+++ b/feature-author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorScreen.kt
@@ -34,11 +34,9 @@ import androidx.compose.foundation.lazy.LazyListScope
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.material.icons.Icons.Filled
 import androidx.compose.material.icons.filled.ArrowBack
-import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
@@ -46,7 +44,6 @@ import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
-import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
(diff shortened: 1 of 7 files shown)


ricknout (commented):
For the device parameter, why not use https://developer.android.com/reference/kotlin/androidx/compose/ui/tooling/preview/Devices?

ricknout:
@JolandaVerhoef hmm but the device strings it generates seem to be the same as these constants:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui-tooling-preview/src/androidMain/kotlin/androidx/compose/ui/tooling/preview/Device.kt;l=55

feature-author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorScreen.kt
```diff
@@ -53,11 +55,13 @@ import androidx.hilt.navigation.compose.hiltViewModel
...
 import com.google.samples.apps.nowinandroid.core.model.data.Author
 import com.google.samples.apps.nowinandroid.core.model.data.FollowableAuthor
+import com.google.samples.apps.nowinandroid.core.model.data.previewAuthors
+import com.google.samples.apps.nowinandroid.core.model.data.previewNewsResources
 import com.google.samples.apps.nowinandroid.core.ui.LoadingWheel
+import com.google.samples.apps.nowinandroid.core.ui.component.NiaBackground
 import com.google.samples.apps.nowinandroid.core.ui.component.NiaFilterChip
 import com.google.samples.apps.nowinandroid.core.ui.newsResourceCardItems
-import com.google.samples.apps.nowinandroid.feature.author.AuthorUiState.Loading
-import com.google.samples.apps.nowinandroid.feature.author.R.string
```
alexvanyo, line 60: 🎉

feature-interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/InterestsScreen.kt
```diff
@@ -164,3 +175,99 @@ private fun InterestsContent(
...
+@OptIn(ExperimentalMaterial3Api::class)
+@Preview(name = "phone", device = "spec:shape=Normal,width=360,height=640,unit=dp,dpi=480")
+@Preview(name = "landscape", device = "spec:shape=Normal,width=640,height=360,unit=dp,dpi=480")
+@Preview(name = "foldable", device = "spec:shape=Normal,width=673,height=841,unit=dp,dpi=480")
+@Preview(name = "tablet", device = "spec:shape=Normal,width=1280,height=800,unit=dp,dpi=480")
+@Composable
+fun InterestsScreenLoading() {
+    NiaTheme {
+        NiaBackground {
+            Scaffold(
```
alexvanyo, line 222: I'm curious why these previews are being wrapped in a `Scaffold`? I don't see an obvious reason to have it.

alexvanyo (approved):
LGTM, just one question about adding in the `Scaffold` in the previews.

ricknout:
> There's a current bug that doesn't allow you to use constants like Devices.PHONE. Can we merge this and revisit later?

SGTM

ricknout (approved):

feature-interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/InterestsScreen.kt
JolandaVerhoef: Hmm I thought it was needed.. But you're right, it doesn't make a difference. Removed them!

alexvanyo (approved):
Looks good!

camina-apps:
why we are not using `PreviewParameterProvider`?

Issue #89, opened 2022-05-31
Consolidate version catalog definition of kotlinxCoroutines and kotlinxCoroutinesTest


alexvanyo:
Right now, the version catalog has two independent entries for kotlinx.coroutines, `kotlinxCoroutines` and `kotlinxCoroutinesTest`.

Since the runtime and test libraries are published with the same versioning, we should just have a single `kotlinxCoroutines` entry and use that version for both. That would avoid the potential to have mismatching versions (for example, if #87 and #88 were to be merged separately)

alexvanyo:
Fixed with #100

PR #100, opened 2022-06-03
Consolidate version catalog definition

Rocksnake:
#89 
What I did: 
1.removed kotlinxCoroutinesTest entry
2.replaced kotlinxCoroutinesTest with kotlinxCoroutines


alexvanyo (approved):
This looks great, thank you for the contribution!

PR #106, opened 2022-06-05
fix: gradle build failed - 'VERSION_CATALOGS' feature

Joehaivo:
to fix gradle build failed - "Using dependency catalogs requires the activation of the matching feature preview." when first running `app` module


Joehaivo:
CLA have been updated.

PR #107, opened 2022-06-06
Rename FollowedInterestsState to FollowedInterestsUiState

madroidmaq:
read more: #42


JolandaVerhoef:
The build is failing due to some unresolved references to the UiState. Can you take a look?

madroidmaq:
@JolandaVerhoef I tried many times and basically it gets cancelled during the execution of the Android UnitTest (suspect that it might be taking too long to execute). Is there any way I can pass the CI check?
image

JolandaVerhoef (approved):

dturner:
I was recently looking at `ForYouViewModel` trying to understand where each `UiState` variable was used in `ForYouScreen`. It turns out that `FollowedInterestsUiState` isn't used for UI state at all. It's an intermediate private variable used by other `combine` functions to construct UI state. With this in mind, I don't think it should have `Ui` in its suffix.

PR #108, opened 2022-06-06
use serializedName to serialize NewsResourceType

yanzm:
NewsResourceTypeSerializer uses serializedName to deserialize, so it should also use serializedName to serialize.


tunjid (approved):
Great catch, thanks!

JolandaVerhoef (approved):

PR #116, opened 2022-06-06
[CI] Collect all test reports, not just app's

JoseAlcerreca:


JoseAlcerreca:
I verified that reports are there now.

PR #124, opened 2022-06-07
remove explicit cancel calls in tests

### LINKED ISSUE #122: Simplify tests with turbine by removing explicit cancel calls
At some point `turbine` started cancelling automatically, so we can remove the explicit `cancel` calls in tests to simplify them.

Reference:
https://github.com/cashapp/turbine/issues/4
https://github.com/cashapp/turbine/issues/90
https://github.com/cashapp/turbine/pull/107

camina-apps:
fixes #122
Simplify tests with turbine by removing explicit cancel calls


JoseAlcerreca:
@camina-apps please describe why you do the changes. I assume they were added for a reason. Why are they redundant?

camina-apps:
@JoseAlcerreca
In short the turbine frameworks cancels automatically, there is no need anymore to do this in the tests.
More info in the linked issue from @alexvanyo

PR #125, opened 2022-06-08
Add compose compiler metrics option

takahirom:
You can check the compiler metrics and reports by using Gradle options.

> -PenableComposeCompilerMetrics=true -PenableComposeCompilerReports=true
image

image

https://github.com/androidx/androidx/blob/androidx-main/compose/compiler/design/compiler-metrics.md


build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
```diff
@@ -56,7 +57,7 @@ internal fun Project.configureKotlinAndroid(
                 "-opt-in=kotlin.Experimental",
                 // Enable experimental kotlinx serialization APIs
                 "-opt-in=kotlinx.serialization.ExperimentalSerializationApi"
-            )
+            ) + buildComposeMetricsParameters()
```
takahirom, line 60: I think it is possible to do this with AndroidCompose.kt, but I think there is a tradeoff.
The advantage is that you can put Compose stuff into Compose build scripts.
The disadvantage is that freeCompilerArgs cannot be managed centrally.
Either is fine. 🙆‍♂️

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -38,5 +38,33 @@ internal fun Project.configureAndroidCompose(
...
         }
+
+        kotlinOptions {
+            freeCompilerArgs = freeCompilerArgs + buildComposeMetricsParameters()
+        }
+    }
+}
+
+private fun Project.buildComposeMetricsParameters(): List<String> {
+    val metricParameters = mutableListOf<String>()
```
takahirom, line 49: I think buildList{} could be used, but that API is still ExperimentalStdlibApi, so I am using MutableList.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
```diff
@@ -70,6 +70,6 @@ internal fun Project.configureKotlinAndroid(
     }
 }
 
-private fun CommonExtension<*, *, *, *>.kotlinOptions(block: KotlinJvmOptions.() -> Unit) {
+fun CommonExtension<*, *, *, *>.kotlinOptions(block: KotlinJvmOptions.() -> Unit) {
```
takahirom, line 73: It can be separated into files like kotlinOptions.kt, but I'm changing it to public because it looks like the rules weren't established yet and I didn't think it was wrong to have it in Kotlin's configuration.
If you have any opinions, please let me know. 🙆‍♂️

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -38,5 +38,33 @@ internal fun Project.configureAndroidCompose(
...
+    val enableMetricsProvider = project.providers.gradleProperty("enableComposeCompilerMetrics")
+    val enableMetrics = (enableMetricsProvider.orNull == "true")
+    if (enableMetrics) {
+        val metricsFolder = File(project.buildDir, "compose-metrics")
```
takahirom, line 53: I don't think using File is a good idea, but I'm trying to use File to take into account Windows paths and such. There may be a better API, but I'm using this one because it only affects me when I add options and has little impact.

takahirom (commented):
I've tried to fix it. Please let me know if you have any input. 🙏

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
alexvanyo: Yup that makes sense, `mutableListOf` looks fine!

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
alexvanyo: This makes sense for sure! We don't have any strong rules for how functions can be shared in the build logic, so I see no issue with this right now.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
alexvanyo: Yeah, `File` here looks good for those reasons.

alexvanyo (approved):
Sorry for the long delay here, this looks great, thanks for the contribution and the detail comments!

JolandaVerhoef (approved):

PR #129, opened 2022-06-09
Make :core-model to pure kotlin module

fornewid:
After [this commit](https://github.com/android/nowinandroid/commit/9651c6a581c93721cd0976b7d58f76af2413eee0), android dependencies were removed from `:core-model` module.
So we can make `:core-model` to pure kotlin module.
This avoids adding dependencies like Room or Serialization.


mmoczkowski (approved):
Everything seems to be working ok. Thanks for the contribution!

simona-anomis (approved):

PR #137, opened 2022-06-15
use contentPadding of LazyColumn for TabContent

yanzm:
|before|after|
|-|-
|Screenshot_20220615_214543|Screenshot_20220615_214500|

### DIFF
--- a/feature-interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/InterestsScreen.kt
+++ b/feature-interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/InterestsScreen.kt
@@ -150,7 +150,6 @@ private fun InterestsContent(
                     topics = uiState.topics,
                     onTopicClick = navigateToTopic,
                     onFollowButtonClick = followTopic,
-                    modifier = Modifier
                 )
             }
             1 -> {
@@ -158,7 +157,6 @@ private fun InterestsContent(
                     authors = uiState.authors,
                     onAuthorClick = navigateToAuthor,
                     onFollowButtonClick = followAuthor,
-                    modifier = Modifier
                 )
             }
         }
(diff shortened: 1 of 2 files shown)


feature-interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/InterestsScreen.kt
```diff
@@ -152,15 +150,15 @@ private fun InterestsContent(
                     topics = uiState.topics,
                     onTopicClick = navigateToTopic,
                     onFollowButtonClick = followTopic,
-                    modifier = Modifier.padding(top = 8.dp)
+                    modifier = Modifier
```
alexvanyo, line 153: Nit: since it's customary for the default parameter for a modifier argument to just be `Modifier`, I think it's generally preferable to remove passing it at all and just use the default.

feature-interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/TabContent.kt
```diff
@@ -40,7 +41,8 @@ fun TopicsTabContent(
     modifier: Modifier = Modifier
 ) {
     LazyColumn(
-        modifier = modifier.padding(horizontal = 16.dp)
+        modifier = modifier.padding(horizontal = 16.dp),
+        contentPadding = PaddingValues(vertical = 8.dp)
```
alexvanyo, line 45: This does something ever so slightly different, since this is using `vertical` instead of `top`. Can we keep the old spacing amount, but looking much better on the top?

alexvanyo (commented):
This is a great change, thank you for the contribution! Just left two minor comments for you.

alexvanyo (approved):
Looks good, thanks for the contribution!

JolandaVerhoef (approved):

PR #147, opened 2022-06-21
Add bookmarking to UserDataRepository

tunjid:
This allows `UserDataRepository` add or remove bookmarks.

### DIFF
--- a/core-data/src/test/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstUserDataRepositoryTest.kt
+++ b/core-data/src/test/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstUserDataRepositoryTest.kt
@@ -47,7 +47,7 @@ class OfflineFirstUserDataRepositoryTest {
     }
 
     @Test
-    fun offlineFirstTopicsRepository_toggle_followed_topics_logic_delegates_to_nia_preferences() =
+    fun offlineFirstUserDataRepository_toggle_followed_topics_logic_delegates_to_nia_preferences() =
         runTest {
             subject.toggleFollowedTopicId(followedTopicId = "0", followed = true)
 
@@ -68,7 +68,8 @@ class OfflineFirstUserDataRepositoryTest {
             )
 
             assertEquals(
-                niaPreferencesDataSource.followedTopicIds
+                niaPreferencesDataSource.userDataStream
+                    .map { it.followedTopics }
                     .first(),
                 subject.userDataStream
                     .map { it.followedTopics }
@@ -77,7 +78,7 @@ class OfflineFirstUserDataRepositoryTest {
         }
 
     @Test
-    fun offlineFirstTopicsRepository_set_followed_topics_logic_delegates_to_nia_preferences() =
+    fun offlineFirstUserDataRepository_set_followed_topics_logic_delegates_to_nia_preferences() =
         runTest {
             subject.setFollowedTopicIds(followedTopicIds = setOf("1", "2"))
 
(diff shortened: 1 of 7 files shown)


calren (approved):
👍

core-datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
```diff
@@ -106,51 +120,47 @@ class NiaPreferencesDataSource @Inject constructor(
...
+        )
     }
 
-    suspend fun toggleFollowedAuthorId(followedAuthorId: String, followed: Boolean) {
+    /**
+     * Sets the value provided by [listSetter] into the [DslList] read by [listGetter]
+     */
+    private suspend fun <T : DslProxy> DataStore<UserPreferences>.setList(
+        listGetter: (UserPreferencesKt.Dsl) -> DslList<String, T>,
+        listSetter: (DslList<String, T>) -> List<String>,
```
alexvanyo, line 149: It seems like there's maybe a little too much abstraction here?

It's a bit confusing that `listSetter` doesn't actually set the list passed to it, but it (may) modify the list it is given, and returns the list that should later be added.

I get why you're doing this though, since the Kotlin bindings that are automatically generated this less than ideal to do.

I think the original way without the `editList`/`setList` helpers was fine, and more straightforward, if a bit verbose.

Maybe we can add a feature request to the protobuf Kotlin bindings to allow exposing these as `MutableList`s?

alexvanyo (commented):
Looks good, just two small comments.

core-data/src/test/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstUserDataRepositoryTest.kt
```diff
@@ -89,11 +90,43 @@ class OfflineFirstUserDataRepositoryTest {
...
+                    .map { it.followedTopics }
                     .first(),
                 subject.userDataStream
                     .map { it.followedTopics }
                     .first()
             )
         }
+
+    @Test
+    fun offlineFirstTopicsRepository_bookmark_news_resource_logic_delegates_to_nia_preferences() =
```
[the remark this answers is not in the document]
tunjid: Great catch ty!

core-datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
tunjid: Great point on the complexity of the abstraction, but it felt clumsy and error prone copy pasting the same try block 5 times with minimum variation, especially amongst the toggle methods for topics, authors and news resources.

I've renamed `listSetter` to `listModifier` to better convey what it does.

core-data/src/test/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstUserDataRepositoryTest.kt
```diff
@@ -47,7 +47,7 @@ class OfflineFirstUserDataRepositoryTest {
     }
 
     @Test
-    fun offlineFirstTopicsRepository_toggle_followed_topics_logic_delegates_to_nia_preferences() =
+    fun offlineFirstUserRepository_toggle_followed_topics_logic_delegates_to_nia_preferences() =
```
alexvanyo, line 50: Nit: Technically the full name here is `OfflineFirstUserDataRepository` 😄

alexvanyo (approved):
Looks good, just one nit!

core-data/src/test/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstUserDataRepositoryTest.kt
tunjid: Lol ty!

PR #151, opened 2022-06-23
Add topic chips for NewsResourceCard

calren:
Two follow ups for future PRs:
- Chip should be disabled if user is not following the topic
- Handle topic chip interaction

topic_chips


core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -267,9 +267,35 @@ fun NewsResourceShortDescription(
 
 @Composable
 fun NewsResourceTopics(
-    newsResource: NewsResource
+    topics: List<Topic>,
+    modifier: Modifier = Modifier
 ) {
-    TODO()
+    LazyRow(
```
alexvanyo, line 273: Just confirming the choice of a scrollable `LazyRow` here, instead of a `FlowRow` or something similar?

core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Chip.kt
```diff
@@ -101,6 +107,80 @@ fun NiaFilterChip(
...
+@OptIn(ExperimentalMaterial3Api::class)
+fun NiaTopicChip(
+    enabled: Boolean,
+    onClick: (Boolean) -> Unit,
+    modifier: Modifier = Modifier,
+    label: @Composable () -> Unit
+) {
+    SuggestionChip(
+        enabled = enabled,
+        onClick = { onClick },
```
alexvanyo, line 128: I think this should just be `onClick = onClick`? Right now, `onClick` won't be called I believe.

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
calren: yup, design in figma has it as a scrollable row right now.

core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Chip.kt
calren: good catch, fixed

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -267,9 +267,35 @@ fun NewsResourceShortDescription(
...
+private fun uppercaseFormat(string: String): String {
+    val locale = ConfigurationCompat.getLocales(LocalConfiguration.current).get(0)
+
+    val uppercaseFormatted = if (locale != null) {
+        string.uppercase(locale)
+    } else {
+        string.uppercase()
+    }
+
+    return uppercaseFormatted
```
JolandaVerhoef, line 298: I feel like there should be an easier fool-proof way to uppercase a String. Asked about it in chat.

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -306,7 +332,9 @@ private val newsResource = NewsResource(
     id = "1",
     episodeId = "1",
     title = "Title",
-    content = "Content",
+    content = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor" +
+        " incididunt ut labore et dolore magna aliqua. Habitant morbi tristique senectus et netus" +
+        " et malesuada fames.",
```
JolandaVerhoef, line 337: If you want to, you can replace this preview element with `previewNewsResources[0]`

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
JolandaVerhoef: I think this should not be a `LazyRow`, as these are not recommended when the number of elements is small (too much overhead). Instead, you could simply use a `Row` with `horizontalScroll` modifier. @bentrengrove to confirm.

JolandaVerhoef (changes requested):

core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Chip.kt
```diff
@@ -101,6 +107,80 @@ fun NiaFilterChip(
...
+ * Now in Android topic chip that displays a topic name text.
+ *
+ * @param enabled Whether the chip is currently enabled.
+ * @param onClick Called when the user clicks the chip for more options.
+ * @param modifier Modifier to be applied to the chip.
+ * @param label The text label content.
+ */
+@Composable
+@OptIn(ExperimentalMaterial3Api::class)
+fun NiaTopicChip(
```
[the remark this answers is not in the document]
ricknout: Yes, `NiaTopicTag` is the component from Figma and the one we should be using. I might need a little more context on the scrollable row issue, but happy to try and fix it

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
bentrengrove: It would depend on how many total there are, how many are displayed on screen at once and how heavy each item is to compose. You would have to benchmark it for a proper answer but in general it does switch over to Lazy being better pretty quick but if the majority of items are always displayed on the screen then Row is probably better.

calren:
Planning to look into this again late next week.

(We have a team offsite this week, and first half of next week is holidays in US.)

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
calren: most items have just 1-3 elements in the row, so updated to use `Row`

core-designsystem/build.gradle.kts
```diff
@@ -41,4 +41,10 @@ dependencies {
     api(libs.androidx.compose.runtime)
     lintPublish(project(":lint"))
     androidTestImplementation(project(":core-testing"))
+
+    // TODO : Remove these dependency once we upgrade to Android Studio Dolphin b/228889042
+    // These dependencies are currently necessary to render Compose previews
+    debugImplementation(libs.androidx.customview.poolingcontainer)
+    debugImplementation(libs.androidx.lifecycle.viewModelCompose)
+    debugImplementation(libs.androidx.savedstate.ktx)
```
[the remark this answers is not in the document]
calren: unfortunately still needed in chipmunk

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
calren: removed in favor of `uppercase(Locale.getDefault())` based on chat. seems to be what most compose samples are using too https://github.com/android/compose-samples/search?q=uppercase

core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Chip.kt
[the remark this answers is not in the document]
calren: using `NiaTopicTag` with a `Row` and `horizontalScroll` makes the chips really narrow for some reason:

narrow_chips

```
    Row(
        modifier = modifier.horizontalScroll(rememberScrollState()), // causes narrow chips
        horizontalArrangement = Arrangement.spacedBy(4.dp),
        ) {
        for (topic in topics) {
            NiaTopicTag(
                followed = true, // Todo
                onFollowClick = { }, // Todo
                onUnfollowClick = { }, // Todo
                onBrowseClick = { }, // Todo
                text = { Text(text = topic.name.uppercase(Locale.getDefault())) }
            )
        }
    }
```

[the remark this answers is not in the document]
calren: Digging into it a bit, removing the `weight` on the buttons fixes the issue: https://github.com/android/nowinandroid/blob/main/core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Button.kt#L327

but I'm not sure if removing that line causes other issues. A quick test didn't surface obvious issues.

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
JolandaVerhoef: @calren can you address this comment as well? :)

calren: yup, done!

calren:
alrighty, i think this is ready for another review now that https://github.com/android/nowinandroid/pull/178 is merged

JolandaVerhoef (approved):

JolandaVerhoef:
@calren can you try to fix the branch conflict?

PR #156, opened 2022-06-25
Added a variable to declare a route concatenated with arguments

andhiratobing:
Added a route with argument contracts to make it easier to read, but if too many arguments to send, a workaround could be to use parcelable.


core-navigation/src/main/java/com/google/samples/apps/nowinandroid/core/navigation/NiaNavigationDestination.kt
```diff
@@ -36,3 +36,11 @@ interface NiaNavigationDestination {
      */
     val destination: String
 }
+
+interface NiaNavigationWithArguments {
```
simona-anomis, line 40: Having another interface just for this seems like an overkill. You could just add:

`val routeWithArgs = "$route/{$topicIdArg}"`
` val routeWithArgs = "$route/{$authorIdArg}"`

To TopicDestination and AuthorDestination accordingly.

simona-anomis (changes requested):

core-navigation/src/main/java/com/google/samples/apps/nowinandroid/core/navigation/NiaNavigationDestination.kt
andhiratobing: I think the interface is different. There needs to be a separation between destination and route with arguments. By making the interface usable in multiple navigations, it eliminates repetitive write variables.

simona-anomis:
Hey, thanks for the contribution! However, we had some internal changes on navigation side just merged, so this PR won't fit in with the structure currently, apologies for that and please keep continue contributing, we really appreciate it :)

Issue #167, opened 2022-07-04
[FR]: Gradle properties are not passed down to included builds


oheyadam:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

Discussion [here](https://github.com/gradle/gradle/issues/2534). I noticed that the included build doesn't have a `gradle.properties` file, which means things like parallel execution, task output caching, etc, are not enabled for it since the global properties are not passed down to included builds.

In my testing on an M1 Pro machine, enabling these basic flags takes build times of the `build-logic` project from seconds down to milliseconds. 
```
org.gradle.parallel=true
org.gradle.caching=true
org.gradle.configureondemand=true
```

### Describe the solution

Add a `gradle.properties` file to the build-logic project with some configs. I'd love to submit a PR once my Individual CLA is approved if you think this is worthwhile.

### Additional context

|Before|After|
|------|------|
|![image](https://user-images.githubusercontent.com/18615546/177143564-ce854614-4e4f-4f18-88c8-2cf8408e9848.png)|![image](https://user-images.githubusercontent.com/18615546/177143583-b02d10d8-9cdf-43ec-84af-038673907554.png) |

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dturner:
Good suggestion. I was able to decrease the `build-logic` build time (from clean) from 6s to 2s by including `gradle.properties`. 

Please feel free to submit a PR for this and we will approve.

oheyadam:
Hey @dturner, thanks! I just submitted [this](https://github.com/android/nowinandroid/pull/177) PR

PR #177, opened 2022-07-15
Add a gradle.properties file to the build-logic build

oheyadam:
This PR closes adds a `gradle.properties` to the `build-logic` build because the global properties aren't passed down to composite builds. Gradle issue [here](https://github.com/gradle/gradle/issues/2534)

I've set the following flags to true, but feel free to modify these properties further according to the project requirements down the line.

```
org.gradle.parallel=true
org.gradle.caching=true
org.gradle.configureondemand=true

```
This closes https://github.com/android/nowinandroid/issues/167


alexvanyo (approved):
LGTM, and that initial set of properties also looks good!

JolandaVerhoef (approved):

PR #186, opened 2022-07-20
Add state holder for the NiaApp composable

manuelvicnt:
The following changes have been added in this PR:

- `NiaAppState` is introduced to simplify the `NiaApp` composable. It handles UI logic, hoists some UI elements' state and exposes non-application-data-related state.
- The NavController has the state holder as the source of truth. Before, there was no source of truth, the responsibilities were shared between multiple composables.
- `NiaApp` class is renamed to `NiaApplication`
- The navigation solution is more type safe, no other classes apart from the different destinations know how to create routes or retrieve arguments from a `backStackEntry`


feature-topic/src/main/java/com/google/samples/apps/nowinandroid/feature/topic/navigation/TopicNavigation.kt
```diff
@@ -24,20 +25,26 @@ import com.google.samples.apps.nowinandroid.core.navigation.NiaNavigationDestina
...
 
 object TopicDestination : NiaNavigationDestination {
-    override val route = "topic_route"
-    override val destination = "topic_destination"
     const val topicIdArg = "topicId"
+    override val route = "topic_route/{$topicIdArg}"
+    override val destination = "topic_destination"
+
+    fun createNavigationRoute(topicIdArg: String): String {
+        return "topic_route/$topicIdArg"
```
alexvanyo, line 33: Since we're consolidating these here, does it make sense to also add the `Uri.encode`/`Uri.decode` pair to handle ids with special characters?

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
```diff
@@ -0,0 +1,131 @@
...
+            unselectedIcon = ImageVectorIcon(NiaIcons.Grid3x3),
+            iconTextId = interestsR.string.interests
+        )
+    )
+
+    // --------------------
+    // NAVIGATION LOGIC
+    // --------------------
+
+    fun navigate(destination: NiaNavigationDestination, route: String? = null) {
```
alexvanyo, line 91: Can I request a KDoc here? Just looking at the method being called, it's not super clear to me how the `route` and `destination` relate to each other, and the implications of a `route` not being supplied.

feature-topic/src/main/java/com/google/samples/apps/nowinandroid/feature/topic/navigation/TopicNavigation.kt
manuelvicnt: Sure thing! Added, thanks

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
manuelvicnt: Sure! Added, thanks

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -108,53 +92,44 @@ fun NiaApp(windowSizeClass: WindowSizeClass) {
                             )
                         )
                 ) {
-                    if (windowSizeClass.widthSizeClass != WindowWidthSizeClass.Compact &&
-                        windowSizeClass.heightSizeClass != WindowHeightSizeClass.Compact
-                    ) {
+                    if (appState.shouldShowNavRail) {
```
tunjid, line 95: Neat

tunjid (approved):
LGTM, especially the centralization of `shouldShowBottomNav` and `shouldShowNavRail`

kaustubhbhagwat (approved):

dturner (approved):
LGTM. Thanks for doing this.

PR #203, opened 2022-07-30
Put some modifiers in correct order to respect official API guidelines

necatisozer:
https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter

> Elements accept and respect a Modifier parameter
> Element functions MUST accept a parameter of type Modifier. This parameter MUST be named "modifier" and MUST appear as the first optional parameter in the element function's parameter list.


calren (approved):
Thanks for this!

alexvanyo (approved):
Looks good to me!

One technicality note: the guidelines don't state that all parameters with defaults have to come after parameters without defaults, so technically these two composables don't violate the guidance.

The case where that's most useful is when there should be a `Modifier` and a `content` parameter which is a trailing lamda.

But in these cases there aren't trailing lambdas, and I agree with putting all of the parameters with defaults last.

PR #212, opened 2022-08-03
Add timber for debugging

oyeraghib:
I am new to Android and still learning. I went through the complete app and the code looks very interesting. Thanks to the Google team for building such an amazing and helpful app.

I thought of adding Timber for debugging as it really eases the process of logging by avoiding necessarily adding TAG. Not sure if it is a part of AndroidX library. 
I remember learning to to use Timber instead of Logs from my classes at Udacity (where most of educators were from Google) so I think it might be a good add to the project. Thanks!

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -127,6 +127,9 @@ dependencies {
     kapt(libs.hilt.compiler)
     kaptAndroidTest(libs.hilt.compiler)
 
+    //Timber
+    implementation("com.jakewharton.timber:timber:5.0.1")
+
     // androidx.test is forcing JUnit, 4.12. This forces it to use 4.13
     configurations.configureEach {
         resolutionStrategy {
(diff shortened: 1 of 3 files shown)


calren:
Glad you're finding this app helpful, and thanks for bringing this up!

Agreed that we need to use a logging library for the project, but we want to do a bit more investigation before committing to Timber. Created an issue to track this: https://github.com/android/nowinandroid/issues/214

dturner:
I don't think we need a logging library. 

We don't do a lot of logging - only one instance for jank stats. Also using the debugger and breakpoints is often more efficient than `Log.d(...)` for understanding the code.

If we can avoid using, and by extension, recommending a 3rd party library that's a good thing imo because it's one more decision we don't have to justify.

dturner:
If you disagree please speak up 😃 otherwise I'll close this.

oyeraghib:
@dturner I agree, there are not many cases of `Log.d` except few. And debugging is better. Upto the maintainers if they think it's worth it, otherwise I am okay with closing it :)

mmoczkowski:
Thank you for contributing Mohd! After reviewing the discussion I'm closing this issue, but please don't feel discouraged from submitting next PRs.

PR #215, opened 2022-08-08
fix bottom space on forYou's NewsFeed last item

### LINKED ISSUE #204: [Bug]: forYou's NewsFeed has no space at the bottom of the last item

### What happened?

The NewsFeed of forYou does not leave spaces at the bottom of the last item.

if spaces are added, the user's visual experience may be better.

Consider simply adding contentPadding = PaddingValues(bottom = 16.dp).

### Relevant logcat output

_No response_

### Code of Conduct

YvesKalume:
fix #204 by adding `8.dp` as padding at the bottom of the last item

Screenshot_20220808_161238


feature-foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -219,7 +218,7 @@ fun ForYouScreen(
                     )
 
                     item {
-                        Spacer(Modifier.windowInsetsBottomHeight(WindowInsets.safeDrawing))
```
alexvanyo, line 222: We don't want to remove the insets support here, as otherwise cards could be obscured by the bottom insets when we're using the navigation rail instead of the navigation bar.

If we add another `Spacer` with the `8.dp` height that should fix the issue and avoid overlapping with insets!

alexvanyo (changes requested):

feature-foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
YvesKalume: Done

alexvanyo (approved):
Looks great! Would it be possible to update with the latest changes that were merged in? Thank you!

YvesKalume:
Getting an error : error: failed to push some refs to 'https://github.com/YvesKalume/nowinandroid'
To https://github.com/YvesKalume/nowinandroid
!	refs/heads/main:refs/heads/main	[remote rejected] (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/Build.yaml` without `workflow` scope)

But my token has all permissions.

alexvanyo:
Hmm, not sure how to help debug. If you're using GitHub's UI, maybe instead doing it locally would work? Or vice versa.

PR #220, opened 2022-08-10
Add data logic for theme switcher

alexvanyo:
Adds the data logic for the theme switcher (but not the UI implementation)

To manually test, you can override the resolved theme as returned by the data layer.


alexvanyo:
Superceded by #330

PR #228, opened 2022-08-16
fix bottom space on forYou's NewsFeed last item 

### LINKED ISSUE #204: [Bug]: forYou's NewsFeed has no space at the bottom of the last item

YvesKalume:
fix #204  
I have resolved the issue of merging previous PR #215 with fixed issues

### DIFF
--- a/feature-foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
+++ b/feature-foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
@@ -28,13 +28,13 @@ import androidx.compose.foundation.layout.WindowInsetsSides
 import androidx.compose.foundation.layout.consumedWindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.only
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.safeDrawing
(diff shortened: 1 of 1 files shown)


feature-foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -203,8 +205,13 @@ fun ForYouScreen(
                     loadingContentDescription = R.string.for_you_loading
                 )
 
-                item(span = { GridItemSpan(maxLineSpan) }) {
```
alexvanyo, line 206: Ah sorry, this wasn't a clean merge. We changed to use a `LazyGrid` view here, and therefore we want the spacing items to fill up the entire span (all columns, if there are more than one).

That ensures that this spacing is applied correctly, as opposed to being an item at the end of a row after an item that doesn't fill up all of the space.

Let's add the `span = { GridItemSpan(maxLineSpan) }` back in to the `item`s. We also probably could have just one `item` instead, and have a `Column` of the two `Spacer`s.

YvesKalume: Done ! Can you check ?

alexvanyo (approved):
Looks good, sorry for the merge conflicts and thank you for updating!

YvesKalume:
Now that the conflict is resolved, I hope this will be merged before there is another 😅

JolandaVerhoef (approved):

PR #231, opened 2022-08-18
Update app icon using Image Asset Studio

### LINKED ISSUE #216: [Bug]: App icon doesn't exist on older APIs

### What happened?

The default icon is used on older APIs, instead of the normal Now in Android one.

### Relevant logcat output

_No response_

### Code of Conduct

JolandaVerhoef:
Fixes #216


app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -17,5 +17,4 @@
 <adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
     <background android:drawable="@drawable/ic_launcher_background"/>
     <foreground android:drawable="@drawable/ic_launcher_foreground"/>
-    <monochrome android:drawable="@drawable/ic_launcher_foreground" />
```
alexvanyo, line 20: It looks like this is going to remove adaptive icon support, these files should probably remain unchanged?

Image Asset Studio through AS should probably have an option/add this in automatically too.

alexvanyo (changes requested):
Looks good, verified icon is showing up on API 23. Just one note to keep adaptive icon support.

app/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
```diff
@@ -17,5 +17,4 @@
 <adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
     <background android:drawable="@drawable/ic_launcher_background"/>
     <foreground android:drawable="@drawable/ic_launcher_foreground"/>
-    <monochrome android:drawable="@drawable/ic_launcher_foreground" />
```
alexvanyo, line 20: `ic_launcher_round` still has the `monochrome` item being removed, so round icons don't have a monochrome version available.

Could we keep the `monochrome` item in `ic_launcher` and ic_launcher_round` (net result is those files are unchanged by this PR)

That avoids adding another overload file, and aligns with the approach here: https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive#add_your_adaptive_icon_to_your_app

JolandaVerhoef: Removed roundIcon from the manifest, as per documentation. Removed v33 overload and included in v26 instead.

alexvanyo (commented):
Removing `android:roundIcon` from the manifest means we lose round icon support on API 25.

alexvanyo (approved):

Issue #234, opened 2022-08-20
[Bug]: Potentially redundant configuration inside AndroidLibraryComposeConventionPlugin convention plugin


StylianosGakis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

This is more of a question, but is there a reason that the plugin "com.android.library" is applied both inside [AndroidLibraryConventionPlugin](https://github.com/android/nowinandroid/blob/2cf02177a11b931479db1ed07d7e557f3ebd8030/build-logic/convention/src/main/kotlin/AndroidLibraryConventionPlugin.kt#L31) and inside [AndroidLibraryComposeConventionPlugin](https://github.com/android/nowinandroid/blob/2cf02177a11b931479db1ed07d7e557f3ebd8030/build-logic/convention/src/main/kotlin/AndroidLibraryComposeConventionPlugin.kt#L26)?

I've checked and there is no place (rightly so) where a module only imports "nowinandroid.android.library.compose" but does not import "nowinandroid.android.library", therefore I feel like this does not need to be inside the `AndroidLibraryComposeConventionPlugin`.

Is my assumption correct, or am I potentially missing something?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

StylianosGakis:
Closed as after the discussion in [this PR](https://github.com/android/nowinandroid/pull/235) made it obvious that this is actually not a good idea to change this after all!

PR #235, opened 2022-08-20
Remove redundant configuration in AndroidLibraryComposeConventionPlugin

StylianosGakis:
Related to https://github.com/android/nowinandroid/issues/234

### DIFF
--- a/build-logic/convention/src/main/kotlin/AndroidLibraryComposeConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/AndroidLibraryComposeConventionPlugin.kt
@@ -23,7 +23,6 @@ import org.gradle.kotlin.dsl.getByType
 class AndroidLibraryComposeConventionPlugin : Plugin<Project> {
     override fun apply(target: Project) {
         with(target) {
-            pluginManager.apply("com.android.library")
             val extension = extensions.getByType<LibraryExtension>()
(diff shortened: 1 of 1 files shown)


alexvanyo (commented):
Yeah this is redundant, but not sure if it has an ill effects? I wonder if removing this would enforce an ordering requirement of `nowinandroid.android.library` having to be declared before `nowinandroid.android.library.compose`. And not sure if such an ordering requirement would be bad?

I don't feel strongly either way, although I do sort of like the idea that each individual convention plugin defines the plugins it is depending on, and therefore are each more independent from each other (doesn't require that another one is being applied)

Overall consistency would probably be the main thing to achieve, I think `AndroidApplicationComposeConventionPlugin` has the same redundancy?

StylianosGakis:
I just tried changing the ordering the other way around and it is in fact a problem. My initial thought and intent was to reduce the mental overhead when looking inside each of the plugins and what they do since I found it a bit confusing that many plugins were doing the same thing yet we were always applying both and couldn't apply the compose one without the core library one.
But with this newfound knowledge I'd say your point is absolutely valid, it's not a good idea to remove it, despite all the other reasons I had in my mind. Thanks for this, will close my other PR too!

alexvanyo:
Sounds good, thanks for the discussion here!

PR #238, opened 2022-08-23
Slightly more precise flow transformations in the ForYouViewModel

tunjid:
This PR tries to improve the `Flow` processing in the `ForYouViewModel`.

At the moment both the `feedState` and `interestsSelectionState` start with `combine`, recollecting from all `Flows` if if any of them change.

This PR tries to identify `driver` `Flows` and starts collecting from the other `Flows` only if the driver `Flow` changes, reducing the amount of `Flows` being collected from at any one time.


tunjid:
There's some weird stuff with the automerger. I'm gonna close this and reopen.

PR #241, opened 2022-08-24
Move core&feature modules to subfolders

madroidmaq:
Use the subfolder method in Android Studio to move the **"core-"** and **"feature-"** modules to the corresponding subfolder. Use the form of directories instead of prefixes to distinguish modules.

The introduction of subfolder in Android Studio is as follows:
image

Changes in the directory structure before and after adjustment：

|  before adjustment    |   after adjustment(Expand)   |  after adjustment(Collapse)  | 
| ---- | ---- | ---- |
|  image    |   image   |  image |


madroidmaq:
According to the Figure 1 in the [official document](https://developer.android.com/topic/modularization#what-is-modularization), it seems to be done in the way of subfolder.

image
Figure 1: Dependency graph of a sample multi-module codebase

dturner (approved):

mmoczkowski (approved):
Looking good, thank you for your contribution!

madroidmaq:
@dturner @mmoczkowski 

in this [commit](https://github.com/android/nowinandroid/pull/241/commits/43f7ccd7fdc6cfdc5a69d9db43ceda554edb481d) , I redrawn the modularization graph. It mainly solves the problem of inconsistency between pictures and texts caused by renaming, and also adopts the painting style according to the [official documents](https://developer.android.com/topic/modularization#what-is-modularization).

modularization-graph


PS: I saved the file in `.drawio.png` format so that others can use draw.io to update the file later.

If you think there is something wrong, please bring it up and I can make adjustments.

PR #242, opened 2022-08-24
Update material3 with built-in insets support

alexvanyo:
Feedback for updated insets support in https://issuetracker.google.com/issues/243713323


dturner (approved):
Thanks. Much easier to read without all that boilerplate.

dturner:
You'll need to rebase onto main before merging because of the `lintDebug` issue.

joreilly:
I was looking to apply changes here to one of my own projects and noticed change in color of bottom bar after making it.....seems like same difference with NowInAndroid app (with these changes)....not sure if intentional?

Screenshot_20220928_160605
Screenshot_20220928_160630

StylianosGakis:
That's very odd, why would the colors change after changing some inset-related code?
Does it by any chance have to do with the new `androidxComposeMaterial3` dependency bump instead? Have you found out what exactly is it that changes the colors?

joreilly:
I think it's related to removal of `Surface(color = MaterialTheme.colorScheme.surface)`

alexvanyo:
> I think it's related to removal of Surface(color = MaterialTheme.colorScheme.surface)

Yup that was the reason. Consulting with our designs, I've updated to match the style we want here. @dturner and @ricknout please take another look!

ricknout (approved):

PR #246, opened 2022-08-30
fix multi drop menu open for topics tags

RobertMaged:
Fix multi drop menu display when two or more topics tags clicked at same moment.

## Before
before-fix

## After
after-fix

### DIFF
--- a/core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
+++ b/core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
@@ -268,7 +268,9 @@ fun NewsResourceTopics(
     topics: List<Topic>,
     modifier: Modifier = Modifier
 ) {
-    var expandedTopic by remember { mutableStateOf<String?>(null) }
+    // Store the ID of the Topic which has its "following" menu expanded, if any.
+    // To avoid UI confusion, only one topic can have an expanded menu at a time.
+    var expandedTopicId by remember { mutableStateOf<String?>(null) }
 
     Row(
         modifier = modifier.horizontalScroll(rememberScrollState()), // causes narrow chips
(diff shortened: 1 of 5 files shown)


core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -268,13 +268,19 @@ fun NewsResourceTopics(
     topics: List<Topic>,
     modifier: Modifier = Modifier
 ) {
+    var expandedTopic by remember { mutableStateOf<String?>(null) }
```
dturner, line 271:
```suggestion
    // Store the ID of the Topic which has its "following" menu expanded, if any. 
    // To avoid UI confusion, only one topic can have an expanded menu at a time. 
    var expandedTopicId by remember { mutableStateOf<String?>(null) }
```

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -268,13 +268,19 @@ fun NewsResourceTopics(
...
+
     Row(
         modifier = modifier.horizontalScroll(rememberScrollState()), // causes narrow chips
         horizontalArrangement = Arrangement.spacedBy(4.dp),
     ) {
         for (topic in topics) {
             NiaTopicTag(
+                expanded = expandedTopic == topic.id,
```
dturner, line 279:
```suggestion
                expanded = expandedTopicId == topic.id,
```

core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -268,13 +268,19 @@ fun NewsResourceTopics(
...
                 followed = true, // ToDo: Check if topic is followed
+                onDropMenuToggle = { show ->
+                    expandedTopic = if (show) topic.id else null
```
dturner, line 282:
```suggestion
                    expandedTopicId = if (show) topic.id else null
```

dturner (changes requested):

dturner (approved):

mmoczkowski:

PR #263, opened 2022-09-04
Add episode_id index to NewsResourceEntity

### LINKED ISSUE #224: KSP warnings related to not indexed foreign key
While building the project, there is a KSP warning related to episode_id not being indexed as explained in this build log:

`
[ksp] C:/Users/User/Desktop/nowinandroid-main/core-database/src/main/java/com/google/samples/apps/nowinandroid/core/database/model/NewsResourceEntity.kt:42: episode_id column references a foreign key but it is not part of an index. This may trigger full table scans whenever parent table is modified so you are highly advised to create an index that covers this column.
`

It can be fixed by adding `episode_id` as an index to indices argument of `NewsResourceEntity` like below:
```kotlin
@Entity(
    tableName = "news_resources",
    foreignKeys = [
        ForeignKey(
            entity = EpisodeEntity::class,
            parentColumns = ["id"],
            childColumns = ["episode_id"],
            onDelete = ForeignKey.CASCADE
        ),
    ],
    indices = [
        Index(value = ["episode_id"])
    ]
)
data class NewsResourceEntity(
```

zohrevand:
Fixes #224

### DIFF
--- a/core-database/schemas/com.google.samples.apps.nowinandroid.core.database.NiaDatabase/10.json
+++ b/core-database/schemas/com.google.samples.apps.nowinandroid.core.database.NiaDatabase/10.json
@@ -1,0 +1,455 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 10,
+    "identityHash": "d90e438da4b3b9cf552df536431dacb3",
+    "entities": [
+      {
+        "tableName": "authors",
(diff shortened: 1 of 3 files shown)


mmoczkowski (approved):
The warning is gone. Thank you!

PR #264, opened 2022-09-04
Add Transaction annotation to queries with populated return type

### LINKED ISSUE #225: KSP warnings related to room queries populated return type not being transactional
While building the project, there are some KSP warnings related to queries with populated return type that marked with `@Relation`. Suggested that it is desired that these methods to be annotated with `@Transaction`

`
[ksp] C:/Users/User/Desktop/nowinandroid-main/core-database/src/main/java/com/google/samples/apps/nowinandroid/core/database/dao/EpisodeDao.kt:36: The return value includes a POJO with a @Relation. It is usually desired to annotate this method with @Transaction to avoid possibility of inconsistent results between the POJO and its relations. See https://developer.android.com/reference/androidx/room/Transaction.html for details.
`

`
[ksp] C:/Users/User/Desktop/nowinandroid-main/core-database/src/main/java/com/google/samples/apps/nowinandroid/core/database/dao/NewsResourceDao.kt:43: The return value includes a POJO with a @Relation. It is usually desired to annotate this method with @Transaction to avoid possibility of inconsistent results between the POJO and its relations. See https://developer.android.com/reference/androidx/room/Transaction.html for details.
`

`[ksp] C:/Users/User/Desktop/nowinandroid-main/core-database/src/main/java/com/google/samples/apps/nowinandroid/core/database/dao/NewsResourceDao.kt:61: The return value includes a POJO with a @Relation. It is usually desired to annotate this method with @Transaction to avoid possibility of inconsistent results between the POJO and its relations. See https://developer.android.com/reference/androidx/room/Transaction.html for details.
`

It can be fixed by adding `@Transaction` annotation to related query methods like:

```kotlin
@Transaction
@Query(
    value = """
        SELECT * FROM news_resources
        ORDER BY publish_date DESC
"""
)
fun getNewsResourcesStream(): Flow<List<PopulatedNewsResource>>
```

zohrevand:
Fixes #225

### DIFF
--- a/core-database/src/main/java/com/google/samples/apps/nowinandroid/core/database/dao/EpisodeDao.kt
+++ b/core-database/src/main/java/com/google/samples/apps/nowinandroid/core/database/dao/EpisodeDao.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.flow.Flow
  */
 @Dao
 interface EpisodeDao {
+    @Transaction
     @Query(value = "SELECT * FROM episodes")
     fun getEpisodesStream(): Flow<List<PopulatedEpisode>>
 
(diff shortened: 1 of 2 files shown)


mmoczkowski (approved):
I can confirm this makes the warning disappear. Thank you for your contribution!

PR #272, opened 2022-09-08
Fix toggle buttons interaction when using TalkBack

calren:


core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/ToggleButton.kt
```diff
@@ -72,7 +72,9 @@ fun NiaToggleButton(
         onClick = { onCheckedChange(!checked) },
         modifier = modifier
             .size(size)
-            .toggleable(value = checked, enabled = enabled, role = Role.Button, onValueChange = {})
+            .toggleable(value = checked, enabled = enabled, role = Role.Button, onValueChange = {
+                onCheckedChange(!checked)
```
alexvanyo, line 76: Nit: Should this use the parameter of `onValueChange` instead of `checked`?

Probably doesn't really matter, they should do the same thing.

alexvanyo (approved):

core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/ToggleButton.kt
calren: yeah, i think it's essentially the same. will keep it as is!

PR #281, opened 2022-09-13
Update ktlint and spotless

dturner:
Manual fix for #259


dturner:
Ktlint can be updated to 0.45.1 without issue, but when using next version 0.46.0 we get the error: 

```
Execution failed for task ':feature:foryou:spotlessKotlin'.
> com/pinterest/ktlint/core/KtLint$Params
```

With ktlint on 0.45.1, updating Spotless to 6.8.0 gives the following warning as an error: 

```
UserData contains properties [android]. However, userData is deprecated and will be removed in a future version. Please create an issue that shows how this field is actively used.
```

dturner:
Redone in #423

PR #288, opened 2022-09-16
Remove Search from top app bar

calren:
We can reintroduce the search icon on the top app bar when we implement the functionality.


dturner (approved):
Thanks Caren! LGTM.

PR #294, opened 2022-09-19
Change user data fields to use maps instead of lists.

dturner:
When adding a single bookmark, the previous implementation of `NiaPreferencesDataSource` would read the whole list of bookmarks from user preferences, modify it, then write it back. 

This PR changes the proto data type for user data from `repeated string` (a list) to `map<string,bool>` which allows the reading and writing of single values. 

The benefits of this are: 

- it's more efficient
- it makes the read/write code easier to understand, negating the need for the `setList` and `editList` extension functions, which I found quite difficult to parse

The downsides are: 

- There is redundancy in the map with all values being set to `true`. I wanted a set but proto3 doesn't have a set type so this is the closest I could get. 

Comments/feedback welcome.

### DIFF
--- a/core/datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
+++ b/core/datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
@@ -18,8 +18,6 @@ package com.google.samples.apps.nowinandroid.core.datastore
 
 import android.util.Log
 import androidx.datastore.core.DataStore
-import com.google.protobuf.kotlin.DslList
-import com.google.protobuf.kotlin.DslProxy
 import com.google.samples.apps.nowinandroid.core.model.data.UserData
 import java.io.IOException
 import javax.inject.Inject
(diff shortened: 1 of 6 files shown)


core/datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
```diff
@@ -33,54 +31,89 @@ class NiaPreferencesDataSource @Inject constructor(
...
-        userPreferences.setList(
-            listGetter = { it.followedTopicIds },
-            listModifier = { followedTopicIds.toList() },
-            clear = { it.clear() },
-            addAll = { dslList, editedList -> dslList.addAll(editedList) }
-        )
+    suspend fun setFollowedTopicIds(followedTopicIds: Set<String>) {
+        try {
+            userPreferences.updateData {
+                with(it.toBuilder()) {
```
alexvanyo, line 43: Nit: There's a `copy` block here that does the same thing as `.toBuilder()` and `.build()` that I think looks slightly nicer?
```kotlin
it.copy {
    followedTopicIdsMap.clear()
    followedTopicIdsMap.putAll(followedTopicIds.associateWith { true })
}
```

core/datastore/src/main/proto/com/google/samples/apps/nowinandroid/data/user_preferences.proto
```diff
@@ -28,7 +28,11 @@ message UserPreferences {
...
     repeated int32 deprecated_int_followed_author_ids = 7;
     bool has_done_int_to_string_id_migration = 8;
-    repeated string followed_topic_ids = 9;
-    repeated string followed_author_ids = 10;
-    repeated string bookmarked_news_resource_ids = 11;
+    repeated string deprecated_followed_topic_ids = 9;
+    repeated string deprecated_followed_author_ids = 10;
+    repeated string deprecated_bookmarked_news_resource_ids = 11;
+    bool has_done_list_to_map_migration = 12;
+    map<string, bool> followed_topic_ids = 13;
```
alexvanyo, line 35: I think these should be commented with a note that the value `bool` here has no meaning, and to treat them as just a set.

core/datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
dturner: Probably me missing something obvious but when I try: 

```
it.copy { 
    followedTopicIdsMap.clear()
    followedTopicIdsMap.putAll(followedTopicIds.associateWith { true })
}
```

I get an `Collections$UnmodifiableMap` error from the following unit test `offlineFirstUserDataRepository_set_followed_topics_logic_delegates_to_nia_preferences`. 

I had to change it to: 

```
userPreferences.updateData {
                with(it.toBuilder()) {
                    followedTopicIdsMap.clear()
                    followedTopicIdsMap.putAll(followedTopicIds.associateWith { true })
                    build()
                }
            }
```

to get it to pass.

core/datastore/src/main/proto/com/google/samples/apps/nowinandroid/data/user_preferences.proto
dturner: Thanks! Comment added to this effect.

core/datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
alexvanyo: Oh, maybe it needs to be this?

```
it.copy {
    this.followedTopicIds.clear()
    this.followedTopicIds.putAll(followedTopicIds.associateWith { true })
}
```

The `this` there is since the name in the `copy` DSL overlaps with the parameter

tunjid (approved):
LGTM, definitely more readable.

core/datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
```diff
@@ -33,54 +31,85 @@ class NiaPreferencesDataSource @Inject constructor(
     val userDataStream = userPreferences.data
         .map {
             UserData(
-                bookmarkedNewsResources = it.bookmarkedNewsResourceIdsList.toSet(),
-                followedTopics = it.followedTopicIdsList.toSet(),
-                followedAuthors = it.followedAuthorIdsList.toSet(),
+                bookmarkedNewsResources = it.bookmarkedNewsResourceIdsMap.keys,
```
tunjid, line 34: Won't this take all the keys? Should it be filtering on the values as well?

dturner: Yes, it'll take all the keys. No, it shouldn't filter on the values because the value irrelevant (it's always `true`). The presence of a key in the map indicates that it's a bookmark. 

From `user_preferences.proto`: 

```
// Each map is used to store a set of string IDs. The bool has no meaning, but proto3 doesn't
    // have a Set type so this is the closest we can get to a Set.
    map<string, bool> followed_topic_ids = 13;
    map<string, bool> followed_author_ids = 14;
    map<string, bool> bookmarked_news_resource_ids = 15;
```

I chose `bool` as the value type because it has the smallest size. 

Should I document this elsewhere as well? If it wasn't obvious to you, it's likely not obvious to others.

tunjid: Gotcha!

PR #304, opened 2022-09-26
Offline-first UX improvements

tunjid:
offline

This CL improves the offline-first story by clearly communicating to the user:

- When offline on the ForYouScreen
- When the ForYouScreen is fetching data again as a result of sync kicking in

It also adds two utilities:

- A `NetworkMonitor` for checking when online
- A `SyncStatusMonitor` for checking when sync is running.

Implementations delegate to `ConnectivityManager` and `WorkManager` respectively.


core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -0,0 +1,77 @@
...
+            override fun onAvailable(network: Network) {
+                channel.trySend(true)
+            }
+
+            override fun onLost(network: Network) {
+                channel.trySend(false)
+            }
+        }
+
+        val connectivityManager = context.getSystemService(
```
alexvanyo, line 47: Nit: there's an `androidx.core` version of this that I like that allows for

```
context.getSystemService<ConnectivityManager>()
```

core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -0,0 +1,77 @@
...
+                .build(),
+            callback
+        )
+
+        channel.trySend(connectivityManager.isCurrentlyConnected())
+
+        awaitClose {
+            connectivityManager?.unregisterNetworkCallback(callback)
+        }
+    }
```
alexvanyo, line 63: Nit: We could consider adding `.conflate()` to the end here, since I would guess we should only care about the most recent value (and therefore `trySend` should never fail)

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
```diff
@@ -148,15 +160,20 @@ class ForYouViewModel @Inject constructor(
...
+            combine(
+                snapshotFlow { inProgressTopicSelection },
+                snapshotFlow { inProgressAuthorSelection },
+                ::Pair,
+            )
+        ) { isSyncing, followedInterestsUserState, availableTopics, availableAuthors,
+            (inProgressTopicSelection, inProgressAuthorSelection) ->
 
-            when (followedInterestsUserState) {
+            if (isSyncing) ForYouInterestsSelectionUiState.Loading
```
alexvanyo, line 175: Nit: I find the single line `if`/`else` a bit harder to read here, over using brackets.

sync/work/src/main/java/com/google/samples/apps/nowinandroid/sync/status/WorkManagerSyncStatusMonitor.kt
```diff
@@ -0,0 +1,62 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.callbackFlow
+
+/**
+ * [SyncStatusMonitor] backed by [WorkInfo] from [WorkManager]
+ */
+class WorkManagerSyncStatusMonitor @Inject constructor(
+    @ApplicationContext context: Context
+) : SyncStatusMonitor {
+    override val isSyncing: Flow<Boolean> = callbackFlow {
```
alexvanyo, line 40: Can this be simplified with `asFlow` from `lifecycle-livedata-ktx`?

```suggestion
    override val isSyncing: Flow<Boolean> =
        WorkManager.getInstance(context)
            .getWorkInfosForUniqueWorkLiveData(SyncWorkName)
            .asFlow()
            .map { it.anyRunning }
```

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
alexvanyo, line 175: I'm slightly concerned about this logic.

If we currently have data, shouldn't we be showing that, even if we are syncing in the background?

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
```diff
@@ -174,7 +191,7 @@ class ForYouViewModel @Inject constructor(
                     }
 
                     if (topics.isEmpty() && authors.isEmpty()) {
-                        ForYouInterestsSelectionUiState.Loading
+                        ForYouInterestsSelectionUiState.LoadFailed
```
alexvanyo, line 194: Maybe the logic should be this (with no `if` above):
```suggestion
                        if (isSyncing) {
                            ForYouInterestsSelectionUiState.Loading
                        } else {
                            ForYouInterestsSelectionUiState.LoadFailed
                        }
```

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
tunjid: Great point, I did want to show feedback to the user when loading even though there's data however. I've added a `isLoading` field for `ForYouInterestsSelectionUiState.WithInterestsSelection` to achieve this.

tunjid: Dropped as per the comment below.

core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
tunjid: Done, why `conflate()` over `distinctUntilChanged()`?

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
tunjid: Applied and added a loading field to `ForYouInterestsSelectionUiState.WithInterestsSelection`. Would you model this differently? Is it weird to have two different kinds of loading states?

tunjid: Resolved in the comment above.

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
tunjid: I ended up exposing the `isSyncing` field separately and de-duping the loading spinners in the interests and news components. The loading spinner is now owned by the `ForYouScreen` (and subsequently the `BookmarksScreen`) as a whole.

The spinner now shows:
On the `ForYouScreen` if sync is in progress or either interests or news are loading.
On the `BookmarksScreen` is news are loading. A future PR can neatly establish loading if sync is in progress here as well.

core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
alexvanyo: I think `distinctUntilChanged()` doesn't change the backpressure like `conflate` does. If we had a really slow collector, and were emitting `true`, `false`, `true`, `false`, etc., then `distinctUntilChanged` wouldn't have any effect.

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -205,17 +242,13 @@ fun ForYouScreen(
...
-                    showLoadingUIIfLoading =
-                    interestsSelectionState !is ForYouInterestsSelectionUiState.Loading,
                     onNewsResourcesCheckedChanged = onNewsResourcesCheckedChanged,
-                    loadingContentDescription = R.string.for_you_loading
                 )
 
                 item(span = { GridItemSpan(maxLineSpan) }) {
                     Column {
                         Spacer(modifier = Modifier.height(8.dp))
+                        if (isOffline) Spacer(modifier = Modifier.height(48.dp))
```
alexvanyo, line 251: Just confirming, this is to add extra space for the snackbar?

tunjid: Yeah, too magic number-y?

alexvanyo: A little bit, but I don't really see a better solution. Maybe just add a comment saying what there's extra space?

alexvanyo (approved):
New loading logic looks good to me!

core/designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/LoadingWheel.kt
```diff
@@ -142,5 +161,22 @@ fun NiaLoadingWheelPreview() {
     }
 }
 
+@Preview(
```
alexvanyo, line 164: Small nit: We have a `@ThemePreviews` private to `core/designsystem`, would be great to reuse that here.

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -198,31 +220,43 @@ fun ForYouScreen(
...
             item(span = { GridItemSpan(maxLineSpan) }) {
                 Column {
                     Spacer(modifier = Modifier.height(8.dp))
+                    // Add space for the content to clear the "offline" snackbar.
+                    if (isOffline) Spacer(modifier = Modifier.height(48.dp))
                     Spacer(Modifier.windowInsetsBottomHeight(WindowInsets.safeDrawing))
                 }
             }
         }
+        if (
```
alexvanyo, line 235: Nit: Could be nice to have some version of `AnimatedVisibility` here to animate appearing/disappearing.

alexvanyo (approved):
Looks good, just some minor nits

PR #307, opened 2022-09-29
Opt in to experimental kotlinx.serialization API locally

zsmb13:
Opt in to using the experimental `asConverterFactory` function locally instead of enabling experimental APIs module-wide


alexvanyo (approved):
Looks good, thanks for fixing that warning!

PR #311, opened 2022-10-02
Use @Upsert from Room instead of a custom helper

### LINKED ISSUE #243: [FR] Use @Upsert from Room
Version [`2.5.0-alpha03`](https://developer.android.com/jetpack/androidx/releases/room#2.5.0-alpha03) of Room added an `@Upsert` annotation, which we probably should be able to use instead of our own custom upsert helper.

yveskalume:
Fix #243


core/database/src/main/java/com/google/samples/apps/nowinandroid/core/database/dao/AuthorDao.kt
```diff
@@ -56,12 +56,8 @@ interface AuthorDao {
     /**
      * Inserts or updates [entities] in the db under the specified primary keys
      */
-    @Transaction
-    suspend fun upsertAuthors(entities: List<AuthorEntity>) = upsert(
```
alexvanyo, line 60: I believe we should be able to remove the now-unused `upsert` method as part of this PR.

alexvanyo (commented):
Thanks for contributing! One request to remove the code that will be unused as part of this, and also to fix up the tests that are failing.

alexvanyo (commented):
Thanks for contributing! One request to remove the code that will be unused as part of this, and also to fix up the tests that are failing.

yveskalume:
> Thanks for contributing! One request to remove the code that will be unused as part of this, and also to fix up the tests that are failing.

thanks for the review, what should i put in the tests ?

tunjid (approved):
Thanks for doing this! It looks great! Just waiting on CI to pass.

PR #318, opened 2022-10-03
Use gradle version catalog's type-safe accessors

### LINKED ISSUE #317: [FR]: Use gradle version catalog's type-safe accessors in `build-logic`

### Describe the problem

Currently, `build-logic` uses the gradle version catalog without type-safe accessors and the names of libraries and plugins are hard-coded.

For example, the `AndroidFeatureConventionPlugin` has the following code snippet:

```kt
dependencies {
    add("implementation", libs.findLibrary("androidx.hilt.navigation.compose").get())
    add("implementation", libs.findLibrary("androidx.lifecycle.runtimeCompose").get())
    add("implementation", libs.findLibrary("androidx.lifecycle.viewModelCompose").get())
}
```

### Describe the solution

With the gradle version catalog's type-safe accessors, the above code snippet can be replaced with:

```kt
dependencies {
    add("implementation", libs.androidx.hilt.navigation.compose)
    add("implementation", libs.androidx.lifecycle.runtimeCompose)
    add("implementation", libs.androidx.lifecycle.viewModelCompose)
}
```

### Additional context

_No response_

### Code of Conduct

MaximillianLeonov:
Closes #317

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -28,6 +28,7 @@ java {
 dependencies {
     compileOnly(libs.android.gradlePlugin)
     compileOnly(libs.kotlin.gradlePlugin)
+    compileOnly(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
 }
 
 gradlePlugin {
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -28,6 +28,7 @@ java {
 dependencies {
     compileOnly(libs.android.gradlePlugin)
     compileOnly(libs.kotlin.gradlePlugin)
+    // Workaround for https://github.com/gradle/gradle/issues/15383
     compileOnly(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
 }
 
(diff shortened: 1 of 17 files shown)


build-logic/convention/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ java {
 dependencies {
     compileOnly(libs.android.gradlePlugin)
     compileOnly(libs.kotlin.gradlePlugin)
+    compileOnly(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
```
alexvanyo, line 31: Is this the main line that makes this possible?

Could you provide a link to documentation for how this works? At first glance, it looks a little magical and brittle.

MaximillianLeonov: This is the workaround for https://github.com/gradle/gradle/issues/15383#issuecomment-779893192.

wojtek-kalicinski: this whole setup is probably too hacky and brittle to show in a sample, as Alex said

I would gladly accept this when it's fully supported by Gradle, it would be great to have proper catalog support in the build-logic

alexvanyo:
Thanks for the proposal! Closing at the moment as this is a workaround that we probably don't want to recommend broadly, and could be brittle with different versions of Gradle.

Once Gradle has full support this would be great to add!

PR #321, opened 2022-10-04
Remove the AndroidX snapshot repo from the list of repositories

zsmb13:
This repo is currently unused and every dependency fetch tries to go to this repo first as it's the first in the list (even though none will be found here) which adds several seconds to build times


alexvanyo (approved):
LGTM!

PR #326, opened 2022-10-07
Remove redundant label on activity

### LINKED ISSUE #325: [FR]: Redundant label on activity can be removed

### Describe the problem

The `AndroidManifest` has the following code snippet:
```xml
<activity
    android:name=".MainActivity"
    android:exported="true"
    android:label="@string/app_name">
```
When an activity does not have a label attribute, it will use the one from the application tag. Since the application has already specified the same label, the label on the activity can be omitted.

### Describe the solution

Remove redundant label on activity.

### Additional context

_No response_

### Code of Conduct

MaximillianLeonov:
Closes #325.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -32,8 +32,7 @@
 
         <activity
             android:name=".MainActivity"
-            android:exported="true"
-            android:label="@string/app_name">
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 


tunjid:
Unless I'm missing something, the `label` [doesn't seem unnecessary](https://developer.android.com/guide/topics/manifest/activity-element#label) as it's on the main launcher `Activity`.

I'm closing this, but feel free to reopen if the above isn't true.

tunjid:
Reopening bc I re-read the description in the issue you opened. I'll find out what the best practice is on this and let you know.

MaximillianLeonov:
> Reopening bc I re-read the description in the issue you opened. I'll find out what the best practice is on this and let you know.

Thanks. I would like to mention that there is a `RedundantLabel` lint rule for this and has a 5/10 warning priority.

dturner:
Thanks for spotting and fixing this. I can confirm that the lint check in Android Studio states that this label is redundant.

PR #333, opened 2022-10-12
Migrate packages to namespaces in build files

wojtek-kalicinski:
Migrate packages to namespaces in build files to prepare for AGP 8 breaking changes and show best practice

### DIFF
--- a/core/common/build.gradle.kts
+++ b/core/common/build.gradle.kts
@@ -19,6 +19,10 @@ plugins {
     id("nowinandroid.android.hilt")
 }
 
+android {
+    namespace = "com.google.samples.apps.nowinandroid.core.common"
+}
+
 dependencies {
     implementation(libs.kotlinx.coroutines.android)
     testImplementation(project(":core:testing"))
--- a/core/common/src/main/AndroidManifest.xml
+++ b/core/common/src/main/AndroidManifest.xml
@@ -14,7 +14,6 @@
      See the License for the specific language governing permissions and
      limitations under the License.
 -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.google.samples.apps.nowinandroid.core.common">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
 </manifest>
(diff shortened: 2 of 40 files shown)


core/common/build.gradle.kts
```diff
@@ -22,4 +22,7 @@ plugins {
 dependencies {
     implementation(libs.kotlinx.coroutines.android)
     testImplementation(project(":core:testing"))
-}
\ No newline at end of file
+}
+android {
```
alexvanyo, line 26: Nit: should we place the `android` block above `dependencies` to stay consistent?

alexvanyo (approved):

core/common/build.gradle.kts
wojtek-kalicinski: yeah probably, I used an automatic tool to make the changes but I'll go and change that

core/common/build.gradle.kts
```diff
@@ -19,6 +19,10 @@ plugins {
     id("nowinandroid.android.hilt")
 }
 
+android {
+    namespace = "com.google.samples.apps.nowinandroid.core.common"
+}
```
MaximillianLeonov, line 24: Nit: the `android` block can be omitted.
```kts
android.namespace = "com.google.samples.apps.nowinandroid.core.common"
```

core/common/src/main/AndroidManifest.xml
```diff
@@ -14,7 +14,6 @@
      See the License for the specific language governing permissions and
      limitations under the License.
 -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.google.samples.apps.nowinandroid.core.common">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
 </manifest>
```
MaximillianLeonov, line 19: Nit: the unused android namespace can be omitted and the `manifest` tag can be collapsed.
```xml
<manifest />
```

core/common/build.gradle.kts
wojtek-kalicinski: Done.

core/common/build.gradle.kts
wojtek-kalicinski: I prefer to always use blocks in case more properties are added in the future, then someone would need to refactor this into a block

PR #316, opened 2022-10-03
Update Baseline Profile generation

keyboardsurfer:

PR #337, opened 2022-10-14
Regenerate profile after some refactoring

mlykotom:
The current profile contained classes for `NiaApp`, which doesn't exist anymore.
Just regenerated profile and applied.
No need to check too much 👀 . 
Just ✅ 

We will automated this process in the future.


keyboardsurfer:
How does this relate to #316 ?

mlykotom:
Oh, I missed it .. just realised it's not generated, so I did without further research

PR #348, opened 2022-10-21
Remove unnecessary dependencies after releasing Android Studio Dolphin

fahime-ghasemi:
I cloned the codes today but I've got errors for rendering previews. I couldn't see any previews for ForYouScreen or other pages. I found that the problem is related to these dependencies that aren't needed in Android Studio Dolphin. I removed them and all previews shows correctly without any problem.


MaximillianLeonov:
Duplicate of #297.

mmoczkowski:
Thank you for reporting, the #297 was merged. That should fix the issue.

PR #356, opened 2022-10-24
Update to new Compose release

JolandaVerhoef:


build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -42,6 +43,12 @@ internal fun Project.configureAndroidCompose(
         kotlinOptions {
             freeCompilerArgs = freeCompilerArgs + buildComposeMetricsParameters()
         }
+
+        dependencies {
+            val bom = libs.findLibrary("androidx-compose-bom").get()
+            add("implementation", platform(bom))
```
alexvanyo, line 49: ❤️  it!

So by specifying the platform version here in the convention plugin, we can pull in Compose dependencies individually with the version taken from the bom, right? That's super cool!

alexvanyo (approved):
Looks great!

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
JolandaVerhoef: Yes! And if any module needs to override the version (e.g. wants to use an alpha version) it can just explicitly add that dependency and it will override what's provided in the BOM.

core/testing/build.gradle.kts
```diff
@@ -15,6 +15,7 @@
  */
 plugins {
     id("nowinandroid.android.library")
+    id("nowinandroid.android.library.compose")
```
dturner, line 18: Why does the testing module need the compose plugin?

dturner: Ah, because it has Compose dependencies which use the version provided by the BOM 👍

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
dturner: Top work, thanks for doing this.

PR #365, opened 2022-10-26
Resolves #206, enables predictive back navigation support

### LINKED ISSUE #206: [FR]: Support predictive back navigation

### Describe the problem

Support Android 13 predictive back navigation

### Describe the solution

Opt-in to predictive back navigation as specified here: https://developer.android.com/about/versions/13/features/predictive-back-gesture

### Additional context

We are using the AndroidX APIs (via `navigation-compose`/`BackHandler`), so this should require no additional work except for enabling the feature.

### Code of Conduct

YuraLaguta:
Demo:

https://user-images.githubusercontent.com/1853043/197951024-b6e34662-e0a9-447a-89be-06ff4b2cee39.mp4

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -22,6 +22,7 @@
     <application
         android:name=".NiaApplication"
         android:allowBackup="true"
+        android:enableOnBackInvokedCallback="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"


app/build.gradle.kts
```diff
@@ -101,6 +101,7 @@ dependencies {
     debugImplementation(libs.androidx.compose.ui.testManifest)
 
     implementation(libs.accompanist.systemuicontroller)
+    implementation(libs.androidx.activity.activity)
```
alexvanyo, line 104: I think this import should be unnecessary, since we already depend on `activity-compose` below. Can we avoid the new import?

alexvanyo (approved):
Looks great, thanks for the contribution!

mmoczkowski:
@YuraLaguta Thank you for you contribution, merging :)

PR #375, opened 2022-10-27
Attempt to fix CI from erroring on emulator API 23

dturner:
Another attempt to fix failing CI.


dturner:
This results in `Error: Timeout waiting for emulator to boot.`

JolandaVerhoef (approved):

dturner:
After looking at the code, `force-avd-creation: true` is the same as the default behaviour so will have no effect.

PR #385, opened 2022-11-04
Store run configuration for generating demo flavor Baseline Profile

mlykotom:
Store the gradle task that generates Baseline Profiles for demo flavor in the project run configurations, which can help with discovery of the feature.
image

Currently it uses `--rerun-tasks` that ommits any cached tasks, because the task itself is cacheable and therefore skips running if ran multiple times.

But once Gradle wrapper 7.6 is stable (currently 7.6-rc1), we can use `--rerun` that would only rerun the current task without the need to rebuild the whole project.


.run/Generate Demo Baseline Profile.run.xml
```diff
@@ -0,0 +1,23 @@
+<component name="ProjectRunConfigurationManager">
```
JolandaVerhoef, line 1: Add copyright notice?

.run/Generate Demo Baseline Profile.run.xml
```diff
@@ -0,0 +1,23 @@
...
+  <configuration default="false" name="Generate Demo Baseline Profile" type="GradleRunConfiguration" factoryName="Gradle">
```
JolandaVerhoef, line 2: Can we add a comment pointing to documentation about baseline profiles, to aid in discoverability?

.run/Generate Demo Baseline Profile.run.xml
```diff
@@ -0,0 +1,23 @@
...
+    <ExternalSystemSettings>
+      <option name="executionName" />
+      <option name="externalProjectPath" value="$PROJECT_DIR$" />
+      <option name="externalSystemIdString" value="GRADLE" />
+      <option name="scriptParameters" value="--rerun-tasks -Pandroid.testInstrumentationRunnerArguments.androidx.benchmark.enabledRules=BaselineProfile" />
```
JolandaVerhoef, line 7: Can you add a comment regarding the update needed after library goes stable?

dturner (approved):

PR #394, opened 2022-11-04
Sort authors in onboarding by number of news resources

JolandaVerhoef:


dturner:
Waiting on decision about whether to remove "Follow authors"

core/domain/src/main/java/com/google/samples/apps/nowinandroid/core/domain/GetSortedFollowableAuthorsStreamUseCase.kt
```diff
@@ -17,33 +17,40 @@
...
 ) {
     /**
      * Returns a list of authors with their associated followed state sorted alphabetically by name.
      *
-     * @param followedTopicIds - the set of topic ids which are currently being followed.
+     * @param followedAuthorIds - the set of topic ids which are currently being followed.
      */
     operator fun invoke(followedAuthorIds: Set<String>): Flow<List<FollowableAuthor>> {
-        return authorsRepository.getAuthorsStream().map { authors ->
+        return combine(
```
tunjid, line 40: We should probably do this in SQL in the `AuthorDao`. The `news_resources_authors` table has all the information we need to make this query efficient.

```
SELECT * FROM authors
WHERE id in (
    SELECT  author_id
    FROM news_resources_authors
    GROUP BY author_id
    ORDER BY COUNT(author_id) DESC;
)
```

tunjid: The onboarding can then have its own model and each author can have a little badge showing the amount of articles they've written.

dturner:
Closing since we removed authors.

PR #406, opened 2022-11-10
Save interests during onboarding process

dturner:
This PR changes the onboarding behaviour and code in the following ways:

- Interests (authors and topics) are immediately saved. This fixes b/239425922. 
- Removed `GetPersistentSortedFollowableAuthorsUseCase` and refactored other use cases since there's no longer any need for a transient set of followed author or topic ids. This simplifies the domain layer. 
- The onboarding is no longer automatically dismissed when the user selects interests on the Interests tab. It stays visible until the user taps the "Done" button. This is achieved with a new user property: `hasDismissedOnboarding`. The default value is `false`. It's set to `true` when the user dismisses the onboarding and is reset to `false` if the user ever removes all their followed authors and topics. The logic for this is placed in `NiaPreferencesDataSource`.


core/datastore/src/main/proto/com/google/samples/apps/nowinandroid/data/user_preferences.proto
```diff
@@ -43,4 +43,9 @@ message UserPreferences {
 
     ThemeBrandProto theme_brand = 16;
     DarkThemeConfigProto dark_theme_config = 17;
+
+    // Note that proto3 only allows a default value of `false` for boolean types. This means that
+    // whilst the name `should_show_onboarding` would be preferable here we are forced to choose a
+    // name which works with this default value constraint.
+    bool has_dismissed_onboarding = 18;
```
tunjid, line 50: Would `shouldHideOnboarding` work as well?

tunjid (approved):
LGTM, just the small request for the field name change if possible.

core/datastore/src/main/proto/com/google/samples/apps/nowinandroid/data/user_preferences.proto
dturner: Yeah, it's still not ideal (`hide` is harder to understand than `show`), but better than `has_dismissed` which, to quote adamp, is an "event masquerading as state". 

Code updated. Thanks for the review 👍

core/datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/NiaPreferencesDataSource.kt
```diff
@@ -188,4 +193,19 @@ class NiaPreferencesDataSource @Inject constructor(
...
+    suspend fun setShouldHideOnboarding(shouldHideOnboarding: Boolean) {
+        userPreferences.updateData {
+            it.copy {
+                this.shouldHideOnboarding = shouldHideOnboarding
+            }
+        }
+    }
+}
+
+fun UserPreferencesKt.Dsl.updateShouldHideOnboardingIfNecessary() {
```
tunjid, line 206: Minor, this should probably be private since it should only be called atomically.

PR #413, opened 2022-11-10
Enable Gradle Scan on CI

SimonMarquis:
Gradle Scans are currently not enabled on CI.
This means we can't easily diagnose CI build issue.

For instance: https://github.com/android/nowinandroid/actions/runs/3440519227#summary-9422900516
image


dturner:
This looks great, but we're going to need to review the enterprise license for using this tool before approving this PR.

SimonMarquis:
This makes perfect sense.
Hopefuly we can have something similar to https://github.com/androidx/androidx (they are currently using their own instance: https://ge.androidx.dev).

mmoczkowski:
@SimonMarquis Thank you for your contribution but we're not going to use Gradle Scan for the time being.

PR #417, opened 2022-11-11
Migrate hard-coded JSON data into JSON assets files

### LINKED ISSUE #391: [FR]: Have `FakeDataSource` load from JSON files rather than hardcoded data

### Describe the problem

Hardcoding JSON data into `FakeDataSource` makes it difficult to update and not very readable. It'd be better to have the JSON data stored as assets or resources in the app, then read and parsed by an `AssetsDataSource` or similar. 

Context: https://github.com/android/nowinandroid/pull/387#discussion_r1014004896

### Describe the solution

- Create JSON files for topics, authors and news resources (this is the output of the spreadsheet script)
- Store these files as resources or assets in the app package
- Create a data source which can read and parse these files

### Additional context

_No response_

### Code of Conduct

SimonMarquis:
JSON minification is no longer necessary since assets files are already gzipped during APK/AAB packaging.

Fixes #391


SimonMarquis:
This solution tries to modify as less code as possible.
It requires enabling `isIncludeAndroidResources` instead of overriding sourceSets between Android and JVM.

mmoczkowski:
@SimonMarquis Thank you for your contribution Simon! Great idea, I was meaning to do this myself. Could you fix the failing test please? Thanks again!

mmoczkowski (changes requested):
Could you address the failing tests and the conflict please? Thanks!

mmoczkowski (approved):

mmoczkowski:
Thank you for your contribution Simon!

PR #418, opened 2022-11-11
Fix typos

SimonMarquis:


dturner (approved):
Thanks very much for spotting and fixing these 💚

PR #423, opened 2022-11-11
Update ktlint version

dturner:


gradle/init.gradle.kts
```diff
@@ -14,7 +14,7 @@
  *   limitations under the License.
  */
 
-val ktlintVersion = "0.43.0"
+val ktlintVersion = "0.46.1"
```
anticafe, line 17: According to [Spotless v6.11.0](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#6110---2022-09-14), its default Ktlint version is `0.47.1`. Why don't we just use Ktlint default version instead of using `0.46.1`?

mmoczkowski: Any idea @dturner ?

mmoczkowski:
Closing due to inactivity

PR #427, opened 2022-11-12
Handling empty bookmarks on Saved view

LinX64:
Previously, the bookmark screen was not showing anything when there were no available bookmarks. This PR, adds the handling of an empty list, shows a Text and notifies the user about the empty list.

PR #443, opened 2022-11-17
New empty bookmarks handling

LinX64:
**This is a new PR for the saved screen. Replacement for [this PR](https://github.com/android/nowinandroid/pull/427).**

The output after the code review and replaced vector:


feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -102,18 +110,65 @@ fun BookmarksScreen(
...
+    /**
+     ** The [Column] is handling the Empty state when the feed is empty.
+     **/
+    if (feedState is NewsFeedUiState.Success && feedState.feed.isEmpty()) {
+        OnEmptyStateView()
+    }
 }
+
+@Composable
+private fun OnEmptyStateView() {
```
JolandaVerhoef, line 127: A Composable function should not have an `on` name, consider renaming to "EmptyState()"

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -102,18 +110,65 @@ fun BookmarksScreen(
...
+    Column(
+        modifier = Modifier
+            .padding(start = 16.dp, end = 16.dp)
```
JolandaVerhoef, line 130: Why not padding all around? `padding(16.dp)`

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -102,18 +110,65 @@ fun BookmarksScreen(
...
+            .fillMaxSize()
+            .testTag("bookmarks:empty"),
+        verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+        Image(
+            modifier = Modifier
+                .fillMaxWidth(),
+            painter = painterResource(id = R.drawable.img_no_bookmarks_error),
+            contentDescription = stringResource(id = R.string.bookmarks_empty_description),
```
JolandaVerhoef, line 140: Since you also add text, you can set the content description of the image to null (it's purely decorative)

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -102,18 +110,65 @@ fun BookmarksScreen(
...
+        )
+
+        Spacer(modifier = Modifier.height(16.dp))
+
+        Text(
+            text = stringResource(id = R.string.bookmarks_empty_error),
+            textAlign = TextAlign.Center,
+            modifier = Modifier
+                .fillMaxWidth(),
```
JolandaVerhoef, line 149: move to previous line

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -102,18 +110,65 @@ fun BookmarksScreen(
...
+        modifier = Modifier
+            .padding(start = 16.dp, end = 16.dp)
+            .fillMaxSize()
+            .testTag("bookmarks:empty"),
+        verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+        Image(
+            modifier = Modifier
+                .fillMaxWidth(),
```
JolandaVerhoef, line 138: move to previous line

JolandaVerhoef (changes requested):
I'll have to look into the png > svg transformation, will get back to you on that!

JolandaVerhoef:
You can export the greyscale icon instead, that imports successfully. Then you can use the tint parameter to give it the right themed color!

LinX64:
> You can export the greyscale icon instead, that imports successfully. Then you can use the tint parameter to give it the right themed color!

Amazing! thanks so much, Jolanda. 

Just one more thing, I am not sure if you already noticed, but there was also some shadow behind that icon which probably caused that converting issue. Anyways, I hope we can move forward without considering that shadow. I've updated the PR description after code reviewing and using the vector you mentioned, will also push the new commits soon enough. Thanks a lot! :)

LinX64:
Not sure what was the issue with Spotless, however, here is the error:

```
> The following files had format violations:
      src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
          @@ -156,6 +156,5 @@
           ············textAlign·=·TextAlign.Center,
           ············style·=·MaterialTheme.typography.bodyMedium
           ········)
          -
           ····}
          -}
          +}
```

hoc081098:
> Not sure what was the issue with Spotless, however, here is the error:
> 
> ```
> > The following files had format violations:
>       src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
>           @@ -156,6 +156,5 @@
>            ············textAlign·=·TextAlign.Center,
>            ············style·=·MaterialTheme.typography.bodyMedium
>            ········)
>           -
>            ····}
>           -}
>           +}
> ```
> 
> Any hints would be appreciated.

Run `gradlew spotlessApply`

LinX64:
> > Not sure what was the issue with Spotless, however, here is the error:
> > ```
> > > The following files had format violations:
> >       src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
> >           @@ -156,6 +156,5 @@
> >            ············textAlign·=·TextAlign.Center,
> >            ············style·=·MaterialTheme.typography.bodyMedium
> >            ········)
> >           -
> >            ····}
> >           -}
> >           +}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any hints would be appreciated.
> 
> Run `./gradlew spotlessApply`

It's almost similar to the `'./gradlew :feature:bookmarks:spotlessApply'` command which Gradle mentioned already and It didn't work. Check the errors:

```
* What went wrong:
Task 'spotlessApply' not found in root project 'nowinandroid'.
```

```
* What went wrong:
Task 'spotlessApply' not found in project ':feature:bookmarks'.
```
The question is, do I need to include something from Spotless on my computer? Or perhaps it's a bug?

I also removed the space for the last line before the curly bracket, but still there seems to be an issue in there.

hoc081098:
@LinX64 
Try it
```
./gradlew spotlessApply --init-script gradle/init.gradle.kts --no-configuration-cache
```

JolandaVerhoef:
I'm rewriting the BookmarksScreen file a bit to be more readable, will push to this branch and would love to hear your thoughts!

LinX64:
> @LinX64 I see a command in workflow file, and change `check` to `apply`
> 
> ```
> ./gradlew spotlessApply --init-script gradle/init.gradle.kts --no-configuration-cache
> ```

It worked, thanks!

JolandaVerhoef:
Can you allow me to push to your fork so I can push my changes? See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -61,8 +70,29 @@ internal fun BookmarksRoute(
...
+    removeFromBookmarks: (String) -> Unit,
+    modifier: Modifier = Modifier
+) {
+    when (feedState) {
+        Loading -> LoadingState(modifier)
+        is Success ->
+            if (feedState.feed.isNotEmpty()) {
+                BookmarksGrid(feedState, removeFromBookmarks, modifier)
+            } else {
+                EmptyState(modifier)
```
LinX64, line 89: Looks awesome now, thanks! 

However, we are pretty sure there will be no other situation except `isNotEmpty()` in the future, so why not removing the curly bracket to make it even more clear? 

This way, we will be able to put the `if` condition on `Success` body without using the curly bracket, as well as removing it from the last line and the result will be:

```
when (feedState) {
        Loading -> LoadingState(modifier)
        is Success -> if (feedState.feed.isNotEmpty()) {
            BookmarksGrid(feedState, removeFromBookmarks, modifier)
        } else EmptyState(modifier)
    }

```

Please let me know what you think :)

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -79,41 +109,60 @@ fun BookmarksScreen(
...
+            text = stringResource(id = R.string.bookmarks_empty_description),
+            modifier = Modifier.fillMaxWidth(),
+            textAlign = TextAlign.Center,
+            style = MaterialTheme.typography.bodyMedium
+        )
     }
 }
+
+@Composable
+private fun LoadingState(modifier: Modifier = Modifier) {
```
LinX64, line 160: Amazing separation :) just one thing I always care about is priorities. For instance, the `when` expression checks for the `Loading` state **first**, and you put the `Loading` state composable function at the end of the file. I wouldn't do that, I would put the `Loading()` function before the `BookmarksGrid()`.

- This becomes very handy when we have a huge class (or we will have?) and we are trying to scroll down and find the right function for it, so I am always placing those functions based on the priorities and the places where they are getting called.

Please let me know what you think :) thank you.

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -79,41 +109,60 @@ fun BookmarksScreen(
...
+
+        Text(
+            text = stringResource(id = R.string.bookmarks_empty_error),
+            modifier = Modifier.fillMaxWidth(),
+            textAlign = TextAlign.Center,
+            style = MaterialTheme.typography.titleMedium,
+            fontWeight = FontWeight.Bold
+        )
+
+        Spacer(modifier = Modifier.height(5.dp))
```
mmoczkowski, line 148: this spacing is a bit odd. Won't `Arrangement.spacedBy` be a better alternative?

mmoczkowski (approved):

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
JolandaVerhoef: Great point, I changed the order and added some previews as well :)

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
JolandaVerhoef: Agreed that the Spacer makes more sense here, although in Figma it seems to be 8dp, not 5dp.

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
JolandaVerhoef: Hmm I'm not really a big fan of if/else statements without curly brackets unless they're really trivial and you can fit the whole thing on one line. So in this case I'd opt for the version with the curly brackets, to be most explicit. But I don't have a super strong preference either way.

JolandaVerhoef (approved):

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
LinX64: Good point! Just changed it from 5 to 8dp. Please check the latest commit.

feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
LinX64: Alright then, I just removed the space before the `if` condition.

JolandaVerhoef:
@LinX64 Thank you for your contribution, it's much appreciated! Code looks great :)

Issue #425, opened 2022-11-11
[Bug]: No internet connection SnackBar keeps showing even with Internet connection while changing themes!


LinX64:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

A bug happened when I was trying to change the themes too fast. It is a bit tricky to reproduce, you will need to select a theme a few times, and then Boom!, you will see the snake Bar. 

I've tested it on the latest emulator API 33. It would be so great to check it on a Physical device as well.

Screenshot:

<img src="https://i.imgur.com/uDq4NnI.png" height="400" />

Proof:

https://i.imgur.com/ki8d9sY.mp4

### Relevant logcat output

```shell
--------- beginning of main
--------- beginning of system
2022-11-11 21:53:58.671  1576-1576  studio.deploy           pid-1576                             E  Could not remove dir '/data/data/com.google.samples.apps.nowinandroid.demo.debug/code_cache/.ll/': No such file or directory
2022-11-11 21:55:34.461  4994-4994  studio.deploy           pid-4994                             E  Could not remove dir '/data/data/com.google.samples.apps.nowinandroid.demo.debug/code_cache/.ll/': No such file or directory
2022-11-11 21:56:21.357  5211-5211  studio.deploy           pid-5211                             E  Could not remove dir '/data/data/com.google.samples.apps.nowinandroid.demo.debug/code_cache/.ll/': No such file or directory
---------------------------- PROCESS STARTED (5223) for package com.google.samples.apps.nowinandroid.demo.debug ----------------------------
2022-11-11 21:56:21.782  5223-5223  ziparchive              com....apps.nowinandroid.demo.debug  W  Unable to open '/data/app/~~aeCUpu2cizf6Y8W2LfsoLg==/com.google.samples.apps.nowinandroid.demo.debug-4FrqWk7TSLb8bqM4SV7wiQ==/base.dm': No such file or directory
2022-11-11 21:56:21.782  5223-5223  ziparchive              com....apps.nowinandroid.demo.debug  W  Unable to open '/data/app/~~aeCUpu2cizf6Y8W2LfsoLg==/com.google.samples.apps.nowinandroid.demo.debug-4FrqWk7TSLb8bqM4SV7wiQ==/base.dm': No such file or directory
[...]

PR #445, opened 2022-11-18
Check active network for connectivity state each time network changes

tunjid:
Checks the active network for connectivity when a network is gained or lost.

### DIFF
--- a/core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
+++ b/core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
@@ -35,18 +35,30 @@ import kotlinx.coroutines.flow.conflate
 class ConnectivityManagerNetworkMonitor @Inject constructor(
     @ApplicationContext private val context: Context
 ) : NetworkMonitor {
-    override val isOnline: Flow<Boolean> = callbackFlow<Boolean> {
+    override val isOnline: Flow<Boolean> = callbackFlow {
+        val connectivityManager = context.getSystemService<ConnectivityManager>()
+
+        /*
+        The callback's methods are invoked on changes to *any* network, not just the active
+        network. So to check for network connectivity, one must query the active network of the
+        ConnectivityManager.
+        */
         val callback = object : NetworkCallback() {
             override fun onAvailable(network: Network) {
-                channel.trySend(true)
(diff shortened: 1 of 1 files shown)


core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -35,18 +35,25 @@ import kotlinx.coroutines.flow.conflate
 class ConnectivityManagerNetworkMonitor @Inject constructor(
     @ApplicationContext private val context: Context
 ) : NetworkMonitor {
-    override val isOnline: Flow<Boolean> = callbackFlow<Boolean> {
+    override val isOnline: Flow<Boolean> = callbackFlow {
+        val connectivityManager = context.getSystemService<ConnectivityManager>()
+
         val callback = object : NetworkCallback() {
```
dturner, line 41: I think it's worth adding a comment why we need to call `connectivityManger.isCurrentlyConnected` for each method in the callback as it's not immediately obvious (as we found from trying fix #425).

dturner (approved):
Please add that comment, then merge

core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -35,18 +35,28 @@ import kotlinx.coroutines.flow.conflate
 class ConnectivityManagerNetworkMonitor @Inject constructor(
     @ApplicationContext private val context: Context
 ) : NetworkMonitor {
-    override val isOnline: Flow<Boolean> = callbackFlow<Boolean> {
+    override val isOnline: Flow<Boolean> = callbackFlow {
+        val connectivityManager = context.getSystemService<ConnectivityManager>()
+
+        // The callback's methods are invoked on changes to *any* network, not just the active
```
LinX64, line 41: Shouldn't we use Multiline comments (`\**`) instead of single-line comments for three lines of description? This way it will also make it easier to collapse and uncollapse the comment later.

tunjid: Fair point, done.

dturner: Good spot. Fixed on `main`

PR #447, opened 2022-11-18
Sort authors stream by news count

### LINKED ISSUE #393: [FR]: Sort authors by how many resources they have written

### Describe the problem

Currently, authors are sorted alphabetically, which leads to a lot of noise (someone who has written a single article might end up more prominently than someone that has written many). 

### Describe the solution

Should we consider sorting them by number of resources added?

### Additional context

_No response_

### Code of Conduct

SimonMarquis:
Fixes #393

### DIFF
--- a/core/domain/src/main/java/com/google/samples/apps/nowinandroid/core/domain/GetSortedFollowableAuthorsStreamUseCase.kt
+++ b/core/domain/src/main/java/com/google/samples/apps/nowinandroid/core/domain/GetSortedFollowableAuthorsStreamUseCase.kt
@@ -17,34 +17,49 @@
 package com.google.samples.apps.nowinandroid.core.domain
 
 import com.google.samples.apps.nowinandroid.core.data.repository.AuthorsRepository
+import com.google.samples.apps.nowinandroid.core.data.repository.NewsRepository
 import com.google.samples.apps.nowinandroid.core.data.repository.UserDataRepository
 import com.google.samples.apps.nowinandroid.core.domain.model.FollowableAuthor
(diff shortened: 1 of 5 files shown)


SimonMarquis:
This is definitely not the most efficient way to order by news by count, but since `getNewsResourcesStream()` is already called by the `GetSaveableNewsResourcesStreamUseCase` I guess this is good enough for now.

SimonMarquis:
image

SimonMarquis:
There is unfortunately one issue at the moment.
During the first launch, `GetSortedFollowableAuthorsStreamUseCase()` is called 4 times.
The first 3 times, `newsRepository.getNewsResourcesStream()` returns an empty list.
Which means authors are initially displayed on the `AuthorsCarousel` sorted by name.
But then, on the 4th invocation, `newsRepository.getNewsResourcesStream()` contains data.
`AuthorsCarousel` is then updated with the expected sort, but does not scroll to the start as it tries to remain stable.

Potential solutions are:
- return an empty list of authors from `GetSortedFollowableAuthorsStreamUseCase` until the `newsRepository.getNewsResourcesStream()` is no longer empty (but it creates a layout shift)
- wait for non-empty `newsRepository.getNewsResourcesStream()` in the `StateFlow<OnboardingUiState>`
- always scroll the `AuthorsCarousel` to the start when it changes

JolandaVerhoef:
This is a duplicate of #394

SimonMarquis:
Ho, my bad. I did not see this draft PR as it did not reference the issue 🙈

PR #452, opened 2022-11-19
Migrate to testFixtures, closes #451

### LINKED ISSUE #451: [FR]: Migrate project to testFixtures

### Describe the problem

Even though [test fixtures](https://docs.gradle.org/current/userguide/java_testing.html) are not completely supported by the Android Gradle Plugin (specifically with Kotlin sources) we can still benefit a lot from this feature on JVM modules, and this should definitely be part of Android dev's best practices!

### Describe the solution

This would remove **a lot** of code duplication and prevent unecessary hacks like shared `sourcesSet` and gradle fixtures `module`s that could impact build times, especially in large code bases.
In the current state of the code base, it sheds ~1000 lines of code.

At my current company, we chose to use `kotlin.Random` as receiver for these testFixtures to make them more explicit about their behavior (being random by default), but this can be changed here if needed.

Next steps will be to migrate:
- "test" modules: `:core:data-test`, `:core:datastore-test`, `:sync-test`
- fakes and test classes like `FakeStringDecoder`, `TestAuthorsRepository`, `TestNewsRepository`, `TestTopicsRepository`, `TestUserDataRepository`

These next steps could already be done, but would need to rewrite classes in Java to be visible, and I don't think this is a path we want to go for this project.
So I guess we'll wait for the full Android/Kotlin support.

### Additional context

_No response_

### Code of Conduct

SimonMarquis:


LinX64:
Thanks for the work, and the time you spent :) But I would like to mention something;

I've seen your PRs, and everything seems to be perfect, but I believe you forgot something which is really important and might help all of us in the future, especially the code reviewers. You have many changes on your PRs, which is good, but merging everything into a single commit, doesn't seem to be fine. Imagine someone's trying to code review your PR, and this becomes a problem with 22 file changes inside a single commit.

What do you think if we separate the changes into **_single commits_** and merge **ONLY** those which are related to each other at the end of the work before pushing? This way, it will make it much easier to code review and check the changes.

SimonMarquis:
> merging everything into a single commit, doesn't seem to be fine

Well, for such changes (like migrations) I don't see the point of it, at all.
Splitting these kind of PR into multiple commits, each migrating a file (or module) wont make it more "readable" for the reviewer.

> What do you think if we separate the changes into single commits and merge ONLY those which are related to each other at the end of the work before pushing?

Again, the changes of this PR is well scoped. I did not changed any behavior of the codebase.

The only thing that would make sense to split is:
- one commit adding the testFixtures
- another one replacing all usages
🤷

core/model/src/testFixtures/kotlin/com/google/samples/apps/nowinandroid/core/model/data/AuthorFixtures.kt
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.samples.apps.nowinandroid.core.model.data
+
+import kotlin.random.Random
+
+fun Random.nextFakeAuthor(
```
JolandaVerhoef, line 21: Why are you using random values to create Test Fixtures? Wouldn't this make the tests prone to flakiness, as a test could fail with one version of the random value, but succeed with another (and thus you can't replicate test failures)

mmoczkowski:
Closing due to inactivity. If you still wish to submit this change, please create a new PR.

SimonMarquis:
For anyone interested in updates, I've pushed a new PR that enables "almost" proper Android test fixtures thanks to AGP 8.5.0:
- #1515

PR #453, opened 2022-11-20
[Saved screen] Added the test for empty state and refactor 

LinX64:
This PR is about the test for `EmptyState()` function.

I basically removed the unused `windowSizeClass` variable, and added the check for both Texts for the `EmptyState()`. I was also looking for a test function to check **if All** elements/Composables are available or not, but I wasn't able to find it; Something like:

`onAllNodesExists()`

Which should obviously receive the ids/strings of the elements/Composable.

Any thoughts would be appreciated :)


dturner (approved):
Thanks for doing this. Looks good.

PR #454, opened 2022-11-20
Refactor | Added extension functions

LinX64:
This PR is about polishing codes, and using extension functions for `StateIn()` (mostly in ViewModels).

Basically, all of the `StateIn()` functions inside ViewModels were following/using the same pattern, so I decided to create two extension functions for use in both `ViewModel`s using `viewModelScope`, as well as `CoroutineScope` outside of `ViewModels`.

### DIFF
--- a/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/Extenstions.kt
+++ b/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/Extenstions.kt
@@ -23,23 +23,12 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.stateIn
 
 /**
- * This is a helper extension function to make it easier to use [Flow.stateIn] within a ViewModel.
+ * This is a helper extension function for making it easier to use [Flow.stateIn] within a ViewModel or etc.
  * @Returns a [StateFlow] that shares the latest value emitted by the original [Flow] and starts
  * with the [initialValue].
  */
-fun <T> Flow<T>.stateInViewModelScope(
-    viewModelScope: CoroutineScope,
+fun <T> Flow<T>.stateInScope(
+    coroutineScope: CoroutineScope,
     started: SharingStarted = SharingStarted.WhileSubscribed(5_000),
     initialValue: T
-): StateFlow<T> = stateIn(viewModelScope, started, initialValue)
-
-/**
- * This is a helper extension function to make it easier to use [Flow.stateIn] within a [CoroutineScope].
- * @Returns a [StateFlow] that shares the latest value emitted by the original [Flow] and starts
- * with the [initialValue].
- */
-fun <T> Flow<T>.stateInCoroutineScope(
-    viewModelScope: CoroutineScope,
-    started: SharingStarted = SharingStarted.WhileSubscribed(5_000),
-    initialValue: T
-): StateFlow<T> = stateIn(viewModelScope, started, initialValue)
+): StateFlow<T> = stateIn(coroutineScope, started, initialValue)
--- a/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/Extenstions.kt
+++ b/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/Extenstions.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.samples.apps.nowinandroid.core.ui
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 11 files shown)


core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/Extenstions.kt
```diff
@@ -0,0 +1,45 @@
...
+    initialValue: T
+): StateFlow<T> = stateIn(viewModelScope, started, initialValue)
+
+/**
+ * This is a helper extension function to make it easier to use [Flow.stateIn] within a [CoroutineScope].
+ * @Returns a [StateFlow] that shares the latest value emitted by the original [Flow] and starts
+ * with the [initialValue].
+ */
+fun <T> Flow<T>.stateInCoroutineScope(
+    viewModelScope: CoroutineScope,
```
SimonMarquis, line 42:
```suggestion
    coroutineScope: CoroutineScope,
```

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/Extenstions.kt
```diff
@@ -0,0 +1,45 @@
...
+    started: SharingStarted = SharingStarted.WhileSubscribed(5_000),
+    initialValue: T
+): StateFlow<T> = stateIn(viewModelScope, started, initialValue)
+
+/**
+ * This is a helper extension function to make it easier to use [Flow.stateIn] within a [CoroutineScope].
+ * @Returns a [StateFlow] that shares the latest value emitted by the original [Flow] and starts
+ * with the [initialValue].
+ */
+fun <T> Flow<T>.stateInCoroutineScope(
```
SimonMarquis, line 41: This extension has the same signature as `stateInViewModelScope`.

feature/author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorViewModel.kt
```diff
@@ -54,75 +56,67 @@ class AuthorViewModel @Inject constructor(
...
-            )
+    val newsUiState: StateFlow<NewsUiState> = getSaveableNewsResourcesStream
+        .newsUiStateStream(authorId = authorArgs.authorId)
+        .stateInViewModelScope(viewModelScope, initialValue = NewsUiState.Loading)
 
-    fun followAuthorToggle(followed: Boolean) {
-        viewModelScope.launch {
-            userDataRepository.toggleFollowedAuthorId(authorArgs.authorId, followed)
-        }
+    fun followAuthorToggle(followed: Boolean) = viewModelScope.launch {
```
SimonMarquis, line 65: Refactoring this with an expression body leaks the `Job` instance to the caller.
The implicit signature of the function becomes `fun followAuthorToggle(followed: Boolean) : Job`.

feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsViewModel.kt
```diff
@@ -44,29 +44,23 @@ class SettingsViewModel @Inject constructor(
...
-                started = SharingStarted.Eagerly,
-                initialValue = Loading
-            )
+            }.stateInViewModelScope(viewModelScope, SharingStarted.Eagerly, Loading)
 
-    fun updateThemeBrand(themeBrand: ThemeBrand) {
-        viewModelScope.launch {
-            userDataRepository.setThemeBrand(themeBrand)
-        }
+    /**
```
SimonMarquis, line 49: We should keep the comment next to the corresponding code

```kotlin
            // Starting eagerly means the user data is ready when the SettingsDialog is laid out
            // for the first time. Without this, due to b/221643630 the layout is done using the
            // "Loading" text, then replaced with the user editable fields once loaded, however,
            // the layout height doesn't change meaning all the fields are squashed into a small
            // scrollable column.
            // TODO: Change to SharingStarted.WhileSubscribed(5_000) when b/221643630 is fixed
            }.stateInViewModelScope(viewModelScope, SharingStarted.Eagerly, Loading)

```

or


```kotlin
            }.stateInViewModelScope(
                viewModelScope,
                // Starting eagerly means the user data is ready when the SettingsDialog is laid out
                // for the first time. Without this, due to b/221643630 the layout is done using the
                // "Loading" text, then replaced with the user editable fields once loaded, however,
                // the layout height doesn't change meaning all the fields are squashed into a small
                // scrollable column.
                // TODO: Change to SharingStarted.WhileSubscribed(5_000) when b/221643630 is fixed
                started = SharingStarted.Eagerly,
                initialValue = Loading,
            )
```

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/Extenstions.kt
LinX64: Thanks for the notice! I decided to remove the other extension and used only one extension because it was following the same pattern, this way, we will be able to use `stateInScope()` almost everywhere.

feature/author/src/main/java/com/google/samples/apps/nowinandroid/feature/author/AuthorViewModel.kt
LinX64: That's right, I missed that. Thanks for mentioning it!

manuelvicnt:
Unsure about adding this API / helper function to the Now in Android app. 

It just saves you one line of code and the name (currently `stateInScope`) isn't clear what is doing or how it's different from the `stateIn` operator.

The ideal API for this would be a context receiver on both `Flow` and `ViewModel` so that you could have a Flow operator using the `viewModelScope`. e.g.

```
context(ViewModel)
fun Flow.produceState(
    coroutineScope: CoroutineScope = viewModelScope,
    started: SharingStarted = SharingStarted.WhileSubscribed(5_000),
    initialValue: T
): StateFlow<T> = ...
```
Until context receivers aren't widely available in Android and Compose, I'd refrain from adding a helper function for this functionality in a highly visible Android sample.

LinX64:
Closing this PR as it is suggested to use Context receivers as a solution/replacement for the implementation.

P.S: This PR can wait for the Context Receivers to be available widely on Android and Kotlin.

PR #458, opened 2022-11-23
Fix benchmarks

mlykotom:
- Fix app suffix for benchmarks
- enable R8 for benchmarks
- add release build type to catalog, so that studio doesn't complain with variant conflicts (could/should we make it a composite build?)
- add helper ui automator function to check for number of children
- improve waiting for content
- select random authors


build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiABuildType.kt
```diff
@@ -0,0 +1,27 @@
...
+ *   limitations under the License.
+ */
+
+package com.google.samples.apps.nowinandroid
+
+/**
+ * This is shared between :app and :benchmarks module to provide configurations type safety.
+ */
+@Suppress("unused")
+enum class NiABuildType(val applicationIdSuffix: String? = null) {
```
wojtek-kalicinski, line 23: nit: class for supporting flavors is just `Flavor` but this one is `NiaBuildType` (**Nia** prefix). Maybe make it consistent?

wojtek-kalicinski (changes requested):
As discussed offline please remove randomness from profile generation.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiABuildType.kt
mlykotom: Renamed 👍🏼

mlykotom:
Ok, fixed the randomness and testing if author is already checked.

I regenerated the BLP profiles after the fixes with interesting numbers:

**BEFORE**
```
StartupBenchmark_startupBaselineProfile
   timeToFullDisplayMs   min 468.0,   median 516.8,   max 601.8
timeToInitialDisplayMs   min 244.7,   median 256.5,   max 271.8
```
**AFTER**
```
StartupBenchmark_startupBaselineProfile
   timeToFullDisplayMs   min 267.5,   median 298.2,   max 593.0
timeToInitialDisplayMs   min 179.7,   median 187.4,   max 353.3
```

wojtek-kalicinski (approved):

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiAFlavor.kt
```diff
@@ -2,31 +2,35 @@ package com.google.samples.apps.nowinandroid
...
 fun Project.configureFlavors(
-    commonExtension: CommonExtension<*, *, *, *>
+    commonExtension: CommonExtension<*, *, *, *>,
+    flavorConfigurationBlock: ProductFlavor.(flavor: NiAFlavor) -> Unit = {}
 ) {
     commonExtension.apply {
         flavorDimensions += FlavorDimension.contentType.name
         productFlavors {
-            Flavor.values().forEach{
+            NiAFlavor.values().forEach {
```
dturner, line 30: Convention is to prefix with `Nia` rather than `NiA`

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -126,15 +126,14 @@ internal fun ForYouScreen(
     onNewsResourcesCheckedChanged: (String, Boolean) -> Unit,
     modifier: Modifier = Modifier,
 ) {
+    val onboardingLoading = onboardingUiState is OnboardingUiState.Loading
```
dturner, line 129:
```suggestion
    val isOnboardingLoading = onboardingUiState is OnboardingUiState.Loading
```

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -126,15 +126,14 @@ internal fun ForYouScreen(
...
+    val feedLoading = feedState is NewsFeedUiState.Loading
```
dturner, line 130:
```suggestion
    val isFeedLoading = feedState is NewsFeedUiState.Loading
```

dturner:
LGTM with minor suggestions. Please merge once they're addressed.

dturner (approved):

PR #460, opened 2022-11-23
Migrate package from core:domain to namespace in build.gradle file

Lonelywood:
Migrate package from core:domain to namespace in build.gradle file. To be compliant with AGP 8+


alexvanyo:
LGTM, thank you for the contribution!

Adding @wojtek-kalicinski to confirm changes as well.

wojtek-kalicinski (approved):

alexvanyo (approved):

PR #465, opened 2022-11-29
Use Chrome Custom tab for opening articles

calren:
Before change:
[before.webm](https://user-images.githubusercontent.com/5016806/204655634-9e27c410-4868-4a41-b148-31cce11de47f.webm)

After:
[after.webm](https://user-images.githubusercontent.com/5016806/204655453-b19b6319-89e9-4b30-b4fe-46aaaf15beae.webm)

### DIFF
--- a/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardList.kt
+++ b/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardList.kt
@@ -20,7 +20,9 @@ import android.content.Intent
 import android.net.Uri
 import androidx.compose.foundation.lazy.LazyListScope
 import androidx.compose.foundation.lazy.items
+import androidx.compose.material3.MaterialTheme
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.platform.LocalContext
 import androidx.core.content.ContextCompat
 import com.google.samples.apps.nowinandroid.core.model.data.NewsResource
@@ -47,8 +49,8 @@ fun <T> LazyListScope.newsResourceCardItems(
     key = { newsResourceMapper(it).id },
     itemContent = { item ->
         val newsResource = newsResourceMapper(item)
-        val launchResourceIntent =
-            Intent(Intent.ACTION_VIEW, Uri.parse(newsResource.url))
+        val resourceUrl = Uri.parse(newsResource.url)
+        val backgroundColor = MaterialTheme.colorScheme.background.toArgb()
         val context = LocalContext.current
 
         NewsResourceCardExpanded(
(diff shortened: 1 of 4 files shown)


dturner:
I haven't had time to review this but just wanted to say this is looking _sweet_ 👌

alexvanyo:
There's similar code in `NewsResourceCardList.kt`, do we want to update that as well?

https://github.com/android/nowinandroid/blob/56b3c1d0b950ed27051424ca6718743d7a3c6b97/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardList.kt#L50-L51

calren:
ah yes, good catch Alex - fixed by implementing in NewsResourceCardList as well!

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardList.kt
```diff
@@ -47,8 +47,8 @@ fun <T> LazyListScope.newsResourceCardItems(
     key = { newsResourceMapper(it).id },
     itemContent = { item ->
         val newsResource = newsResourceMapper(item)
-        val launchResourceIntent =
-            Intent(Intent.ACTION_VIEW, Uri.parse(newsResource.url))
+        val resourceUrl = Uri.parse(newsResource.url)
+        val backgroundColor = MaterialTheme.colorScheme.background.toArgb()
```
dturner, line 51: Just wondering if there's a way we could move the getting of `backgroundColor` and `context` to inside `launchCustomChromeTab` to avoid having to write this code each time we want to call `launchCustomChromeTab`.

LinX64: That's a really good question! I thought about it, and I couldn't really find a suitable solution for it because both `context` and `backgroundColor` are coming from composable functions and we can't really get the `context` out of nowhere, and the same situation for `backgroundColor`. Even if we find a solution for the `context`, then it would probably be a dangerous one because of leaking context and so on. Please also notice that `launchCustomChromeTab()` is not a composable function and maybe it can't even be one!

Anyways, the only thing that came to my mind was to write an extension function to avoid the repetition of the above method;

`fun Context.launchCustomChromeTab()`

Please let me know what you think! Thanks

alexvanyo (approved):

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardList.kt
calren: i couldn't find a way to call `MaterialTheme.colorScheme` without making `launchCustomChromeTab` a Composable as well. 

I think generally, chrome custom tabs doesn't work cleanly with Compose code. Setting a customized `close` button is also difficult with Compose. Perhaps something to file for the chrome custom tab team.

dturner (approved):

PR #466, opened 2022-11-29
Add accompanist/testharness for checking navigation UI

alexvanyo:
Adds a usage of accompanist/testharnness to force a given DpSize for its content, overriding the density if necessary to ensure that the size is achieved. This allows testing UI at arbitrary sizes on arbitrary emulators, lessening the need for using multiple emulators at different sizes (and filtering tests) when Android-specific window size has been hoisted appropriately.

The initial test using it is a test for checking whether or not the navigation bar or navigation rail is being shown, and these tests can run on any emulator.


app/build.gradle.kts
```diff
@@ -98,8 +98,10 @@ dependencies {
     androidTestImplementation(project(":core:data-test"))
     androidTestImplementation(project(":core:network"))
     androidTestImplementation(libs.androidx.navigation.testing)
+    androidTestImplementation(libs.accompanist.testharness)
     androidTestImplementation(kotlin("test"))
     debugImplementation(libs.androidx.compose.ui.testManifest)
+    debugImplementation(project(":ui-test-hilt-manifest"))
```
JoseAlcerreca, line 104: Is this used anywhere else? I'm surprised it's only needed in app/

JoseAlcerreca (approved):

PR #471, opened 2022-12-05
Update to wrapper 7.6 and use --rerun task

mlykotom:
Gradle wrapper 7.6 adds --rerun task that will always run the required task regardless of cache.
Until now we had to use --rerun-tasks, which ignore cache for all tasks.

Also added --enable-display parameter to GMD, so that emulator doesn't run in headless mode locally and we're able to verify the generator runs properly.

### DIFF
--- a/.run/Generate Demo Baseline Profile.run.xml
+++ b/.run/Generate Demo Baseline Profile.run.xml
@@ -16,22 +16,27 @@
 -->
 <component name="ProjectRunConfigurationManager">
   <!--
-  Baseline Profiles improve code execution speed by around 30% from the first launch by avoiding interpretation and just-in-time (JIT) compilation steps for included code paths.
+  Baseline Profiles improve code execution speed by around 30% from the first launch by avoiding
+  interpretation and just-in-time (JIT) compilation steps for included code paths.
   More information at http://d.android.com/baseline-profiles.
+
+  In this run configuration we leverage rerun parameter that always reruns the requested task regardless of cache.
+  We also leverage enable-display parameter to be able to verify the generator works as intended.
   -->
   <configuration default="false" name="Generate Demo Baseline Profile" type="GradleRunConfiguration" factoryName="Gradle">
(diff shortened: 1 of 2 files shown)


.run/Generate Demo Baseline Profile.run.xml
```diff
@@ -16,22 +16,27 @@
...
-      <!-- TODO Once we use Gradle wrapper 7.6, we can use rerun instead of rerun-tasks that will skip cache only for one task -->
-      <option name="scriptParameters" value="--rerun-tasks -Pandroid.testInstrumentationRunnerArguments.androidx.benchmark.enabledRules=BaselineProfile" />
+      <option name="scriptParameters" value="-Pandroid.testInstrumentationRunnerArguments.androidx.benchmark.enabledRules=BaselineProfile" />
       <option name="taskDescriptions">
         <list />
       </option>
       <option name="taskNames">
         <list>
           <option value=":benchmark:pixel6Api31DemoBenchmarkAndroidTest" />
+          <option value="--rerun" />
```
keyboardsurfer, line 38: Please comment what the options do in the script.

keyboardsurfer (approved):

.run/Generate Demo Baseline Profile.run.xml
mlykotom: It's described in the header comment. I thought it's more visible than directly above the parameters.
https://github.com/android/nowinandroid/pull/471/files#diff-d76529b46a07c19070f97353beeac4477e4803633f2d1c69f6c7d2199c661fd1R23-R24

dturner: Nudge to either change the comment or merge as-is. As long as there's an expert reviewer approval (in this case @keyboardsurfer) you're fine to merge yourself.

mlykotom: Ok, thanks for the info. Merging as is

PR #475, opened 2022-12-06
Fix benchmarks after dec5 merge

mlykotom:
- instead of selecting people, we select topics now
- adds waiting for loading and title in interests screen


benchmarks/src/main/java/com/google/samples/apps/nowinandroid/foryou/ForYouActions.kt
```diff
@@ -23,12 +23,53 @@ import androidx.test.uiautomator.untilHasChildren
...
+    val topics = device.findObject(By.res("forYou:topicSelection"))
+
+    // Set gesture margin from sides not to trigger system gesture navigation
+    val horizontalMargin = 10 * topics.visibleBounds.width() / 100
+    topics.setGestureMargins(horizontalMargin, 0, horizontalMargin, 0)
+
+    // Select some topics to show some feed content
+    repeat(3) { index ->
+        // Selecting topics that are in the first column
+        val topic = topics.children[(2 * index) % topics.childCount]
```
alexvanyo, line 49: Could you elaborate on the logic here? The `2 * index` seems a bit strange, interacting with the `0`, `2` and `4` items.

mlykotom: Ellaborated in the comment. Basically it doesn't matter, because the whole point is to select something so we have content to scroll. But since you can see 1 item fully and just part of the next item, I thought it would be nicer to select the fully visible ones.

alexvanyo: Ah I see, so the `children` object here goes from left to right, top to bottom.

Is the `2` an assumption that there are only 2 topics visible in each row?
In other words, if this test were to be run on a tablet where there is more room, the `0`, `2` and `4` topics probably wouldn't all be in the first column?

mlykotom: Yes :) I think it doesn't matter much, because we just basically need to select some topics for the feed to be shown.
But if it causes confusion, I can keep it indexed normally

alexvanyo: I'd probably slightly prefer just doing `0` `1` and `2` to reduce complexity, but that's a small nit.

alexvanyo (approved):

PR #476, opened 2022-12-07
Display resource type for each news resource

calren:
resource_type


core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -195,10 +196,22 @@ private fun dateFormatted(publishDate: Instant): String {
...
-    Text(dateFormatted(publishDate), style = MaterialTheme.typography.labelSmall)
+    val date = dateFormatted(publishDate)
+
+    if (resourceType != NewsResourceType.Unknown) {
+        Text(
+            "$date • ${resourceType.displayText}", style = MaterialTheme.typography.labelSmall
+        )
+    } else {
+        Text(
+            date, style = MaterialTheme.typography.labelSmall
```
alexvanyo, line 212: Nit: We could probably put the `if` inside the `Text`? That'd avoid a duplication on `labelSmall` for the `style`

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -195,10 +196,22 @@ private fun dateFormatted(publishDate: Instant): String {
...
+    publishDate: Instant,
+    resourceType: NewsResourceType
+
 ) {
-    Text(dateFormatted(publishDate), style = MaterialTheme.typography.labelSmall)
+    val date = dateFormatted(publishDate)
+
+    if (resourceType != NewsResourceType.Unknown) {
+        Text(
+            "$date • ${resourceType.displayText}", style = MaterialTheme.typography.labelSmall
```
alexvanyo, line 208: Should this string template be in a `strings.xml` for translation purposes? I'm just thinking if a RTL language would want the order of `date` and the resource type to be swapped.

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -195,10 +196,22 @@ private fun dateFormatted(publishDate: Instant): String {
...
-    publishDate: Instant
+fun NewsResourceDateAndType(
+    publishDate: Instant,
+    resourceType: NewsResourceType
+
 ) {
-    Text(dateFormatted(publishDate), style = MaterialTheme.typography.labelSmall)
+    val date = dateFormatted(publishDate)
+
+    if (resourceType != NewsResourceType.Unknown) {
```
alexvanyo, line 206: Could we add a test for both possible codepaths here?

calren: I added a test, but I keep getting this error when running the test locally:
```
FINGERPRINT must not be null
java.lang.NullPointerException: FINGERPRINT must not be null
	at androidx.compose.ui.test.AndroidComposeUiTestEnvironment.runTest(ComposeUiTest.android.kt:280)
```

still trying to figure out why...

calren: ah, i had my test in `test` directory instead of `androidTest`. now it run as expected. will fix my test now that it's running.

core/ui/src/test/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        composeTestRule
+            .onNodeWithText(dateFormatted + " • " + newsResource.type.displayText)
+            .assertIsDisplayed()
+
+        // Resource with Unknown resource type
+        val newsResource2 = previewNewsResources[4]
+
+        composeTestRule.setContent {
```
alexvanyo, line 54: Nit: I think it would be preferable if this was in two different tests, to have a single `setContent` call in each.

alexvanyo (approved):
LGTM, just one nit on splitting the test to have one for each case.

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -195,10 +196,19 @@ private fun dateFormatted(publishDate: Instant): String {
 }
 
 @Composable
-fun NewsResourceDate(
-    publishDate: Instant
+fun NewsResourceDateAndType(
```
dturner, line 199: Since you're using the term "meta data" elsewhere to refer to the information being displayed by this composable, a better name for this might be `NewsResourceMetaData`

dturner (approved):
Approved with one naming nit

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
calren: good catch, updated

PR #477, opened 2022-12-07
[ForYou] Implementing offline screen and removing SnakeBar 

LinX64:
There are a few changes I've made:

- Removed SnakeBar from the `NiaApp`.
- Implemented the offline mode and error based on design specs.
- Separated components to make it easier to read on the ForYou screen.
- Refactor for the `BookmarksScreen`.

Output:


tunjid:
The design needs to be updated. With your change, if there's no internet, items in the database will not be shown. This is not a good user experience.

LinX64:
> The design needs to be updated.

Exactly!

PR #479, opened 2022-12-12
Introduce GMD + ATD

thagikura:
- Introduce Gradle Managed Device definition in a single module (core-database at the moment) using Automated Test Device

### DIFF
--- a/.github/workflows/AndroidCIWithGmd.yaml
+++ b/.github/workflows/AndroidCIWithGmd.yaml
@@ -31,4 +31,4 @@ jobs:
       with:
         name: test-reports
         path: |
-          '**/build/outputs/androidTest-results/'
+          '**/build/reports/androidTest-results/'
--- a/.github/workflows/AndroidCIWithGmd.yaml
+++ b/.github/workflows/AndroidCIWithGmd.yaml
@@ -22,5 +22,13 @@ jobs:
       uses: android-actions/setup-android@v2
 
     - name: Run instrumented tests with GMD
+      continue-on-error: true
       run: ./gradlew cleanManagedDevices --unused-only && ./gradlew pixel4api30DemoDebugAndroidTest -Dorg.gradle.workers.max=1  -Pandroid.testoptions.manageddevices.emulator.gpu="swiftshader_indirect" -Pandroid.experimental.testOptions.managedDevices.emulator.showKernelLogging=true --info
 
+    - name: Upload test reports
+        if: success() || failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-reports
+          path: |
+            '**/build/outputs/androidTest-results/'
--- a/.github/workflows/AndroidCIWithGmd.yaml
+++ b/.github/workflows/AndroidCIWithGmd.yaml
@@ -31,4 +31,4 @@ jobs:
       with:
         name: test-reports
         path: |
-          '*/build/reports/androidTest-results/'
+          '**/*/build/reports/androidTest/'
--- a/.github/workflows/AndroidCIWithGmd.yaml
+++ b/.github/workflows/AndroidCIWithGmd.yaml
@@ -25,13 +25,10 @@ jobs:
       continue-on-error: true
       run: ./gradlew cleanManagedDevices --unused-only && ./gradlew pixel4api30DemoDebugAndroidTest -Dorg.gradle.workers.max=1  -Pandroid.testoptions.manageddevices.emulator.gpu="swiftshader_indirect" -Pandroid.experimental.testOptions.managedDevices.emulator.showKernelLogging=true --info
 
-    - name: Setup upterm session
-      uses: lhotari/action-upterm@v1
-
     - name: Upload test reports
       if: success() || failure()
       uses: actions/upload-artifact@v3
       with:
         name: test-reports
         path: |
-          '**/*/build/reports/androidTest/'
+          '**/*/build/reports/androidTests/'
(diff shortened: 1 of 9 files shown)


dturner:
GMD setup is failing with: 

```
Execution failed for task ':core:database:pixel4api30Setup'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.ManagedDeviceSetupTask$ManagedDeviceSetupRunnable
   > java.lang.IllegalStateException: Snapshot setup ran successfully, but the snapshot failed to be created. This is
     likely to a lack of disk space for the snapshot. Try the cleanManagedDevices
     task with the --unused-only flag to remove any unused devices for this project.
```

https://github.com/android/nowinandroid/actions/runs/3673045398/jobs/6212504226

thagikura:
I ran it 10 times and failed it due to the lack of disk space 2 times.
https://github.com/android/nowinandroid/actions/runs/3673045398

I'll see if it can be stabilized.

thagikura:
After this commit https://github.com/android/nowinandroid/pull/479/commits/392408cfa29374f89fed499fe0681a78de852352 as advised in https://issuetracker.google.com/issues/216744701, 

2 attempts still failed out of 10 attempts.

Now I enabled more logging in this set of Actions.
https://github.com/android/nowinandroid/actions/runs/3683976750

If we still see failed jobs, I'm considering  filing a bug against GMD.

thagikura:
The last 10 attempts succeeded as in https://github.com/android/nowinandroid/actions/runs/3683976750

Do you think it's acceptable @dturner ?
In case if it fails in the future, I also enabled the detailed logging to pass it to the product team.

Also this PR only enables GMD in the core-database module, the rest of the modules need to be addressed, too.

dturner (approved):

core/database/build.gradle.kts
```diff
@@ -34,6 +34,18 @@ android {
         testInstrumentationRunner = "com.google.samples.apps.nowinandroid.core.testing.NiaTestRunner"
     }
     namespace = "com.google.samples.apps.nowinandroid.core.database"
+
+    testOptions {
```
dturner, line 38: Can we move this into a common plugin so that GMD is used for every module?

dturner (changes requested):
Let's try using GMD everywhere so we can compare it with ReactiveCircus for the whole codebase.

core/database/build.gradle.kts
thagikura: There was a bug where defining GMD wasn't possible in Kotlin that made impossible to define it as a convention plugin. (b/242515648)

That is now fixed, but is included in Flamingo canary 9.

thagikura:
> Let's try using GMD everywhere so we can compare it with ReactiveCircus for the whole codebase.

Agreed. Let's see the flakiness to compare with ReactiveCircus.

thagikura:
> Agreed. Let's see the flakiness to compare with ReactiveCircus.

Added GMD to modules that have instrumented tests. Can you have a look again?

.github/workflows/AndroidCIWithGmd.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+
+  android-ci:
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
```
alexvanyo, line 17: I don't know if it matters, but what is the reason for `temurin` here? I think in our other workflows we have `zulu`

.github/workflows/AndroidCIWithGmd.yaml
```diff
@@ -0,0 +1,34 @@
...
+        java-version: '11'
+    - uses: actions/checkout@v2
+
+    - name: Setup Android SDK
+      uses: android-actions/setup-android@v2
```
alexvanyo, line 22: Is this necessary? I don't see it in the other build scripts we have.

core/database/build.gradle.kts
alexvanyo: I was going to suggest the same, thanks for the info. Can we create a TODO and a GitHub issue to track?

alexvanyo (approved):
Woo! 🎉 

Screen Shot 2023-01-04 at 2 29 53 PM

Looks good to me, just two nits in the workflow and it's unfortunate we can't make a convention plugin at the moment. Ideally we could apply that plugin everywhere, so we can't make a mistake of leaving out any new tests by not adding the managed devices configuration.

.github/workflows/AndroidCIWithGmd.yaml
thagikura: I don't have a strong opinion here either. Let's stick with other workflow files. Changed this to `zulu`.

.github/workflows/AndroidCIWithGmd.yaml
thagikura: Thanks. It wasn't necessary. The last attempt succeeded without this plugin.

thagikura:
Thanks Alex. Addressed your feedback. 
Filed #523 as a reminder to convert GMD definitions as a convention plugin.
@dturner is it possible to have a look again and resolve the change request if it looks good?

Thanks!

app/build.gradle.kts
```diff
@@ -73,6 +73,7 @@ android {
         unitTests {
             isIncludeAndroidResources = true
         }
+        // TODO: Convert it as a convention plugin once Flamingo goes out (b/242515648)
```
alexvanyo, line 76: Nit: Let's avoid links to bugs that aren't publicly accessible, we could link to the newly created GitHub issue instead.

thagikura: Done. Sgtm!

thagikura:
Friendly reminder to @dturner please resolve the change request if the PR looks good. Thanks!

alexvanyo (approved):

dturner (approved):
Many thanks for doing this.

PR #490, opened 2022-12-15
Center top app bar when nav rail is visible

alexvanyo:
Moves the top app bar inside the row with the navigation rail, so that it doesn't overlap with the navigation rail.


dturner (approved):

dturner:
It's no longer painful to look at :D 

image

Issue #349, opened 2022-10-21
[FR]: NiaApp background composable could be improved


dturner:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

Context: https://github.com/android/nowinandroid/pull/330/files/ee4eb25ba07e7a017c52f54fa49547a0e733ff8b#r1001237941

From @manuelvicnt:
What's this complexity? Could this be a slot API instead? And the content inside it could be using moveableContent so that state isn't lost

### Describe the solution

Let's discuss this in our weekly. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

emitchel:
Was this referencing how changing the background gradient by screen makes the navigation more janky? Example, going from ForYou <-> Saved, it's a bit jarring, no soft transition because ForYou has a custom `background`. Saved <->Interests has a pleasant transition because they have the same background, thus no scaffolding recomposition. 

Wonder if there's a better way to go about this, otherwise, we should just keep the background local to the screen.

alexvanyo:
The original comment was for the non-standard setup for the background that looks like `val background: @Composable (@Composable () -> Unit) -> Unit = // ...`

The setup here is a bit tricky, since we want the background to be displayed behind some shared elements (like the nav bar), but it is different depending on the destination like you mentioned.

We should be able to replace the current setup with one that uses `movableContentOf`, to ensure that the state of the application isn't lost when the background in use changes.

PR #491, opened 2022-12-15
Updated background composable for gradient

### LINKED ISSUE #415: [Bug]: Scroll position is not maintained when switching screens from bottom nav.

### What happened?

when going back from saved screen to home screen scrolled position is not maintained.

### Relevant logcat output

```shell
Scroll position should be preserved
```

### Code of Conduct

alexvanyo:
This PR changes how the overall background composable is applied, to ensure that state isn't lost when switching between the two.

One option here would be `movableContentOf`, but that would be more complicated and when trying that out, was running into issues with the animation looking smooth.

The other option which this PR uses is always calling into both background functions, but effectively disabling the gradient one if we don't actually want to show it.

This fixes #415 and #349

### DIFF
--- a/app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
+++ b/app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
@@ -87,21 +87,22 @@ fun NiaApp(
         windowSizeClass = windowSizeClass
     ),
 ) {
-    val showGradientBackground = appState.currentTopLevelDestination == TopLevelDestination.FOR_YOU
+    val shouldShowGradientBackground =
+        appState.currentTopLevelDestination == TopLevelDestination.FOR_YOU
 
     NiaBackground {
         NiaGradientBackground(
-            topColor = if (showGradientBackground) {
+            topColor = if (shouldShowGradientBackground) {
                 LocalGradientColors.current.top
             } else {
                 Color.Unspecified
             },
-            bottomColor = if (showGradientBackground) {
+            bottomColor = if (shouldShowGradientBackground) {
                 LocalGradientColors.current.bottom
             } else {
                 Color.Unspecified
             },
-            containerColor = if (showGradientBackground) {
+            containerColor = if (shouldShowGradientBackground) {
                 LocalGradientColors.current.container
             } else {
                 Color.Unspecified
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -86,109 +87,119 @@ fun NiaApp(
...
 ) {
-    val background: @Composable (@Composable () -> Unit) -> Unit =
-        when (appState.currentTopLevelDestination) {
-            TopLevelDestination.FOR_YOU -> {
-                content ->
-                NiaGradientBackground(content = content)
-            }
-            else -> { content -> NiaBackground(content = content) }
-        }
+    val showGradientBackground = appState.currentTopLevelDestination == TopLevelDestination.FOR_YOU
```
dturner, line 90: nit: boolean variable names should be prefixed with `is|should|has`

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -86,109 +87,119 @@ fun NiaApp(
...
 
-    background {
+    NiaBackground {
+        NiaGradientBackground(
+            topColor = if (showGradientBackground) {
```
dturner, line 94: I've refactored this in #502

dturner (commented):
LGTM with one nit and one refactor suggestion

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
alexvanyo: Updated!

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
alexvanyo: Sounds good, do you want to merge that PR into this one before merging this one, or merge #502 after this is merged?

dturner: Merge into this one - makes sense to keep it as a single PR

alexvanyo: Added your commit to this branch!

dturner (approved):

PR #485, opened 2022-12-15
Update :core-designsystem module based on Figma file

ricknout:
Various updates to the :core-designsystem module to align with the Figma file:
https://www.figma.com/community/file/1164313362327941158

Issue #422, opened 2022-11-11
[Bug] Changing theme doesn't update tab text color


dturner:
(thanks to the user who reported this bug through the app via goo.gle/nia-app-feedback)

Changing from dark to light theme doesn't update the the text on the tabs.

Steps to repro: 

- Tap on the Settings icon, switch to the opposite theme to the current one

Expected result: 

- Color of text in the tabs at the bottom of the screen changes to one which is easily readable

Actual result: 

- Color of text in tabs at bottom of screen doesn't change making it difficult to read

![image](https://user-images.githubusercontent.com/873212/201386889-282c2791-9498-4322-87b8-db73ffd110a1.png)

RivanParmar:
In my case, on changing the theme from "Default" to "Android" or vice-versa, the color of the currently selected tab remains the same and only changes after clicking on other tab. The device I use runs on Android 10.

![Screenshot_20221111-232205.png](https://user-images.githubusercontent.com/77159650/201399871-ae44f535-5c2a-4250-8602-e0159c056e32.png)

As it can be seen in the above screenshot, I had first set the theme to "Android" and then changed it to "Default" and so all the other elements have got the default colors while the selected tab still has the "Android" theme color.

LinX64:
I was also facing the same issue, to reproduce, try to change the theme ASAP, and boom!...

SaeedNoshadi89:
In this situation, if we select the other bottom bar item the them will be fixed

alexvanyo:
Related upstream issue: https://issuetracker.google.com/issues/254918865

alexvanyo:
This should be fixed when we update to `1.4.0-beta01` or higher for dependencies.

PR #492, opened 2022-12-16
Fix issues with NiaTheme gradients and backgrounds

ricknout:
Fixes issues introduced in #485


dturner:
It looks like when changing themes the background colour of the navigation buttons isn't changed: 
image

dturner:
You're right. It's this bug https://github.com/android/nowinandroid/issues/422 which I hadn't realised also affects the theme color as well as dark/light mode.

dturner (approved):

alexvanyo (approved):
Looks good!

PR #493, opened 2022-12-17
Set visibility modifier of composable preview to private

andhiratobing:
Because it is a composable preview, it's better to make it private because it's not a public function


dturner (approved):

dturner:
Thanks very much for your contribution

PR #496, opened 2022-12-17
Exclude `baseline-prof.txt` from IntelliJ IDEA indexing

SimonMarquis:
... as it pollutes search results and wastes precious indexing time.

Here is how it will appear after this change:

image

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -22,6 +22,7 @@ plugins {
     id("nowinandroid.android.hilt")
     id("jacoco")
     id("nowinandroid.firebase-perf")
+    id("idea")
 }
 
 android {
@@ -126,5 +127,12 @@ configurations.configureEach {
         force(libs.junit4)
         // Temporary workaround for https://issuetracker.google.com/174733673
         force("org.objenesis:objenesis:2.6")
+    }
+}
+
+// Exclude baseline-prof.txt from indexing
+idea {
+    module {
+        excludeDirs = excludeDirs + file("src/main/baseline-prof.txt")
     }
 }


wojtek-kalicinski:
Thank you for the contribution, but I recommend against taking this. AFAIK, Android Studio doesn't index baseline-prof.txt files, and it's our recommended IDE for use with Android projects. Let's not add additional Gradle plugins that generate IDEA project files (as I understand) just for this.

SimonMarquis:
Android Studio does the same thing as IntelliJ. On my machine, search results are coming from the baseline profile file.

SimonMarquis:
For instance, if I search for `OnBackPressedDispatcher`, it will show up 55 results from the `baseline-prof.txt` file.
image

G00fY2:
@SimonMarquis there is this ticket in the issue tracker from April last year: https://issuetracker.google.com/issues/229980813

This PR would be a workaround until this is fixed, right? Maybe Google needs to raise the priority of this issue.:stuck_out_tongue_winking_eye:

keyboardsurfer:
Thanks for raising this issue again. I second rejecting this PR in light of incoming changes to AGP 8.0.0, where baseline profiles can be moved to `src/main/baselineProfiles` which then could be excluded from indexing.

SimonMarquis:
Thanks for the link @G00fY2
I agree, this is just a workaround, and should not be our responsibility, but from what I understand from that issue tracker comment:

> As most developers won't need to search their profiles, we could disable indexing this location by default.

**we could** does not mean it is going to be done. Not even if it is tracked somewhere.

keyboardsurfer:
> we could does not mean it is going to be done.
That's correct.
> Not even if it is tracked somewhere.
I added it to the issue mentioned above.

PR #501, opened 2022-12-21
added a string resources class UiText.kt in core:common this function…

ALI-M0vS5a:
… can be used in the entire project in order to have access to string resources also i have refactored the core:module by adding a manifest and the access to the core:common in core:common in the NewsResourceType.kt i have used the UiText.kt to change the description to stringResources

### DIFF
--- a/core/common/src/main/java/com/google/samples/apps/nowinandroid/core/uitext/UiText.kt
+++ b/core/common/src/main/java/com/google/samples/apps/nowinandroid/core/uitext/UiText.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package com.google.samples.apps.nowinandroid.core.uitext
+
+import android.content.Context
+import androidx.annotation.StringRes
+import androidx.compose.runtime.Composable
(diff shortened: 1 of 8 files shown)


dturner:
Hey thanks for submitting this. It looks like a great solution.

dturner:
Please run `./gradlew --init-script gradle/init.gradle.kts spotlessApply` to fix the formatting issues.

ALI-M0vS5a:
hello.
should i remove the org.gradle.java.home=C\:\\Program Files\\Java\\jdk-11 from gradle.properties after running this 
./gradlew --init-script gradle/init.gradle.kts spotlessApply successfully?

alexvanyo (commented):
I'm a bit confused by this PR, since it looks like `asString` and `unknownError` are both unused, and instances of `DynamicString` are never created.

Is the goal to have a more robust string mechanism for when we do need those methods?

core/common/src/main/java/com/google/samples/apps/nowinandroid/core/uitext/UiText.kt
```diff
@@ -0,0 +1,51 @@
...
+
+package com.google.samples.apps.nowinandroid.core.uitext
+
+import android.content.Context
+import androidx.annotation.StringRes
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import com.google.samples.apps.nowinandroid.core.common.R
+
+sealed class UiText {
```
mmoczkowski, line 25: This wrapper class is unnecessary noise and complexity imo

core/model/src/main/java/com/google/samples/apps/nowinandroid/core/model/data/NewsResourceType.kt
```diff
@@ -16,59 +16,61 @@
...
+
 /**
  * Type for [NewsResource]
  */
 enum class NewsResourceType(
     val serializedName: String,
     val displayText: String,
-    // TODO: descriptions should probably be string resources
-    val description: String
+    val description: UiText
```
mmoczkowski, line 28: I think it's more beneficial to simply store a resource id here

mmoczkowski (changes requested):
Remove the wrapper class and make description field a resource id instead of a string

mmoczkowski:
Closing due to inactivity. If you still wish to submit this change, please create a new PR.

Issue #438, opened 2022-11-15
[Bug]: Warning message ksp-1.7.21-1.0.8 is too new for kotlin-1.7.20


SimonMarquis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

```
ksp-1.7.21-1.0.8 is too new for kotlin-1.7.20. Please upgrade kotlin-gradle-plugin to 1.7.21.
```

Example in GitHub Actions: https://github.com/android/nowinandroid/actions/runs/3473866639/jobs/5806381926#step:9:23

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mmoczkowski:
Thanks Simon, let's wait for the next stable compose version and bump Kotlin to 1.7.21 afterwards

PR #503, opened 2022-12-21
Update library versions

dturner:


alexvanyo (approved):

alexvanyo:
This should fix #438 as well

PR #508, opened 2022-12-22
Update topic chips to show a different color based on whether the user is following that topic 

MagicalMeghan:
Changing the color of topic chip if the topic is followed or not

Bug here: https://b.corp.google.com/issues/259573931


dturner:
We should probably: 

#1 Refactor GetSaveableNewsResourceUseCase to move the mapping of UserData + NewsResource -> UserNewsResource into a separate function which we can test independently. This will make the tests a lot simpler and avoid duplication. 
#2 Refactor GetSaveableNewsResourceUseCaseTest to only test the behaviour of GetSaveableNewsResourceUseCase: 

- returning all UNRs when no filter
- returning filtered list of UNRs

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -239,17 +241,27 @@ fun NewsResourceTopics(
...
                 onBrowseClick = { }, // ToDo
-                text = { Text(text = topic.name.uppercase(Locale.getDefault())) }
+                text = {
+                    Text(
+                        text = followableTopic.topic.name.uppercase(Locale.getDefault()),
+                        modifier = Modifier.semantics {
+                            contentDescription = followableTopic.topic.name +
+                                if (followableTopic.isFollowed) " is followed"
+                                else " is not followed"
+                            // make sure this is the right place to add this - remove hardcoded strings
```
dturner, line 261: @MagicalMeghan to fix

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardList.kt
```diff
@@ -37,7 +38,7 @@ import com.google.samples.apps.nowinandroid.core.model.data.NewsResource
  */
 fun <T> LazyListScope.newsResourceCardItems(
     items: List<T>,
-    newsResourceMapper: (item: T) -> NewsResource,
+    newsResourceMapper: (item: T) -> UserNewsResource, // TODO remove this?
```
dturner, line 41: We should address this before merging

dturner: @tunjid Can't wait to have you back on this project 😄 ... 

Why was this function implemented as a generic function with a mapper? Presumably this was to allow `item` to be mapped to a `NewsResource` regardless of its type (via `newsResourceMapper`), but I'm not sure how this was ever applicable. For now, I'm going to remove the generic type to simplify this function but wanted to leave a note here so we can discuss at some future date.

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -228,28 +230,40 @@ fun NewsResourceShortDescription(
...
 @Composable
 fun NewsResourceTopics(
-    topics: List<Topic>,
+    topics: List<FollowableTopic>,
     modifier: Modifier = Modifier
 ) {
     // Store the ID of the Topic which has its "following" menu expanded, if any.
     // To avoid UI confusion, only one topic can have an expanded menu at a time.
     var expandedTopicId by remember { mutableStateOf<String?>(null) }
+    val isFollowed = stringResource(R.string.topic_is_followed)
```
dturner, line 239: `isFollowed` implies that this is a boolean, but it's actually text. These variables should be suffixed with `Text` or `ContentDescription`. 

Also, it's best to keep variables as close as possible to where they are used so they can be moved into the `text = { ` block.

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -228,28 +230,40 @@ fun NewsResourceShortDescription(
...
                 onFollowClick = { }, // ToDo
                 onUnfollowClick = { }, // ToDo
                 onBrowseClick = { }, // ToDo
-                text = { Text(text = topic.name.uppercase(Locale.getDefault())) }
+                text = {
+                    Text(
+                        text = followableTopic.topic.name.uppercase(Locale.getDefault()),
+                        modifier = Modifier.semantics {
+
+                            contentDescription = followableTopic.topic.name + " " +
```
dturner, line 261: There's a better solution here using string substitution. I will push it so you can see what I mean.

dturner (approved):
Thanks for all your work on this. Congrats on your first feature. Fine to merge once CI passes.

Issue #510, opened 2022-12-23
Compact is misspelled as compcat in NavigationUiTest.kt


anasshamsi97:
compcatWidth_mediumHeight_showsNavigationBar is misspelled in NavigationUiTest.kt which should actually be compactWidth_mediumHeight_showsNavigationBar.

alexvanyo:
Fixed with #511 , thank you!

PR #511, opened 2022-12-23
Typo fix for compact

anasshamsi97:
#510 compcatWidth_mediumHeight_showsNavigationBar is misspelled in NavigationUiTest.kt which should actually be compactWidth_mediumHeight_showsNavigationBar.


alexvanyo (approved):
Thanks for catching that!

PR #512, opened 2022-12-23
Backup and restore a12 and A11 rules for User backing up user preferences

niharika2810:


alexvanyo (commented):
Our current user preferences include information related to the database. In particular, they include the change list information for which information we have synced.

If we only backup the user preferences as-is, without also backing up the database, the preferences will indicate that we have synced data that we actually don't have, likely leading to a broken experience.

Other options are:
- splitting out the change list information into other files, so we can safely back up other preferences
- backup the database as well

mmoczkowski:
Closing due to inactivity. If you still wish to submit this change you can create a new PR.

PR #513, opened 2022-12-27
Fix multi touch issue

nntsl:
Touch events were dispatched to multiple items. 
Now they will be dispatched to the topic item where the first pointer went down until the last pointer goes up.


dturner:
Thanks for submitting this. Before merging I'd like to understand better what problem this PR is attempting to solve. Please raise an issue with reproduction steps and reference this PR.

dturner:
Also, please fix the format violations by running: 

```
./gradlew --init-script gradle/init.gradle.kts --no-configuration-cache spotlessApply
```

alexvanyo (changes requested):
I agree that this would be nice to fix, but I am strongly opposed to doing so by modifying the inputs in this manner for a couple of reasons:

- Disallowing multi-touch might break gestures that use multi-touch in unforseen ways
- Filtering out some user events doesn't really solve the problem, which is that we are  allowing multiple launches out of our app at the same time. Even with these changes, if you double-tap very quickly, the same issue remains.

A more complete approach here is to avoid launching another topic while we know we're already in the process of leaving, which we can tell by checking if the lifecycle of the current `NavBackStackEntry` is still resumed. Upon navigating, the lifecycle state will immediately be updated, so the callback for any additional clicks would no-op.

PR #521, opened 2023-01-04
Move construction of UserNewsResource into separate function

dturner:


core/domain/src/main/java/com/google/samples/apps/nowinandroid/core/domain/model/UserNewsResource.kt
```diff
@@ -39,9 +40,30 @@ data class UserNewsResource(
     val headerImageUrl: String?,
     val publishDate: Instant,
     val type: NewsResourceType,
-    val topics: List<FollowableTopic>,
+    val followableTopics: List<FollowableTopic>,
     val isSaved: Boolean
-)
+) {
+    companion object {
+        fun from(newsResource: NewsResource, userData: UserData): UserNewsResource {
```
mmoczkowski, line 47: Would you consider moving it to a separate file in order do increase readability of UserNewsResource?

dturner:
I will fix the failing UI tests on the mm/chipcolor branch

PR #524, opened 2023-01-07
Set item keys for smart recomposition of topics

andhiratobing:
Set keys to prevent problems if the data set changes.


mmoczkowski (approved):
LGTM, thank you for your contribution Andhi.

PR #483, opened 2022-12-13
Add gmd configuration

AndreSand:
Testing: Android GtiHub actions runs the Gradle task:
./gradlew pixel6Api31DemoDebugAndroidTest

PR #537, opened 2023-01-16
Android-CI workflow -> GMD refactor

LinX64:
This PR is about updating the operating system to macOS-12 as the previous one was going to be deprecated, as well as updating the checkout to v3 to avoid the warnings in the panel; adding setup SDK action to accept the license which was missing from the beginning.

I also added imports instead of adding them directly with the codes.

As a reference, there was another PR with a similar topic, but it wasn't complete: https://github.com/android/nowinandroid/pull/483

### DIFF
--- a/.github/workflows/AndroidCIWithGmd.yaml
+++ b/.github/workflows/AndroidCIWithGmd.yaml
@@ -9,14 +9,14 @@ on:
 jobs:
 
   android-ci:
-    runs-on: macos-latest
+    runs-on: macos-12
 
     steps:
     - uses: actions/setup-java@v3
       with:
         distribution: 'zulu'
         java-version: '11'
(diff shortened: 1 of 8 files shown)


LinX64:
@thagikura I saw the logs and realized that there is an issue but since you've added the `continue-on-error: true` property, it was showing successful most of the time. So, I've added the Setup SDK, where it is accepting the licenses and then removed that property, now it should be working fine. Please let me know what you think of this!

thagikura (approved):
LGTM. Thanks for doing this!

thagikura:
Thanks! 

@alexvanyo or @dturner, is it possible to have another look before merging?

alexvanyo (approved):

PR #540, opened 2023-01-18
Remove topic chip popup menus, route topic click events up to NiaNavHost

zsmb13:
Topic chips no longer display a popup menu when tapped. Their tap events are routed up to `NiaNavHost`, to be handled with navigation later.


core/designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Tag.kt
```diff
@@ -67,25 +57,6 @@ fun NiaTopicTag(
                 text()
             }
         }
-        NiaDropdownMenu(
```
zsmb13, line 70: Note: After the removal of this here, `NiaDropdownMenu` and `NiaDropdownMenuButton` don't have any usages in the app (only in the catalog).

manuelvicnt: I don't know how the UI looked like previously, do you mind adding a before/after screenshot of this feature to the PR description? Thanks

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -176,6 +179,7 @@ internal fun ForYouScreen(
         newsFeed(
             feedState = feedState,
             onNewsResourcesCheckedChanged = onNewsResourcesCheckedChanged,
+            onTopicClick = onTopicClick,
```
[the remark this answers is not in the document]
zsmb13: This was on purpose (though I did not go out of my way to add it everywhere in every file I touched). Look at all the lines that this PR touched just because they didn't have trailing commas!

@dturner to chime in on whether there's a convention? 😄

core/designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Tag.kt
zsmb13: That is the question 😄  If we're not expecting to use them very soon, I'd remove them.

Updated the PR description with a _Before_ video.

dturner (commented):
Looking good. Are we happy that when you're on the topic details screen and you tap on a topic chip _of the same name_ you can navigate to a duplicate topic details screen (albeit with different scroll position)? 

Personally, I'm OK with this but want to make sure this decision is thought about and documented.

manuelvicnt (approved):

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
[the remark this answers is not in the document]
zsmb13: Following this up in another PR.

core/designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/Tag.kt
zsmb13: Removed!

dturner:
Please add UI tests to verify this behaviour in `NavigationTest`

internalG:
It seems coding tests is too hard to a developer...😁

dturner (approved):

dturner:
Just to add some clarity here (mostly for me). There is nothing to test at the moment. This PR simply replaces the topic chip's dropdown menu with an onClick lambda which currently does nothing. 

The outstanding issue with implementing the topic chip navigation is that it creates a new way of showing the topic detail screen. This can have some unwanted side effects (depending on implementation) such as changing the selected tab to "Interests" even though the user tapped on a topic chip from the "ForYou" screen.

Needs more thought/discussion on what the best approach is for users, and how that should be implemented using navigation best practices.

PR #542, opened 2023-01-19
Add dynamic color option in settings for API >=32

angiesasmita:
[NiA Dynamic Color Option After.webm](https://user-images.githubusercontent.com/10851617/213369000-ed2bd639-b6be-4e4e-9fc9-b9152b660df1.webm)


core/designsystem/src/androidTest/java/com/google/samples/apps/nowinandroid/core/designsystem/ThemeTest.kt
```diff
@@ -241,6 +263,18 @@ class ThemeTest {
         )
     }
 
+    private fun defaultTintTheme(): TintTheme {
+        return TintTheme()
+    }
+
+    private fun dynamicTintTheme(colorScheme: ColorScheme): TintTheme {
```
alexvanyo, line 270: Nit: Maybe name this `dynamicTintThemeWithFallback` for consistency with the other methods?

feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsDialog.kt
```diff
@@ -71,19 +76,33 @@ fun SettingsDialog(
...
     onChangeDarkThemeConfig: (darkThemeConfig: DarkThemeConfig) -> Unit
 ) {
+    val configuration = LocalConfiguration.current
 
+    /**
+     * usePlatformDefaultWidth = false is use as a temporary fix to allow
+     * height recalculation during recomposition. This, however, causes
+     * Dialog's to occupy full width in Compact mode. Therefore max width
+     * is configured below. This should be removed when there's fix to
+     * b/221643630
```
alexvanyo, line 101: Nit: I think its better to use the full URL that is publicly accessible: https://issuetracker.google.com/issues/221643630

feature/bookmarks/src/main/res/drawable/img_empty_bookmarks.xml
```diff
@@ -1,31 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
alexvanyo, line 1: Mentioned in chat, let's keep this as an SVG if possible for better resizing support.

core/designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/theme/Tint.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.ui.graphics.Color
+
+/**
+ * A class to model background color and tonal elevation values for Now in Android.
+ */
+@Immutable
+data class TintTheme(
+    val iconTint: Color? = null,
```
alexvanyo, line 28: Nit: Could potentially use `Color.Unspecified` here, as opposed to `null` for the default case.

feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsDialog.kt
```diff
@@ -71,19 +76,33 @@ fun SettingsDialog(
...
+     */
     AlertDialog(
+        properties = DialogProperties(usePlatformDefaultWidth = false),
+        modifier = Modifier.widthIn(max = configuration.screenWidthDp.dp - 80.dp),
```
alexvanyo, line 105: I'm scared this might lead to an awkwardly large dialog when fullscreen on foldables/tablets

feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsDialog.kt
```diff
@@ -145,7 +169,22 @@ private fun SettingsPanel(
             onClick = { onChangeThemeBrand(ANDROID) }
         )
     }
-    SettingsDialogSectionTitle(text = "Dark mode preference")
+    if (settings.brand == DEFAULT && supportDynamicColor) {
+        SettingsDialogSectionTitle(text = stringResource(R.string.dynamic_color_preference))
+        Column(Modifier.selectableGroup()) {
+            SettingsDialogThemeChooserRow(
+                text = stringResource(string.dynamic_color_yes),
+                selected = settings.useDynamicColor == true,
```
alexvanyo, line 177: Nit: can this be simplified to:
```suggestion
                selected = settings.useDynamicColor,
```

feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsDialog.kt
```diff
@@ -145,7 +169,22 @@ private fun SettingsPanel(
...
+                onClick = { onChangeDynamicColorPreference(true) }
+            )
+            SettingsDialogThemeChooserRow(
+                text = stringResource(string.dynamic_color_no),
+                selected = settings.useDynamicColor == false,
```
alexvanyo, line 182: Nit: Same here, can this be simplified to:
```suggestion
                selected = !settings.useDynamicColor,
```

alexvanyo (commented):
This looks great!

Just a few nits, especially to confirm the dialog workaround and that we want to have dynamic theming disabled by default (even when supported)

feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsDialog.kt
angiesasmita: I tested this with other screen sizes, it only affects the Compact size somehow. Also why I added the fix for max width only :)

core/designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/theme/Tint.kt
angiesasmita: Since this is passed to `ColorFilter`, Color.Unspecified causes the icon to disappear. Sticking with null here.

angiesasmita:
@alexvanyo yup, dynamic theming was decided to be disabled by default to allow the NiA branding to show as the main theme as it is optimised to show case the beautiful design.

core/designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/theme/Tint.kt
alexvanyo: Sounds good! I think we could keep the same logic and use `Color.isSpecified` to choose whether to use the `ColorFilter` or not, but I don't feel strongly either way.

feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsDialog.kt
alexvanyo: Oh nice! I tested it and verified that's the case on a tablet, I also found where that's being set:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/AlertDialog.kt;l=215;drc=cf950e6a8ecdc4a3281bb7295f02a98e16798d49

Looks good!

alexvanyo (approved):
Looks good to me, only remaining thing is if we can get the SVG that works!

PR #548, opened 2023-01-26
Fix bookmarking scrolling to top of the page bug

### LINKED ISSUE #426: [Bug]: The screen goes up after bookmarking an item

### What happened?

When I bookmark an article on the home page, the page scrolls up and returns to the default position.

https://user-images.githubusercontent.com/50018300/201514331-e2aa7885-855b-440b-8367-0395fbc8b5de.mp4

### Relevant logcat output

_No response_

### Code of Conduct

manuelvicnt:
Fixes #426 

Bug fixed

### DIFF
--- a/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
+++ b/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
@@ -32,7 +32,6 @@ import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.launch
 import javax.inject.Inject
@@ -76,7 +75,6 @@ class ForYouViewModel @Inject constructor(
             // As the selected topics and topic state changes, this will cancel the old feed
             // monitoring and start the new one.
             .flatMapLatest { it }
-            .onStart { emit(NewsFeedUiState.Loading) }
             .stateIn(
                 scope = viewModelScope,
                 started = SharingStarted.WhileSubscribed(5_000),
--- a/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
+++ b/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
@@ -68,13 +68,15 @@ class ForYouViewModel @Inject constructor(
                 } else {
                     getSaveableNewsResources(
                         filterTopicIds = userData.followedTopics,
-                    ).mapToFeedState()
+                    )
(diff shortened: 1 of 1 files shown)


feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
```diff
@@ -68,13 +68,15 @@ class ForYouViewModel @Inject constructor(
...
-                    ).mapToFeedState()
+                    )
+                        .map<List<UserNewsResource>, NewsFeedUiState>(NewsFeedUiState::Success)
                 }
             }
             // Flatten the feed flows.
             // As the selected topics and topic state changes, this will cancel the old feed
             // monitoring and start the new one.
             .flatMapLatest { it }
+            .onStart { emit(NewsFeedUiState.Loading) }
```
tunjid, line 79: If we still use `onStart`, if we leave the bookmarks screen for > 5 seconds, the bug will show up again. Can we drop  emitting `Loading` each time we resume the stream? It wipes out the cached state in the `StateFlow`.

dturner: I don't think we need `Async`. The initial value for the news feed is already set to `Loading`. 

The only scenario I can think of where we'd want to show a loading indicator to the user again would be during a network sync, but this is already handled by a separate flow in `isSyncing`.

alexvanyo: +1, or at least changing the types here so that it is more clear what behavior we want.

Right now, we clear out the old data when we are loading, which ends up leading to behavior that feels weird. The behavior we want is that we can continue showing the old data while we are loading (but we still want to indicate that we are reloading).

I feel like we should allow the `Loading` state to also have a list of items to display, which would be the set of old items while we are loading new ones.

hoc081098: @tunjid
StateFlow will not reset cache value to Loading (it will reset if we use `WhileSubscribed(replayExpiration = Duration.ZERO)`.

_Sent from my 2201117TG using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

dturner: @hoc081098 `onStart` will be called when the flow is restarted, which will happen if the app is backgrounded for >5 seconds and then brought to the foreground. This will cause the flow to emit a `Loading` state.

hoc081098: My bad 😂

manuelvicnt: I think we were overthinking this... unless I'm missing something obvious `onStart` isn't needed there. `stateIn`'s `initialValue` should already give us the loading signal while data is fetched for the first time

dturner (approved):
Thanks for fixing this 👍

tunjid (approved):

PR #558, opened 2023-02-03
Switch GMD to use a convention plugin

alexvanyo:
Pulls out the GMD configuration into a convention plugin, so that we have a single place to define devices and configure all modules to use them.

### DIFF
--- a/.github/workflows/AndroidCIWithGmd.yaml
+++ b/.github/workflows/AndroidCIWithGmd.yaml
@@ -10,6 +10,9 @@ jobs:
 
   android-ci:
     runs-on: macos-12
+    strategy:
+      matrix:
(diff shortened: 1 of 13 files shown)


thagikura (commented):
Thanks for doing this!

The change looks good, but looking at the recent attempt, https://github.com/android/nowinandroid/actions/runs/4088840030/jobs/7050913035

the duration of the execution is nearly doubled. Is it possible to have a look if it's related to the convention plugin?

alexvanyo:
With the convention plugin, I'm applying GMD to all modules now, instead of just those that have Android tests. Maybe that's the reason for the extra overhead?

thagikura (approved):
I see. Thanks for the clarification

PR #561, opened 2023-02-04
Updgrade androidx-lifecycle from 2.6.0-alpha03 to 2.6.0-alpha05

SimonMarquis:
> `Transformations` is now written in Kotlin. This is a **source incompatible change** for those classes written in Kotlin that were directly using syntax such as `Transformations.map`
> Kotlin code **must** now use the Kotlin extension method syntax that was previously only available when using `lifecycle-livedata-ktx`.

> The `collectAsStateWithLifecycle()` APIs of `lifecycle-runtime-compose` are no longer in experimental status. (I09d42, b/258835424)

Changelog: https://developer.android.com/jetpack/androidx/releases/lifecycle#2.6.0-alpha05


SimonMarquis:
There is a class collision between `kotlin-stdlib-1.8.0` (`org.jetbrains.kotlin:kotlin-stdlib:1.8.0`) and `kotlin-stdlib-jdk8-1.7.21` (`org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.21`).

Let's wait for:
- https://github.com/android/nowinandroid/pull/564

alexvanyo:
@SimonMarquis since #564 was merged, I think this could be rebased now?

alexvanyo (approved):

PR #567, opened 2023-02-04
Migrate last build.gradle to build.gradle.kts

SimonMarquis:


alexvanyo (approved):
Thanks for the contribution here!

Issue #534, opened 2023-01-13
[Bug]: Multiple fingers can touch the screen at the same time


nntsl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

It is possible to click multiple items at TopicsTabContent once (the same happens with People tab). 
If we press with several fingers at once, all screens will open. Going back we will see all of them.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

vishesh211:
@nntsl  I reproduced the bug on my android too. There is this stackoverflow answer for a similar question which uses a extension. I added the code and tested it on my system. It successfully killed the bug and the interests screen was registering one touch at a time. 

I'll paste the link here. You can look it up yourself.
StackOverflow link:-  https://stackoverflow.com/a/72816456

vishesh211:
hi @dturner i am new to this library. I have fixed this bug and created a pull request. Can you guide me on what i need to do next. Thank you it will help me out a lot.

chenlong0225:
Use **launchSingleTop = true**
![image](https://user-images.githubusercontent.com/120089153/227418333-853c9c68-02a8-4464-9552-7e12caec3782.png)
![image](https://user-images.githubusercontent.com/120089153/227418408-8e5fbbe8-fbb4-47c4-a1ed-dcca667ce842.png)

PR #568, opened 2023-02-04
Vishesh211/bug 534 interests multiple click

vishesh211:
Closes: https://github.com/android/nowinandroid/issues/534

Multiple Finger were opening multiple screens at the same time in the interests tab. The fix i have added is to create an extension with modifier that detects only one touch that happens first and then opens that particular screen.


mmoczkowski (changes requested):
Thank you for attempting to fix this issue. However, I think a better approach would be to use navoptions with launchSingleTop

vishesh211:
Okay. Working on it.

vishesh211:
@mmoczkowski Have made the changes you said. Please review it.

build.gradle.kts
```diff
@@ -26,6 +26,8 @@ buildscript {
 }
 
 plugins {
+    java
```
mmoczkowski, line 29: not needed, please remove

feature/interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/TabContent.kt
```diff
@@ -25,22 +25,25 @@ import androidx.compose.foundation.layout.windowInsetsBottomHeight
...
     onFollowButtonClick: (String, Boolean) -> Unit,
-    modifier: Modifier = Modifier,
+    modifier: Modifier = Modifier
 ) {
     LazyColumn(
         modifier = modifier
             .padding(horizontal = 16.dp)
             .testTag("interests:topics"),
-        contentPadding = PaddingValues(top = 8.dp),
+        contentPadding = PaddingValues(top = 8.dp)
```
mmoczkowski, line 46: Let's keep the trailing commas

feature/topic/src/main/java/com/google/samples/apps/nowinandroid/feature/topic/navigation/TopicNavigation.kt
```diff
@@ -37,7 +37,9 @@ internal class TopicArgs(val topicId: String) {
 
 fun NavController.navigateToTopic(topicId: String) {
     val encodedId = Uri.encode(topicId)
-    this.navigate("topic_route/$encodedId")
+    this.navigate("topic_route/$encodedId") {
+        launchSingleTop = true
```
mmoczkowski, line 41: 👌

mmoczkowski (changes requested):
Hey, let's remove the unnecessary stuff and we're good 😄

vishesh211:
Okay sure thanks

build.gradle.kts
vishesh211: Done

feature/interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/TabContent.kt
vishesh211: Done

vishesh211:
@mmoczkowski did the cleanup. If anything else is required. Do tell me. :)

mmoczkowski (changes requested):

build.gradle.kts
```diff
@@ -26,6 +26,7 @@ buildscript {
 }
 
 plugins {
+    application
```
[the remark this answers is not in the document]
vishesh211: removed

vishesh211:
@mmoczkowski did the required cleanup.

mmoczkowski (approved):
Thank you, a small, but important change

feature/topic/src/main/java/com/google/samples/apps/nowinandroid/feature/topic/navigation/TopicNavigation.kt
eggman87: @mmoczkowski Doesn't using `launchSingleTop` here introduce a bug where you can no longer do nested navigation from interests graph. 

For example, if I tap on a topic within a news card, it does not take me to that topic screen (when the code explicitly calls navigate to topic).

PR #570, opened 2023-02-04
Compile and assemble AndroidTest in `pre-emulator-launch-script`

### LINKED ISSUE #569: [FR]: Improve CI androidTest execution by reordering tasks

### Describe the problem

The current workflow for androidTest uses `reactivecircus/android-emulator-runner` and a single `gradlew` script.

https://github.com/android/nowinandroid/blob/8a25efabfc55941c2ef099a7496139914612c5bc/.github/workflows/Build.yaml#L85-L93

This means (as you can see in the run [logs](https://github.com/android/nowinandroid/actions/runs/4092168776/jobs/7056940654)) the emulator first boots-up before the gradle can compile, assemble and run the android tests.
This obviously takes aways resources that could be better allocated for the compilation/assembling tasks.

### Describe the solution

Do you think it might be helpfull to run these tasks before launching the emulator?
I was thinking about using the `pre-emulator-launch-script` argument to attempt running the androidTest (or introduce a preceeding step).

This first action will end-up as a failure because of `DeviceException: No connected devices!` but it will still have compile and assemble all the tasks.
Then, re-executing it once the emulator is started, we will benefit from the local cache and all the tasks (excluding the testing tasks) will be `UP-TO-DATE`.

### Additional context

_No response_

### Code of Conduct

SimonMarquis:
Closes #569


mmoczkowski:
@borisf Hey Boris, our build master, would you be so kind and check this PR out?

borisf (commented):
The change seems  OK, but we need to test it in the environment

mmoczkowski (approved):
Thank you, let's hope it improves the CI speed.

PR #571, opened 2023-02-04
Upgrade JVM source/target compatibility to 11

SimonMarquis:


dturner (approved):

SimonMarquis:
Yes indeed, it seems like some modules are not configured to use the expected jvm toolchain.
I'll prepare a PR to fix that.

SimonMarquis:
@lihenggui, this should be fixed with #583

PR #575, opened 2023-02-08
Provide true to LocalInspectionMode

takagimeow:
When deploying the preview to the emulator, the image was not displayed.
So I fixed it so that the image set in the placeholder can be displayed in the emulator as well.

|before|after|
|---|---|
|Screenshot_20230208_215957|Screenshot_20230208_220014|


mmoczkowski (approved):
It works, thank you for your contribution!

PR #582, opened 2023-02-10
Improve network pooling and avoid startup image fetches.

yschimke:
Coil and Retrofit can share an OkHttp instance, taking more advantage of connection pooling.

Also one problematic image is not cacheable, causing in my local testing a request on every app startup.

 https://developer.android.com/images/social/android-developers.png

Change-Id: If747fb0833d31ce7c9bedffd400160e979bc945e

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -129,7 +129,6 @@ dependencies {
     implementation(libs.androidx.profileinstaller)
 
     implementation(libs.coil.kt)
-    implementation(libs.coil.kt.svg)
 }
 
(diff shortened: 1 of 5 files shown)


yschimke:
After making the change I noticed https://github.com/android/nowinandroid/blob/803e64be53da2d17906a6d71d83f8c35dd580573/CONTRIBUTING.md

No problem if this is undesired and closed.

calren (approved):
LGTM.

I'm not sure why `Android CI with GMD / android-ci (pixel4api30aospatd)` is running on this PR (and failing). Any ideas? I don't think that's usually part of our test suite.

calren:
thanks for confirming!

PR #583, opened 2023-02-11
Configure jvmToolchain for Android and Kotlin projects

SimonMarquis:
There was only a single Kotlin JVM project, so I did not create a dedicated convention plugin for that.
But this can be done if needed.

References #571


core/model/build.gradle.kts
```diff
@@ -19,6 +19,10 @@ plugins {
     id("kotlin")
 }
 
+kotlin {
+    jvmToolchain(11)
```
lihenggui, line 23: Why was this line of code added in the `core/model` module?

SimonMarquis: Because by default this kotlin module uses version 1.8 as you saw in the logs.

mmoczkowski:
Hey @borisf, would you mind checking this out?

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
```diff
@@ -74,3 +76,12 @@ internal fun Project.configureKotlinAndroid(
 fun CommonExtension<*, *, *, *>.kotlinOptions(block: KotlinJvmOptions.() -> Unit) {
     (this as ExtensionAware).extensions.configure("kotlinOptions", block)
 }
+
+/**
+ * Configure Kotlin's jvm toolchain for Android projects
+ */
+internal fun Project.configureKotlinAndroidToolchain() {
+    extensions.configure<KotlinAndroidProjectExtension> {
+        jvmToolchain(11)
```
borisf, line 85: This is our recommendation to use Java bytecode version 11

borisf (commented):
Looks good to me

mmoczkowski (changes requested):
Simon, could you please address the conflicts? Afterwards we're good to go. Thanks a lot! 😄

mmoczkowski (approved):

PR #585, opened 2023-02-11
ui-test-hilt-manifest uses `nowinandroid.android.hilt` plugin

SimonMarquis:


alexvanyo (approved):
Thanks for the contribution!

PR #596, opened 2023-02-14
Run build tests with the Demo variant

tunjid:


dturner (approved):
This is fine to fix the failing tests caused by the addition of Firebase, but if we ever release a public backend we'll need to change this back so that the `prod` variant is tested. For now, we can rely on kokoro to test our Play builds.

PR #605, opened 2023-02-20
Make Hilt provide Firebase.analytics

dturner:
Also stop `FirebaseAnalyticsHelper` from being a singleton as it doesn't need to be one.

### DIFF
--- a/core/analytics/src/prod/java/com/google/samples/apps/nowinandroid/core/analytics/AnalyticsModule.kt
+++ b/core/analytics/src/prod/java/com/google/samples/apps/nowinandroid/core/analytics/AnalyticsModule.kt
@@ -16,14 +16,25 @@
 
 package com.google.samples.apps.nowinandroid.core.analytics
 
+import com.google.firebase.analytics.FirebaseAnalytics
+import com.google.firebase.analytics.ktx.analytics
+import com.google.firebase.ktx.Firebase
 import dagger.Binds
 import dagger.Module
+import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
 
 @Module
 @InstallIn(SingletonComponent::class)
 abstract class AnalyticsModule {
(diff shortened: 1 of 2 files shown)


core/analytics/src/prod/java/com/google/samples/apps/nowinandroid/core/analytics/FirebaseAnalyticsHelper.kt
```diff
@@ -16,19 +16,17 @@
...
-import javax.inject.Singleton
 
 /**
  * Implementation of `AnalyticsHelper` which logs events to a Firebase backend.
  */
-@Singleton
 class FirebaseAnalyticsHelper @Inject constructor() : AnalyticsHelper {
 
-    private val firebaseAnalytics = Firebase.analytics
+    @Inject lateinit var firebaseAnalytics: FirebaseAnalytics
```
manuelvicnt, line 29: If you `@Inject` a parameter, you're declaring an implicit dependency, that's an anti-pattern. 

Please, pass this dependency in the constructor so that it becomes explicit

```
class FirebaseAnalyticsHelper @Inject constructor(
    private val firebaseAnalytics: FirebaseAnalytics
) : AnalyticsHelper { }
```

core/analytics/src/prod/java/com/google/samples/apps/nowinandroid/core/analytics/AnalyticsModule.kt
```diff
@@ -16,14 +16,25 @@
...
 
 @Module
 @InstallIn(SingletonComponent::class)
 abstract class AnalyticsModule {
     @Binds
     abstract fun bindsAnalyticsHelper(analyticsHelperImpl: FirebaseAnalyticsHelper): AnalyticsHelper
+
+    companion object {
+        @Provides
+        @Singleton
```
manuelvicnt, line 37: Awesome!

manuelvicnt (approved):

PR #607, opened 2023-02-21
Provide a CoroutineScope to fake DataStores

manuelvicnt:
This PR makes DataStore tests more deterministic in tests.

`DataStoreFactory.create` takes a `CoroutineScope` as parameter with a default value of `CoroutineScope(Dispatchers.IO + SupervisorJob())` which isn't synchronized in tests. 

We should be passing a `TestScope` in whose scope we should run the tests


tunjid (approved):
LGTM

zsmb13 (approved):
LGTM

Issue #614, opened 2023-03-02
`onItemClick` parameter is not used and should be removed


dturner:
The `onItemClick` parameter isn't used. It should be removed. https://github.com/android/nowinandroid/blob/ebfbb5bafd1d13119fa4cdf75f0176ad84cb33eb/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCardList.kt#L41

vishesh211:
@dturner I want to contribute to this issue. I request you to assign it to me.

dturner:
All yours :)

chanjungkim:
Will this be closed?

vishesh211:
Yes it is completed and a PR has been already linked to this issue.

gururani-abhishek:
It's been more than a month, @dturner @vishesh211, can I work on this?

gururani-abhishek:
@dturner please have a look :)

PR #615, opened 2023-03-02
`onItemClick` parameter removed in NewsResourceCardList

vishesh211:
Fixes: #614 

Removed on `onItemClick` parameter from NewsResourceCardList


vishesh211:
@dturner Please look into this.

rosejr:
Thanks, but it looks like this was already resolved in #706.

PR #620, opened 2023-03-06
Remove unused datastore-preferences library from versions.toml

daryeou:
This library is currently unused in code, 
so I think it would be better to remove it.


mmoczkowski (approved):

Issue #617, opened 2023-03-03
Many errors in project structure


mobilekosmos:
No build type in module xy matches build type "benchmark"

<img width="1115" alt="image" src="https://user-images.githubusercontent.com/1061543/222617047-d4eb9dcc-5121-4c2f-ac25-f23e1dc2f970.png">

Android Studio Electric Eel | 2022.1.1 Patch 2

priyanshnama:
please assign this to me

SimonMarquis:
This is unfortunately the expected behavior here.
The `app` module is setup to fallback to the `release` build type.
Adding default `benchmark` build types on other modules will introduce unexpected behaviors, as we really want to test the `release` code during benchmarks.

https://github.com/android/nowinandroid/blob/0c8457099d13ebbfed75d4db59cfe43d3d527b8c/app/build.gradle.kts#L59

SimonMarquis:
I think we can close this as "unfixable", wdyt @dturner?

PR #623, opened 2023-03-11
Project structure

priyanshnama:
errors removed from project structure
 - build type "benchmark" added
 - flavor dimension and product added
 
 issue - #617


mmoczkowski:
Thank you Priyansh, but thanks to `matchingFallbacks` it's enough to define the `benchmark` flavor in the `app` module only.

PR #624, opened 2023-03-12
Sort and group Gradle dependencies

SimonMarquis:
Automated with https://github.com/square/gradle-dependencies-sorter


mmoczkowski (approved):
I'm a big fan of sorted dependencies myself. Thank you for doing this!

PR #627, opened 2023-03-15
Remove viewModel as Composable parameters

martinbonnin:
Follow up from [this slack discussion](https://kotlinlang.slack.com/archives/CJLTWPH7S/p1678882128432169). I find passing `ViewModels` as parameters is confusing because it hints at the possibility of passing a different `ViewModel` class for preview/testing. Given that `ViewModels` are final classes, it's pretty hard to change in tests. Or is there another reason to do so?

I'm opening this PR for discussion. 

Also ping @manuelvicnt, I think this practice dates from [this video](https://www.youtube.com/watch?v=0z_dwBGQQWQ) but things might have changed since then? Or is it still a good thing to do?

### DIFF
--- a/feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
+++ b/feature/bookmarks/src/main/java/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
@@ -67,8 +67,8 @@ import com.google.samples.apps.nowinandroid.core.ui.newsFeed
 internal fun BookmarksRoute(
     onTopicClick: (String) -> Unit,
     modifier: Modifier = Modifier,
-    viewModel: BookmarksViewModel = hiltViewModel(),
 ) {
+    val viewModel: BookmarksViewModel = hiltViewModel()
     val feedState by viewModel.feedUiState.collectAsStateWithLifecycle()
     BookmarksScreen(
         feedState = feedState,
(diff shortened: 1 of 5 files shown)


martinbonnin:
Excellent point from @Thomas-Boutin that even though `viewModels` are final classes, they can take arguments so it'd be possible to have something like this:

```kotlin
// Production
val bookmarksViewModel = BookmarksViewModel(ApiRepository())

// Tests
val bookmarksViewModel = BookmarksViewModel(FakeRepository())
```

The later allows to test the composable with fake data in the case of manual dependency injection (DI). I'd argue that in a setup using `Hilt` it feels a bit weird to use manual DI as well.

manuelvicnt:
Hi @martinbonnin ! Thanks for the PR, I agree with Thomas here, this pattern allows you to test and preview the stateful composable function (that one with the ViewModel) with different types dependencies if you need it. 

It's also true the point you make about Hilt, but that would only work in `androidTest`. If you want to `test` this composable function or have previews of it, you wouldn't be able to. This pattern gives you that flexibility. A lot of people use Hilt just for bigger integration tests, and JVM tests for smaller ones.

Some other people don't even have a default value for the `ViewModel` param in the composable function. They prefer to construct it from outside (e.g. in the Navigation destination block) and then pass that in to keep this cleaner.

I think this approach is a good compromise between coupling the stateful composable with a ViewModel implementation but also, allowing for easier JVM tests and previews. On the other hand, the risk is that devs could call this VM with the wrong impl in production code.

As pretty much everything in Architecture, this is about trade-offs. Find the pattern that best adjusts to your team and your project, and follow it throughout :D

martinbonnin:
Thanks for the detailed feedback!

> On the other hand, the risk is that devs could call this VM with the wrong impl in production code. 

Indeed, didn't think of that 👍 

> As pretty much everything in Architecture, this is about trade-offs. Find the pattern that best adjusts to your team and your project, and follow it throughout :D

Right! I guess the question is what do we want to do in this repo. If we're to keep the composable parameter, I'd like to write at least one test/preview that showcases this. From a quick scan of the repo, I haven't found a unit test. I'll try to add one. If anyone has ideas, please chip in!

mmoczkowski:
Hey @martinbonnin, seems like your PR sparked a good discussion. If you wish to add a test using custom viewModel feel free to create a new PR. I'll close this one.

PR #629, opened 2023-03-17
Correct a table entry of a code reference from ArchitectureLearningJo…

vketteni:
…urney.md


dturner (approved):

PR #135, opened 2022-06-10
Add duplicate Gradle wrapper to included build

alexvanyo:
This is an attempt to fix #22 

It looks like certain ways of opening the project in AS result in the included build being built in a different way than when doing so with the command line.

Notably the included build doesn't have a Gradle wrapper defined in it, so it appears that Android Studio ends up using the embedded Gradle version. The embedded version of Gradle depends on the version of Android Studio, and this version does not necessarily equate to the version of Gradle that the project is using.

As a result, features that might normally work or be included with the version of Gradle the project is using may not be available with the version.

For instance, if opening `build-logic` in Android Studio Chipmunk directly, version 7.2 of Gradle is used, which causes a build error about enabling version catalogs, even though those are now stable on version 7.4+, which the main project is using.

As a workaround, this PR adds the wrapper and `gradle-wrapper.properties` to `build-logic`, to let Android Studio know which version of Gradle to use.

This is a bit of duplication, but Renovate should be able to update both of these versions of Gradle automatically, and avoids indirection that a solution like symlinking might cause.

After this change, opening `build-logic` directly syncs without errors.

In normal day-to-day use, there is still just one `./gradlew` wrapper at the root which should be used from the command line for all Gradle commands.

Edit: Found bug for this case: https://issuetracker.google.com/issues/181336796

PR #630, opened 2023-03-18
Remove unnecessary copy of Gradle wrapper

drawers:
We are finding this project extremely useful, especially the guidance around included builds.

One thing we noticed when we wrote our own convention plugins based on the implementation in nowinandroid is that there are currently two Gradle wrappers in this project: one in the project root and one in the `build-logic` root. 

I do not think the second copy of the Gradle wrapper is necessary:

1. The Gradle samples for included builds do not have this (https://docs.gradle.org/6.7-rc-5/samples/sample_composite_builds_hierarchical_multirepo.html)
2. If the intention is that the included build can be run as a separate project, this is already possible through `./gradlew --settings-file=build-logic/settings.gradle.kts` 

Testing:

Ran the following commands successfully
`./gradlew :build-logic:convention:clean --rerun-tasks --no-configuration-cache`
`/gradlew :build-logic:convention:jar --rerun-tasks --no-configuration-cache :build-logic:convention:jar`


drawers:
> Thanks for the PR! The extra Gradle wrapper was added in #135 as a workaround for Android Studio sometimes getting into a bad state due to not building with the correct Gradle version (https://issuetracker.google.com/issues/181336796)
> 
> That issue was marked as fixed, so it's possible we can remove the workaround now if that issue no longer occurs.

Ah nice one @alexvanyo ! Did not know about this bug 😖 

I can't see in which version of Android Studio the bug was fixed from that issue tracker, but FWIW I have been running composite builds in an Android project (following this nowinandroid example) even in Dolphin with no problems so far.

PR #643, opened 2023-03-31
Use ReportDrawnWhen

mlykotom:
Use `ReportDrawnWhen` to signal when the UI is ready to be used by user.
- It allows measuring Time to fully drawn metric from Macrobenchmarks and Android Vitals.
- It improves startup performance by marking all executed code before the signal as startup code for dexopt.

### DIFF
--- a/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
+++ b/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
@@ -16,7 +16,7 @@
 
 package com.google.samples.apps.nowinandroid.feature.foryou
 
-import android.app.Activity
+import androidx.activity.compose.ReportDrawnWhen
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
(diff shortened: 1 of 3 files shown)


feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -126,23 +123,8 @@ internal fun ForYouScreen(
...
-        // and is safe to call multiple times though.
-        LaunchedEffect(Unit) {
-            // We're leveraging the fact, that the current view is directly set as content of Activity.
-            val activity = localView.context as? Activity ?: return@LaunchedEffect
-            // To be sure not to call in the middle of a frame draw.
-            localView.doOnPreDraw { activity.reportFullyDrawn() }
-        }
-    }
+    // This code should be called when the UI is ready for use and relates to Time To Full Display.
+    ReportDrawnWhen { !isSyncing && !isOnboardingLoading && !isFeedLoading }
```
alexvanyo, line 127: Is there any downside for this never being called, or being called much later in the case where we aren't on the for you screen?

For instance, if we navigate to another tab, and then rotate the device, we won't call `ReportDrawnWhen` until we navigate back to the for you screen later.

mlykotom: If this is never called, system assumes fully drawn ~5seconds after Activity.onCreate.

There isn't any downside at this point.

If there were multiple entrypoints into the app (e.g deeplinks), we should implement that there as well.

alexvanyo: Great info to know, thanks!

alexvanyo (approved):

mmoczkowski (approved):

Issue #633, opened 2023-03-22
[Bug]: Splashscreen theme doesn't change after switching the color theme


illarionov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

Changing the theme from the application settings does not update the theme used on the splash screen.

###### Steps to reproduce
* When using a light system theme
* Tap on the settings icon and set the application theme to dark
* Close the application, start it again (with the cold start) and note the splash screen

Expected result:
Splash screen should be in dark mode

Actual result:
Light splash screen switches to a dark screen of the application.

I think that UiModeManager#setApplicationNightMode should also be called on saving the theme settings as it is described in  ["Implement dark theme"](https://developer.android.com/develop/ui/views/theming/darktheme#change-themes)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

alexvanyo:
That's a great idea, we should definitely do that.

Code-indomitus:
Hi may I work on this?

alexvanyo:
There was a PR open for fixing this (#646), but I just closed it due to inactivity. This is still a relevant issue to be fixed, so if the creator of that PR or someone else would like to fix this please open a new PR.

Dev-Aditya-More:
Hey! is this still a open issue?

stevenelliottjr:
Hi there! I've submitted a fix for this in PR #2060. 

The approach syncs the user's `DarkThemeConfig` preference to `UiModeManager.setApplicationNightMode()` so the system-level night mode stays in sync with the app's setting. This ensures the splash screen picks up the correct theme on the next cold start.

Key details:
- Guarded behind `Build.VERSION.SDK_INT >= Build.VERSION_CODES.S` since `setApplicationNightMode` was added in API 31
- Uses `distinctUntilChanged()` to avoid redundant system calls
- Lifecycle-aware collection so it only runs while the activity is started

Would love any feedback — happy to make adjustments if needed!

Phoenix1808:
Is this PR still open for contribution or closed?

agrawalha:
Hi, I'd like to work on this issue. Could it be assigned to me?

PR #646, opened 2023-04-04
Added uiModeManager 

anandwana001:
Issue - https://github.com/android/nowinandroid/issues/633
Added uiModeManager for SplashScreen to behave as per user selected theme

[device-2023-04-04-130106.webm](https://user-images.githubusercontent.com/25057618/229720136-f9eb2408-1ab7-47a1-b509-f2f134a66988.webm)

Looking forward for review for learning.

### DIFF
--- a/app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
+++ b/app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
@@ -16,10 +16,14 @@
 
 package com.google.samples.apps.nowinandroid
 
+import android.app.UiModeManager
+import android.content.Context
+import android.os.Build
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
+import androidx.appcompat.app.AppCompatDelegate
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSizeClassApi
 import androidx.compose.material3.windowsizeclass.calculateWindowSizeClass
@@ -95,8 +99,27 @@ class MainActivity : ComponentActivity() {
         WindowCompat.setDecorFitsSystemWindows(window, false)
 
         setContent {
-            val systemUiController = rememberSystemUiController()
+            val uiModeManager = getSystemService(Context.UI_MODE_SERVICE) as UiModeManager
             val darkTheme = shouldUseDarkTheme(uiState)
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+                uiModeManager.setApplicationNightMode(
+                    when (darkTheme) {
+                        true -> UiModeManager.MODE_NIGHT_YES
+                        else -> UiModeManager.MODE_NIGHT_NO
+                    }
+                )
+            }
+            else {
+                AppCompatDelegate.setDefaultNightMode(
+                    when (darkTheme) {
+                        true -> AppCompatDelegate.MODE_NIGHT_YES
+                        else -> AppCompatDelegate.MODE_NIGHT_NO
+                    }
+                )
+            }
+
+            val systemUiController = rememberSystemUiController()
 
             // Update the dark content of the system bars to match the theme
             DisposableEffect(systemUiController, darkTheme) {
--- a/app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
+++ b/app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
@@ -18,12 +18,10 @@ package com.google.samples.apps.nowinandroid
 
 import android.app.UiModeManager
 import android.content.Context
-import android.os.Build
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
-import androidx.appcompat.app.AppCompatDelegate
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSizeClassApi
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -95,9 +99,28 @@ class MainActivity : ComponentActivity() {
         WindowCompat.setDecorFitsSystemWindows(window, false)
 
         setContent {
-            val systemUiController = rememberSystemUiController()
+            val uiModeManager = getSystemService(Context.UI_MODE_SERVICE) as UiModeManager
             val darkTheme = shouldUseDarkTheme(uiState)
 
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+                uiModeManager.setApplicationNightMode(
```
alexvanyo, line 106: I think we should pass through the `UiModeManager.MODE_NIGHT_AUTO` if the user preference is `DarkThemeConfig.FOLLOW_SYSTEM`.

app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -95,9 +99,28 @@ class MainActivity : ComponentActivity() {
         WindowCompat.setDecorFitsSystemWindows(window, false)
 
         setContent {
-            val systemUiController = rememberSystemUiController()
+            val uiModeManager = getSystemService(Context.UI_MODE_SERVICE) as UiModeManager
             val darkTheme = shouldUseDarkTheme(uiState)
 
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
```
alexvanyo, line 105: Since this is a side-effect of composition, we should wrap this into some sort of `SideEffect` or `DisposableEffect` block.

alexvanyo: @tunjid do you think there is a better place to put this logic? It doesn't have to be done as a side-effect of composition, I think it could be some sort of global listener that is observing the user preferences.

alexvanyo (changes requested):

app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
tunjid: I think so too. Seems `NiaAppState` is a better candidate for this. It can be passed a function to invoke when it detects a theme change has occured.

anandwana001: Moved to `NiaAppState`

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
```diff
@@ -164,3 +168,23 @@ private fun NavigationTrackingSideEffect(navController: NavHostController) {
         }
     }
 }
+
+internal fun checkAppThemeOnLaunch(uiModeManager: UiModeManager, darkTheme: Boolean) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+        uiModeManager.setApplicationNightMode(
+            when (darkTheme) {
```
alexvanyo, line 175: I think we should be passing `UiModeManager.MODE_NIGHT_AUTO` here if the user's selection is to follow system.

illarionov:
What do you think about handling this in the data layer? We can think of the UiModeManager as an additional data source for the OfflineFirstUserDataRepository. 
Also it would be interesting to see how to handle cases with saving theme preferences when the user backs up the app or when the user upgrades their device's firmware from an old version to a new one.

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
manuelvicnt: +1

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
```diff
@@ -164,3 +168,23 @@ private fun NavigationTrackingSideEffect(navController: NavHostController) {
         }
     }
 }
+
+internal fun checkAppThemeOnLaunch(uiModeManager: UiModeManager, darkTheme: Boolean) {
```
manuelvicnt, line 172: @tunjid @alexvanyo IMO, this function shouldn't be here because it's not part of the `NiaAppState` API surface. Also, `NiaAppState` responsibility is to handle the UI logic of the `NiaApp` composable function, not the `MainActivity`. 

I'd probably have this as a private function in the Activity itself, and would call it `setAppTheme` instead of `checkAppThemeOnLaunch`. It's not checking anything, it's setting the theme, and also, if this is called "on launch" or not is something this function shouldn't care about

app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -95,12 +98,16 @@ class MainActivity : ComponentActivity() {
         WindowCompat.setDecorFitsSystemWindows(window, false)
 
         setContent {
-            val systemUiController = rememberSystemUiController()
             val darkTheme = shouldUseDarkTheme(uiState)
```
dturner, line 101: Not your code but both the variable `darkTheme` and the method `shouldUseDarkTheme` are poorly named and make reading this PR difficult. 

The variable which contains the user's dark theme preference should be: `shouldUseDarkTheme` or `darkThemePreference`, and the method which obtains the value should be `getDarkThemePreference`. 

If you could address in this PR that'd be great.

manuelvicnt:
Could that be in a separate file in the `ui-common` module? I'm not sure how this behavior is related to the data layer

alexvanyo:
Closing due to inactivity

PR #648, opened 2023-04-04
Added Extra height behind when Snackbar visible

anandwana001:
| With Internet | Without Internet |
| --- | ---- |
|  |   |
|  |   |
|  |   |


anandwana001:
@dturner Could you PTAL?

thagikura:
I think the issue is not about the place of the Snackbar. 

The fix is to show the Snackbar only for a couple of seconds and/or allowing the user to dismiss the Snackbar.

anandwana001:
@alexvanyo Could you help us here with the requirements?

dturner:
A SnackBar is intended to be shown on top of other content, not alongside it.

PR #650, opened 2023-04-06
Initialize sync manually

tunjid:
This PR removes the use of Androidx startup for sync.

Since NiA does not expose content providers for other apps to consume, using androidx startup (using  ContentProviders by proxy) to bootstrap the sync process is unnecessary.

Sync is now explicitly started in the `onCreate` of the `Application`, rather than automatically initialized by Androidx startup.

### DIFF
--- a/sync/work/build.gradle.kts
+++ b/sync/work/build.gradle.kts
@@ -33,7 +33,6 @@ dependencies {
     implementation(project(":core:datastore"))
     implementation(project(":core:model"))
     implementation(libs.androidx.lifecycle.livedata.ktx)
-    implementation(libs.androidx.startup)
     implementation(libs.androidx.tracing.ktx)
     implementation(libs.androidx.work.ktx)
     implementation(libs.firebase.cloud.messaging)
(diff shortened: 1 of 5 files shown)


sync/work/build.gradle.kts
```diff
@@ -33,7 +33,6 @@ dependencies {
     implementation(project(":core:datastore"))
     implementation(project(":core:model"))
     implementation(libs.androidx.lifecycle.livedata.ktx)
-    implementation(libs.androidx.startup)
```
alexvanyo, line 36: This was our only usage of `androidx.startup`, so it might be worth removing entirely from `libs.versions.toml`?

alexvanyo (approved):
LGTM, this will also avoid the testing issues we were facing as well, right? Since we aren't relying on the interplay between `Hilt` and `ContentProviders` via `androidx.startup`.

dturner:
Thanks for doing this @tunjid and reviewing @alexvanyo 👍

Issue #529, opened 2023-01-12
[FR]: Upgrade sourceCompatibility and targetCompatibility to Java 17


angiesasmita:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

Currently sourceCompatibility and targetCompatibility targets Java 8. This might have resulted in cryptic error message for bug: https://github.com/android/nowinandroid/issues/527 and build and runtime performance penalty

### Describe the solution

- Upgrade sourceCompatibility and targetCompatibility to target java 17
- Check if there's build time improvement
- Check if there's any regression issue

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

AdnanHabibMirza:
Hi @dturner, I would like to contribute to it.

chanjungkim:
Isn't that easy issue??

lihenggui:
> 

No, it isn't that easy. It will throw multiple errors.

```
No matching variant of project :build-logic:convention was found. The consumer was configured to find a runtime of a library compatible with Java 11, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.gradle.plugin.api-version' with value '7.6' but:
Variant 'apiElements' capability com.google.samples.apps.nowinandroid.buildlogic:convention:unspecified declares a library, packaged as a jar, and its dependencies declared externally:
                 - Incompatible because this component declares an API of a component compatible with Java 17 and the consumer needed a runtime of a component compatible with Java 11
                 - Other compatible attribute:
                     - Doesn't say anything about org.gradle.plugin.api-version (required '7.6')
             - Variant 'mainSourceElements' capability com.google.samples.apps.nowinandroid.buildlogic:convention:unspecified declares a component, and its dependencies declared externally:
               ...
```

If you upgrade Gradle you will get other issues.

```
A problem occurred configuring project ':core:datastore'.
> Could not get unknown property 'source' for generate-proto-generateDemoDebugProto of type org.gradle.api.internal.file.DefaultSourceDirectorySet.
```

calren:
I believe this branch is related https://github.com/android/nowinandroid/commits/av/agp-8.1

PR #662, opened 2023-04-14
Update to AGP 8.0

### LINKED ISSUE #660: [Bug]: Update to AGP 8.0

### What happened?

AGP 8.0 is stable with Android Studio Flamingo, so we should update our dependencies to it.

### Relevant logcat output

_No response_

### Code of Conduct

alexvanyo:
Fixes #660 and #529


build-logic/convention/build.gradle.kts
```diff
@@ -25,6 +27,12 @@ java {
     targetCompatibility = JavaVersion.VERSION_11
```
mlykotom, line 27: you can raise the JavaVersion to 17. Here and in other places as well.

kokoro/build.sh
```diff
@@ -36,7 +36,7 @@ echo y | ${ANDROID_HOME}/tools/bin/sdkmanager --licenses
 cd $KOKORO_ARTIFACTS_DIR/git/nowinandroid
 
 # The build needs Java 11, set it as the default Java version.
```
mlykotom, line 38:
```suggestion
# The build needs Java 17, set it as the default Java version.
```

mlykotom (approved):
You can raise the Java version to 17, otherwise LGTM

kokoro/build.sh
alexvanyo: Good catch, updated.

build-logic/convention/build.gradle.kts
alexvanyo: Sounds good, raised to 17 everywhere.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
```diff
@@ -40,27 +40,28 @@ internal fun Project.configureKotlinAndroid(
...
+            sourceCompatibility = JavaVersion.VERSION_17
+            targetCompatibility = JavaVersion.VERSION_17
             isCoreLibraryDesugaringEnabled = true
         }
+    }
 
+    // Use withType to workaround https://youtrack.jetbrains.com/issue/KT-55947
+    tasks.withType<KotlinCompile>().configureEach {
         kotlinOptions {
+            // Set JVM target to 11
```
LinX64, line 52: Shouldn't we change this to 17, and not 11?
You're using `VERSION_17` in the next line.

alexvanyo: Ah yep, good catch

app/proguard-rules.pro
```diff
@@ -24,3 +24,13 @@
 
 # @Serializable and @Polymorphic are used at runtime for polymorphic serialization.
 -keepattributes RuntimeVisibleAnnotations,AnnotationDefault
+
+-dontwarn org.bouncycastle.jsse.BCSSLParameters
```
itsandreramon, line 28: just curious why these rules are needed

build-logic/convention/build.gradle.kts
```diff
@@ -14,15 +14,23 @@
...
     `kotlin-dsl`
 }
 
 group = "com.google.samples.apps.nowinandroid.buildlogic"
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
```
carlonzo, line 27: I understand that AGP now requires JDK 17 to run, but why would you bump the source/targetCompatibility, and jvmTarget to 17? is this required as well?

borisf: Basically to have the latest versions, toolchains and formats. We talk about importance of updating early and often - https://www.droidcon.com/2022/08/03/5-ways-to-improve-your-android-build-productivity/

carlonzo: Thanks about your answer. I dont doubt the fact is important to update early and often. Using the latest java toolchain makes sense, but I cant find resource to understand the benefits on bumping the source compatibility and kotlin jvm target here. I assume that the generated bytecode will also need to be desugared before to be dexed? 
are there any references to documentation about this part?

itsandreramon: technically you still get compiler improvements depending on what your actually using. you can still use JDK19 and set targetCompatibility to 17

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -40,16 +44,18 @@ internal fun Project.configureAndroidCompose(
...
-        }
-
         dependencies {
             val bom = libs.findLibrary("androidx-compose-bom").get()
             add("implementation", platform(bom))
             add("androidTestImplementation", platform(bom))
         }
     }
+
+    tasks.withType<KotlinCompile>().configureEach {
```
saihemanth121, line 54: Is setting freeCompilerArgs on tasks different from older way of setting it through kotlinOptions extension?

dturner (approved):

build-logic/convention/build.gradle.kts
dturner: In the interest of getting this change merged I will resolve this conversation. 

@lub0s thanks for your suggestion makes a lot of sense to me. I'll pass this on to the tools team for prioritisation.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
dturner: In the interest of getting this merged I will resolve this conversation, but have passed this on to the tools team. Hopefully they can respond.

SimonMarquis: nit: you don't even need to call `.configureEach`.
The following code does the same thing 🙂 

```kotlin
    tasks.withType<KotlinCompile> {
        /* */
    }
```

PR #666, opened 2023-04-16
Update Kotlin 1.8.20 stack 

SimonMarquis:
- KSP 1.8.20-1.0.11
- Compose compiler 1.4.5

This needs JDK 17, and therefore applied on top of #662.

~Also, we might want to wait for the next KSP version with this fix https://github.com/google/ksp/issues/1280
Otherwise we get errors [like this](https://github.com/android/nowinandroid/actions/runs/4715236703/jobs/8362091931?pr=666#step:9:1721) if KSP uses the build-cache.~ fixed in 1.8.20-1.0.11


app/proguard-rules.pro
```diff
@@ -24,3 +24,13 @@
 
 # @Serializable and @Polymorphic are used at runtime for polymorphic serialization.
 -keepattributes RuntimeVisibleAnnotations,AnnotationDefault
+
+-dontwarn org.bouncycastle.jsse.BCSSLParameters
```
itsandreramon, line 28: just curious why these rules are needed

dturner (approved):
Thanks for doing this. LGTM.

SimonMarquis:
I rebase my changes to make it cleaner now that AGP 8 is merged,

alexvanyo (approved):

PR #667, opened 2023-04-16
Cleanup warnings and replace deprecated methods

### LINKED ISSUE #654: [Cleanup] Remove suppression for plugin applications with alias once updated to Gradle 8.1

### What happened?

#35 added suppression for using the `alias` to apply plugins from `libs.versions.toml`.

In Gradle 8.1, this workaround is no longer necessary: https://docs.gradle.org/8.1-rc-1/userguide/upgrading_version_8.html#kotlin_dsl_plugins_catalogs_workaround

Once we update to Gradle 8.1, we'll be able to remove those suppressions.

### Relevant logcat output

_No response_

### Code of Conduct

SimonMarquis:
Some changes rely on #662

Closes #654

### DIFF
--- a/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt
+++ b/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt
@@ -27,6 +27,7 @@ import org.gradle.kotlin.dsl.withType
 import org.gradle.testing.jacoco.plugins.JacocoPluginExtension
 import org.gradle.testing.jacoco.plugins.JacocoTaskExtension
 import org.gradle.testing.jacoco.tasks.JacocoReport
+import java.util.Locale
 
 private val coverageExclusions = listOf(
     // Android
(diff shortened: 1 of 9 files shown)


SimonMarquis:
Duplicates part of #686

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt
```diff
@@ -36,6 +37,10 @@ private val coverageExclusions = listOf(
     "**/Manifest*.*"
 )
 
+private fun String.capitalize() = replaceFirstChar {
```
rosejr, line 40: Is this needed? It doesn't seem to be used in this change.

SimonMarquis: This fixes the following warning message:

```
w: file:///[...]/nowinandroid/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt:51:48 'capitalize(): String' is deprecated. Use replaceFirstChar instead.
```

It appear unused but in fact, it is because the deprecated `String.capitalize()` extension is already defined in `kotlin.text` and does not need import statement.

rosejr: I see, makes sense.

rosejr (approved):

dturner (approved):
Thanks @SimonMarquis - simpler and clearer 👍

Also, thanks @rosejr for reviewing.

PR #678, opened 2023-04-22
Don't rely on `finalizeDsl` API to configure Crashlytics

SimonMarquis:


thagikura:
Why it shouldn't rely on finalizeDsl?

build-logic/convention/src/main/kotlin/AndroidApplicationFirebaseConventionPlugin.kt
```diff
@@ -41,15 +41,13 @@ class AndroidApplicationFirebaseConventionPlugin : Plugin<Project> {
                 "implementation"(libs.findLibrary("firebase.crashlytics").get())
             }
 
-            extensions.configure<ApplicationAndroidComponentsExtension> {
-                finalizeDsl {
```
mmoczkowski, line 45: Hey @dturner, any idea why `finalizeDsl` was used here?

SimonMarquis:
We don't need it because it's actually a "hack" callback provided by AGP, and regular Gradle APIs can be used instead.
The finalizedBy would have been necessary, otherwise buildTypes would not be created at that time.

build-logic/convention/src/main/kotlin/AndroidApplicationFirebaseConventionPlugin.kt
dturner: It was the result of much hacking around with Firebase config. Happy to go with the simpler solution suggested by @SimonMarquis

dturner (approved):

dturner:
Thanks for doing this @SimonMarquis

PR #679, opened 2023-04-22
Simplify String resource loading in `NavigationTest.kt`

SimonMarquis:


app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
```diff
@@ -69,35 +71,21 @@ class NavigationTest {
...
-            sampleTopic = "Headlines"
-            appName = getString(R.string.app_name)
-            saved = getString(BookmarksR.string.saved)
-            settings = getString(SettingsR.string.top_app_bar_action_icon_description)
-            brand = getString(SettingsR.string.brand_android)
-            ok = getString(SettingsR.string.dismiss_dialog_button_text)
-        }
-    }
+    // The strings used for matching in these tests
+    private val done by composeTestRule.stringResource(FeatureForyouR.string.done)
```
mmoczkowski, line 78: Not used, can be removed

app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
```diff
@@ -69,35 +71,21 @@ class NavigationTest {
...
+    private val navigateUp by composeTestRule.stringResource(FeatureForyouR.string.navigate_up)
+    private val forYouLoading by composeTestRule.stringResource(FeatureForyouR.string.for_you_loading)
```
mmoczkowski, line 80: Not used, can be removed

mmoczkowski (approved):
Thank you for your contribution. I think your solution is better than lateinit properties. Would you be so kind and remove unused variables while you're at it? 🙏

PR #684, opened 2023-04-25
Replace `if` condition with AnimatedVisibility component

### LINKED ISSUE #683: [FR]: Replace `if` condition with AnimatedVisibility component in SettingsDialog

### Describe the problem

When changing the Theme Brand in the Settings Dialog of the app, the dynamic color preference is not showing smoothly. Currently, an `if` condition is being used to control the visibility of the preference, which is causing the abrupt behavior.

### Describe the solution

Replace the `if` condition with the `AnimatedVisibility` component. This will allow for a smoother and more visually appealing transition when changing the Theme Brand.

### Additional context

_No response_

### Code of Conduct

MaximillianLeonov:
Fixes #683.


feature/settings/src/main/java/com/google/samples/apps/nowinandroid/feature/settings/SettingsDialog.kt
```diff
@@ -149,7 +151,7 @@ fun SettingsDialog(
 }
 
 @Composable
-private fun SettingsPanel(
+private fun ColumnScope.SettingsPanel(
```
thagikura, line 154: Why it's in the ColumnScope now?

MaximillianLeonov: A composable function should emit either 0 or 1 pieces of layout, but no more. The `SettingsPanel` composable uses more than 1 layout (`SettingsDialogSectionTitle`, `Column`, `AnimatedVisibility`), and the `SettingsPanel` composable is intended to be called from `ColumnScope`.

In addition, the `ColumnScope.AnimatedVisibility` extension composable is used with the predefined animations for the `Column` layout.

thagikura: I think ColumnScope is only used for the direct children of Column when that Composable wants to use any of the Modifiers specific to [ColumnScope](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/ColumnScope)

In this case, I don't see a reason of using ColumnScope since it doesn't use Modifiers that require ColumnScope.

alexvanyo: The overload of [`ColumnScope.AnimatedVisibility`](https://developer.android.com/reference/kotlin/androidx/compose/animation/package-summary#(androidx.compose.foundation.layout.ColumnScope).AnimatedVisibility(kotlin.Boolean,androidx.compose.ui.Modifier,androidx.compose.animation.EnterTransition,androidx.compose.animation.ExitTransition,kotlin.String,kotlin.Function1)) called here will be tailored to being in a `Column`, which has different animations than the default [`AnimatedVisibility`](https://developer.android.com/reference/kotlin/androidx/compose/animation/package-summary#AnimatedVisibility(kotlin.Boolean,androidx.compose.ui.Modifier,androidx.compose.animation.EnterTransition,androidx.compose.animation.ExitTransition,kotlin.String,kotlin.Function1)) that isn't in a `RowScope` or a `ColumnScope`.

thagikura: Oh sorry. I wasn't aware that there is a AnimatedVisibility specific to ColumnScope. Thanks Alex.

thagikura: How about adding a comment that ColumnScope is used to use the overload of `ColumnScope.AnimatedVisibility` for future readers?

thagikura (approved):

alexvanyo (approved):

PR #689, opened 2023-04-26
Set the source/target compatibility to 11

thagikura:
This is because up to Java 11 features are supported without requiring minimum API level nor AGP versions at the moment as in https://developer.android.com/studio/write/java11-minimal-support-table.

Note that they are separate things that JDK 17 is used to build the app.

Follow up on #662

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -23,13 +23,15 @@ plugins {
 group = "com.google.samples.apps.nowinandroid.buildlogic"
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_17
-    targetCompatibility = JavaVersion.VERSION_17
+    // Up to Java 11 APIs are available through desugaring
+    // https://developer.android.com/studio/write/java11-minimal-support-table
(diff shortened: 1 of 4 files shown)


dturner (approved):
Thanks for spotting this. Approving because you're the expert although I'm not entirely clear on the reason we shouldn't use Java 17 here.

dturner:
From @thagikura: "The Android build tool chain doesn't know how to handle JDK 17 features at the moment. As in https://developer.android.com/studio/write/java11-minimal-support-table, 11 is the reasonable choice for setting for compileOptions.sourceCompatibility/targetCompatibility."

G00fY2:
@thagikura I see your point, but the [minimal-support-table](https://developer.android.com/studio/write/java11-minimal-support-table) is not really an argument for holding back JVM 17 target. R8 can handle JVM 17 classes, so the Dex code will run on all Android devices. And using Java APIs higher than Java 6/7 will always require developer attention, since core library desugaring is just a subset.

> I think it's better every developer can copy/paste the relevant code pieces into their app.

That is something I can agree to. Just didn't like the initial reasoning behind this PR.

PR #694, opened 2023-04-26
Add missing `gradle/gradle-build-action` setup in `AndroidCIWithGmd`

SimonMarquis:
This will prevent downloading and unzipping Gradle distribution but fetch it from the cache instead like other workflows.


.github/workflows/AndroidCIWithGmd.yaml
```diff
@@ -15,11 +15,12 @@ jobs:
         device-config: [ "pixel4api30aospatd", "pixelcapi30aospatd" ]
 
     steps:
+      - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
           java-version: 17
-      - uses: actions/checkout@v3
+      - uses: gradle/gradle-build-action@v2
```
LinX64, line 23: I'd add a name at the top of the actions we are adding here, the rest of the actions already have names, so let's keep the consistency here as well.

```suggestion
        - name: Setup Gradle
        uses: gradle/gradle-build-action@v2
```

What do you think?

SimonMarquis: Personally I would remove them all, unless they are adding valuable information.

LinX64: Sure, you are free to do whatever you want on your own project, but last time I checked this was supposed to be a project for beginners, so any helpful information can help :]

SimonMarquis: I'll let the maintainers decide what is best here. Not all jobs currently have comments, so if we want to enforce it I'll add them on other entries as well. I'm not against adding comments on self explanatory statements, I only have a strong personal opinion against them 😜

dturner: In code, I'm very much against adding comments to self explanatory statements as it adds noise. For CI, however, there is added benefit in adding a `name` to each step as it makes it clear exactly what that step is doing both in the config and in the job output. E.g. 

image

Suggest we add descriptive names to each step. Can either be done in this PR, or in a separate PR.

dturner (approved):

.github/workflows/AndroidCIWithGmd.yaml
SimonMarquis: Then I guess it would be better to make this a distinct commit, and set names uniformly. I'll do that on top of this PR if that's ok for you :)

SimonMarquis: FYI, the names have been added/restored in #710 with missing steps.

abalros:
gradle/gradle-build-actionالإعداد المفقود بتنسيقAndroidCIWithGmd # 694

SimonMarquis:
@dturner, sorry for the ping 🙏 (could we add [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) in the future?)

PR #696, opened 2023-04-26
Disable unnecessary Android instrumented tests

SimonMarquis:
A lot of resources are wasted compiling, packaging, installing and running Android Tests on modules that do not contain any instrumented tests (12 out of 22).

For example, in [this build](https://github.com/android/nowinandroid/actions/runs/4812101266/jobs/8569560329?pr=666#step:6:1287) you can see the following messages:

> Starting 0 tests on AVD

**Note**: this could be improved by checking other potential sourceSets based on `buildTypes` and `flavors`, but none of the modules do at the moment.


SimonMarquis:
There might already be public APIs to get all the sourceSets of a variant, but I did not found it.

dturner (approved):

PR #698, opened 2023-04-28
Fix issue #611 : shouldShowSettingsDialog state save in activity recreation

### LINKED ISSUE #611: [Bug]: shouldShowSettingsDialog is not persisted through activity recreation

### What happened?

When the settings dialog is open, causing activity recreation through resizing or rotating the app causes the settings dialog to close.

We should likely keep this state saved via saved instance state so that the settings dialog remains open.

### Relevant logcat output

_No response_

### Code of Conduct

KwakEuiJin:
Fixes #611

**Issue name**: [Bug]: shouldShowSettingsDialog is not persisted through activity recreation

- use rememberSaveable

> While remember helps you retain state across recompositions, the state is not retained across configuration changes. For this, you must use rememberSaveable. rememberSaveable automatically saves any value that can be saved in a Bundle. For other values, you can pass in a custom saver object.

reference: https://developer.android.com/jetpack/compose/state

https://user-images.githubusercontent.com/93872496/235056844-dba80f0a-d9db-4cc9-95aa-bd8b77585cf1.mp4

### DIFF
--- a/app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NiaAppStateTest.kt
+++ b/app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NiaAppStateTest.kt
@@ -20,7 +20,10 @@ import androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSiz
 import androidx.compose.material3.windowsizeclass.WindowSizeClass
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.MutableState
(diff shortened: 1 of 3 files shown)


KwakEuiJin:
@alexvanyo 
Please review my code, 
If it is not easy to test, I will change the code. 
thanks 🙃

KwakEuiJin:
code review please !!

alexvanyo:
@KwakEuiJin thank you for proposing this PR! Apologies that we didn't review it directly, we evaluated your solution proposed here and the one in #699 as potential solutions along with a third one, and we ended up deciding on the third option merged in #713.

PR #710, opened 2023-05-04
Add missing steps in `AndroidCIWithGmd.yaml`

SimonMarquis:
- `gradle/wrapper-validation-action`
- copy CI's `gradle.properties` file

And restore explicit steps names.

This was applied on top of this PR:
- #694


SimonMarquis:
@dturner, sorry for the ping 🙏 (also, could we add [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) in the future?)

dturner (approved):

dturner:
@SimonMarquis No problem at all, thanks for all your contributions. 

Regarding CODEOWNERS, we have a weekly rotation for triaging PRs so, in theory, reviewers should be assigned then. However, I'm fully aware that sometimes PRs stay open for weeks/months without progressing. I don't think CODEOWNERS is going to solve that because it's due to our internal bandwidth to perform reviews. 

Feel free to do what you did here and ping me directly if there's something particularly important which you feel should be reviewed/merged.

SimonMarquis:
> Regarding CODEOWNERS, we have a weekly rotation for triaging PRs so, in theory, reviewers should be assigned then. However, I'm fully aware that sometimes PRs stay open for weeks/months without progressing. I don't think CODEOWNERS is going to solve that because it's due to our internal bandwidth to perform reviews.

That make sense, thanks for sharing your review process.

> Feel free to do what you did here and ping me directly if there's something particularly important which you feel should be reviewed/merged.

Will do! I was a bit relunctant to ping you at first, but now that I understand how the review process is handled on your side, I'll not hesitate to ping if it needs immediate attention. :)

PR #712, opened 2023-05-05
Notify users when news are updated

tunjid:
This PR:

* Configures the `AndroidNotifier` to show users notifications in the system notification shade
* Requests permissions to show the notifications on the ForYouScreen

Screenshot_20230505_114736

### DIFF
--- a/core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
+++ b/core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
@@ -16,17 +16,126 @@
 
 package com.google.samples.apps.nowinandroid.core.notifications
 
+import android.Manifest.permission
+import android.app.Notification
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.content.Context
+import android.content.pm.PackageManager
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
+import androidx.core.app.ActivityCompat
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationCompat.InboxStyle
+import androidx.core.app.NotificationManagerCompat
 import com.google.samples.apps.nowinandroid.core.model.data.NewsResource
+import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 import javax.inject.Singleton
+
+private const val NEWS_NOTIFICATION_SUMMARY_ID = 1
+private const val NEWS_NOTIFICATION_CHANNEL_ID = ""
+private const val NEWS_NOTIFICATION_GROUP = "NEWS_NOTIFICATIONS"
 
 /**
  * Implementation of [Notifier] that displays notifications in the system tray.
  */
 @Singleton
-class AndroidSystemNotifier @Inject constructor() : Notifier {
+class AndroidSystemNotifier @Inject constructor(
+    @ApplicationContext private val context: Context,
+) : Notifier {
 
-    override fun onNewsAdded(newsResources: List<NewsResource>) {
-        // TODO, create notification and display to the user
+    override fun onNewsAdded(
+        newsResources: List<NewsResource>,
+    ) = with(context) {
+        if (ActivityCompat.checkSelfPermission(
+                this,
+                permission.POST_NOTIFICATIONS,
+            ) != PackageManager.PERMISSION_GRANTED
+        ) {
+            return
+        }
+
+        val newsNotifications = newsResources.map { newsResource ->
+            newsNotification {
+                setSmallIcon(
+                    com.google.samples.apps.nowinandroid.core.common.R.drawable.ic_nia_notification,
+                )
+                    .setContentTitle(newsResource.title)
+                    .setContentText(newsResource.content)
+                    .setGroup(NEWS_NOTIFICATION_GROUP)
+            }
+        }
+        val summaryNotification = newsNotification {
+            val title = getString(
+                R.string.news_notification_group_summary,
+                newsNotifications.size,
+            )
+            setContentTitle(title)
+                .setContentText(title)
+                .setSmallIcon(
+                    com.google.samples.apps.nowinandroid.core.common.R.drawable.ic_nia_notification,
+                )
+                // Build summary info into InboxStyle template.
+                .setStyle(newsInboxStyle(newsResources, title))
+                .setGroup(NEWS_NOTIFICATION_GROUP)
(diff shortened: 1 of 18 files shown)


core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
```diff
@@ -16,17 +16,152 @@
...
+            inboxStyle.addLine(newsResource.title)
+        }
+        .setBigContentTitle(title)
+        .setSummaryText(title)
+}
+
+/**
+ * Creates a notification for configured for news updates
+ */
+private fun Context.newsNotification(
```
dturner, line 119:
```suggestion
private fun Context.createNotification(
```

core/notifications/src/main/res/values/strings.xml
```diff
@@ -15,8 +15,8 @@
      limitations under the License.
 -->
 <resources>
-    <string name="sync_notification_title">Now in Android</string>
-    <string name="sync_notification_channel_name">Sync</string>
-    <string name="sync_notification_channel_description">Background tasks for Now in Android</string>
-
+    <string name="news_notification_title">Now in Android</string>
```
dturner, line 18: The `news_` prefix feels redundant here for these 4 names

core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
```diff
@@ -16,17 +16,152 @@
...
+private const val NEWS_NOTIFICATION_SUMMARY_ID = 1
+private const val NEWS_NOTIFICATION_CHANNEL_ID = ""
+private const val NEWS_NOTIFICATION_GROUP = "NEWS_NOTIFICATIONS"
+
 /**
  * Implementation of [Notifier] that displays notifications in the system tray.
  */
 @Singleton
-class AndroidSystemNotifier @Inject constructor() : Notifier {
+class AndroidSystemNotifier @Inject constructor(
```
dturner, line 49:
```suggestion
class SystemTrayNotifier @Inject constructor(
```

core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
```diff
@@ -16,17 +16,152 @@
...
+    @ApplicationContext private val context: Context,
+) : Notifier {
+
+    override fun onNewsAdded(
```
dturner, line 53:
```suggestion
    // Post a notification for the given list of news resources.
    override fun post(
```

or maybe `postNotification`?

dturner (changes requested):

dturner:
I tested on an API 29 device and there was no grouping, just a long list of notifications with a redundant "Now in Android" notification at the top:

image

I think we should limit the number of notifications to 5, regardless of how much new content has been downloaded. This will avoid overloading the user with notifications (which is highly likely to lead to insta-uninstalls).

Steps to repro: 

- change the binding in the `:core:notifications` module `NotificationModule` for the demo variant to `AndroidSystemNotifier`  
- Replace the `if (hasOnboarded)` block in `OfflineFirstNewsRepository` to: 

```
notifier.onNewsAdded(
                    newsResourceDao.getNewsResources()
                        .first()
                        .take(49)
                        .also { println("SIZE: ${it.size}") }
                        .map(PopulatedNewsResource::asExternalModel),
                )
```
- Run on API 29 emulator

This will post 49 news resource notifications on app startup.

core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
```diff
@@ -16,17 +16,152 @@
...
+        }
+        val summaryNotification = newsNotification {
+            val title = getString(
+                R.string.news_notification_group_summary,
+                newsNotifications.size,
+            )
+            setContentTitle(title)
+                .setContentText(title)
+                .setSmallIcon(
+                    com.google.samples.apps.nowinandroid.core.common.R.drawable.ic_nia_notification,
```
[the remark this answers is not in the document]
tunjid: Can't import `R` here as it is from another module. It will conflict with this module's `R`

core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
```diff
@@ -16,17 +16,152 @@
...
+    NotificationManagerCompat.from(this).createNotificationChannel(channel)
+}
+
+private fun Context.newsPendingIntent(newsResource: NewsResource): PendingIntent? =
+    PendingIntent.getActivity(
+        this,
+        NEWS_NOTIFICATION_REQUEST_CODE,
+        // TODO: Read color from material theme to style the chrome custom tab
+        //  this is currently only readable from composition.
+        CustomTabsIntent.Builder()
```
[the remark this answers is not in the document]
tunjid: I did this, but the experience of opening the app first, then linking externally is jarring. Would you be okay with keeping the current behavior but also having a pending intent that marks the task as read?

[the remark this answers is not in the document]
tunjid: Going back to launching the app, and starting the chrome custom tab because of [notification trampoline restrictions](https://developer.android.com/about/versions/12/behavior-changes-12#notification-trampolines). We should consider moving into an internal app webview because of this, launching 2 activities so quickly is really ugly.

core/notifications/src/main/res/values/strings.xml
tunjid: The notifier package will house notifications for a other types too. The namespacing helps with this

core/notifications/src/main/java/com/google/samples/apps/nowinandroid/core/notifications/AndroidSystemNotifier.kt
tunjid: Renamed to `postNewsNotifications` as this interface will also be responsible for creating any other kind of notification in the app.

tunjid:
@dturner  the notification grouping issue happens when the group is larger than the limit set by the system. Its 50 atm, but OEMs sometimes change this. I've set the limit to 50.

PR #713, opened 2023-05-05
Save the showSettingsDialog flag in instance state

rosejr:
This prevents the settings dialog from being dismissed when the device is rotated.

Since the flag is not used outside of NiaApp, it's moved there and stored with `rememberSaveable`.


dturner (approved):

rosejr:
Closes #611

PR #715, opened 2023-05-08
Replace manual URI handling with Compose's `LocalUriHandler`

SimonMarquis:


alexvanyo (approved):
Nice, I had no idea this API existed. This looks a lot nicer!

PR #716, opened 2023-05-08
Simplify handling of nullability in `ConnectivityManagerNetworkMonitor`

SimonMarquis:
Detect nullability of `ConnectivityManager` as soon as possible to avoid handling unnecessary states.


tunjid (approved):
LGTM, thanks for doing this.

PR #717, opened 2023-05-09
Improvement in the Search screen

thagikura:
Combine the search result of topics and news resources in one LazyVerticalGrid so that topics don't get in the way.

Before this change, the result of topics created a LazyColumn and the result of news resources created the LazyVerticalGrid separately that made the situation where there were two separate vertical scrolling lists.


dturner (approved):
Thanks for doing this @thagikura

PR #719, opened 2023-05-09
Add analytics for search queries

dturner:
This sends an analytics event for each user-initiated search query. This implementation will only capture those searches where the user actively taps on the search button. 

Other ideas/thoughts...

Most of the searches will happen in real time as the user enters their search keywords. I wonder if it'd be better to capture each one of these searches instead. To avoid capturing every intermediate search (every letter being entered) perhaps we could capture the event where search results are being displayed AND the user taps on the search area (the search box losing focus).  

@thagikura Any thoughts here?


thagikura:
Thanks for doing this!

I think the implementation looks good.
It's better to capture the search queries explicitly triggered by the user rather than capturing all the intermediate states.

By the way it looks like analyticsHelper is missing for the SearchViewModelTest

thagikura (approved):

PR #723, opened 2023-05-12
Replace NiaTopLevelNavigation with TopLevelDestination based on #186

Vixb1122:
Replace NiaTopLevelNavigation with TopLevelDestination based on #186


thagikura:
The change looks good. But is it possible to remove the comments that look left from templates?

Vixb1122:
@thagikura done!

thagikura (approved):
Thanks!

PR #731, opened 2023-05-16
Adjust JDK setup to remove toolchain

### LINKED ISSUE #727: [Bug]: Impossible to sync the project with embedded java of Android Studio Flamingo

### What happened?

I have Android Studio Flamingo | 2022.2.1 Patch 1 which is shipped with JDK 17.

With changes provided by https://github.com/android/nowinandroid/pull/583, the synchronization of the project fails with `No matching toolchains found for requested specification: {languageVersion=11, vendor=any, implementation=vendor-specific}.`

Seems like I would need to install a separate instance of Java to also have JDK 11 installed on my computer which is annoying.

It is possible to configure the jvmTarget without requiring a specific toolchain, this way the project can be compiled with the embedded version of Java which is JDK 17.

### Relevant logcat output

```shell
Failed to calculate the value of task ':core:model:compileJava' property 'javaCompiler'.
No matching toolchains found for requested specification: {languageVersion=11, vendor=any, implementation=vendor-specific}.
No locally installed toolchains match (see https://docs.gradle.org/8.1/userguide/toolchains.html#sec:auto_detection) and toolchain download repositories have not been configured (see https://docs.gradle.org/8.1/userguide/toolchains.html#sub:download_repositories).
```

### Code of Conduct

alexvanyo:
Hopefully one of the last of the JDK related PRs 🤞 

After this PR, all of the following should be true:

1. The app code (Android modules and JVM module) all target JDK 11 to match https://developer.android.com/studio/write/java11-minimal-support-table. This is configured in `KotlinAndroid.kt`, via convention plugins based on the type of module.
1. The project is built entirely using JDK 17, which matches the version embedded in Android Studio Flamingo, and is what is used on CI (GitHub Actions and internal). This removes all uses of `jvmToolchain` for `11`, which did help solve item 1, at the cost of needing two separate JDK versions available to build the project. This PR fixes #727.
1. The convention plugins in `build-logic` are configured to target JDK 17, which matches the JDK being used to build the project. This doesn't have anything to do with the code actually shipping to the device (item 1).

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -22,16 +22,15 @@ plugins {
 
 group = "com.google.samples.apps.nowinandroid.buildlogic"
 
+// Configure the build-logic plugins to target JDK 17
+// This matches the JDK used to build the project, and is not related to what is running on device.
 java {
-    // Up to Java 11 APIs are available through desugaring
-    // https://developer.android.com/studio/write/java11-minimal-support-table
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
 }
-
 tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_11.toString()
+        jvmTarget = JavaVersion.VERSION_17.toString()
     }
 }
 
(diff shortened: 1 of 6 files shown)


thagikura (approved):
Looking good, thanks Alex!

dturner (approved):
Thanks for doing this Alex

GSala:
It would be great to have some (better) official guidance on the topic of these PRs. Specifically understanding all the different Java/JDK versions involved in Android development, what implications target/source compatibility, and what desugaring allows us to do. Some of this is covered and spread around the docs, but I think it would be worth unifying it into a single topic.

GSala:
For future reference. New docs were posted about this ☝️ here: https://developer.android.com/build/jdks
@alexvanyo I'm not sure the current implementation is aligned with the docs though 🤔 We're not specifying a `jvmToolchain`.

PR #740, opened 2023-05-19
Add type-safe access to Gradle Catalog versions and libraries in `build-logic`

SimonMarquis:
I'd like to get your feedback on this change.

- Before:
  ```kotlin
  dependencies {
      val libs = extensions.getByType<VersionCatalogsExtension>().named("libs")
      add("implementation", libs.findLibrary("x.y.z").get())
  }
  ```

- After:
  ```kotlin
  dependencies {
      add("implementation", nia().libraries.`x-y-z`)
  }
  ```

If the name in the catalog changes, or an unknown name is requested, the following error will be thrown by Gradle:

```
An exception occurred applying plugin request [id: 'x.y.z']
> Failed to apply plugin 'x.y.z'.
   > Missing catalog library 'foo'
```

It uses a caching strategy (based on `extensions.extraProperties`) to prevent creating multiple instances of `NiaProperties`.

---

And if we really want to push the "transparent" API (and get the same names as we get in other Gradle modules) we could introduce a dynamic getter:

```kotlin
internal val Project.libs get() = NiaProperties(this).libraries
```

----

Edit: one step closer to #317

### DIFF
--- a/build-logic/convention/src/main/kotlin/AndroidApplicationFirebaseConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/AndroidApplicationFirebaseConventionPlugin.kt
@@ -16,12 +16,11 @@
 
 import com.android.build.api.dsl.ApplicationExtension
 import com.google.firebase.crashlytics.buildtools.gradle.CrashlyticsExtension
+import com.google.samples.apps.nowinandroid.nia
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.api.artifacts.VersionCatalogsExtension
 import org.gradle.kotlin.dsl.configure
 import org.gradle.kotlin.dsl.dependencies
-import org.gradle.kotlin.dsl.getByType
 
 class AndroidApplicationFirebaseConventionPlugin : Plugin<Project> {
     override fun apply(target: Project) {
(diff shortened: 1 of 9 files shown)


alexvanyo (commented):
My initial thoughts of this PR is that the additional indirection via delegates in `NiaProperties.kt` isn't worth the overhead for someone approaching the project for the first time and looking to adapt it to their own project.

Right now, the `libs.findLibrary("x.y.z").get()` isn't ideal, but it is using the API as provided by Gradle at the moment. With the changes in this PR, that isn't something built-in, and it would require copying in `NiaProperties.kt` or something similar.

There's also the factor that Gradle might eventually have a better solution to #317 , in which case this would hopefully become obsolete.

So overall, I think it looks good as something you may want to consider adding to your own project if you find it useful to avoid the string-based API, but I don't know if we want to add this to Now in Android.

SimonMarquis:
👍 I agree, this might not be the best fit for this project.
I'll close this PR.

I was thinking of adding this instead, to match the current Gradle API as much as possible and avoid repeating this, wdyt?

```
val Project.libs
    get(): VersionCatalog = extensions.getByType<VersionCatalogsExtension>().named("libs")
```

dturner:
I like this as it removes the duplicated `val libs = ...` code without adding much complexity. Suggest adding it to a new file named `PluginExt.kt` or `ProjectPluginExt.kt` unless you have a better suggestion?

SimonMarquis:
I've created a new PR to extract this into an extension.
I've also chosen `ProjectExtensions.kt` for the filename to follow what is already used in this project.
Having the type name followed by `Extensions` feels more natural, but I can still change it with something else if you prefer.

- #802

I'll close this issue for now. Let's hope proper type safe accessor will come for included builds soon. 🤞

PR #748, opened 2023-05-24
Grant `POST_NOTIFICATIONS` permission in more instrumented tests

SimonMarquis:
Continues the work initiated in #738.
Extract the SDK version check inside a `GrantPostNotificationPermissionRule` class that delegates to a regular `GrantPermissionRule`.


thagikura (approved):
LGTM!

PR #749, opened 2023-05-25
Ensure DAO exposes only reactive types

tunjid:
The [architecture design](https://github.com/android/nowinandroid/blob/main/docs/ArchitectureLearningJourney.md) of NiA requires that all types exposed by the data layer are reactive. This is make sure consumers are aware that data may change over time and must be ready to react to that change.

If a consumer wants a snapshot of data, they _MUST_ explicitly opt out of reactivity, in the case of the PR by using the `first()` operator. This is also done in `OfflineFirstNewsRepository` when it is posting notifications.


dturner (commented):
This is great and brings NiA in line with our data layer guidance. What about the other DAOs though e.g. `TopicDao`?

thagikura (approved):
This looks good. Thanks!

tunjid:
@dturner `TopicDao` currently exposes reactive types only.

PR #750, opened 2023-05-25
Add general application scope, use IO dispatcher for DataStore

dturner:
After chatting to colleagues (@manuelvicnt and @zsmb13) about @SimonMarquis's PR (https://github.com/android/nowinandroid/pull/681) to add an application-wide scope, we concluded that there should only be a single application scope, and if consumers of that scope (e.g. for DataStore access) need to use a different dispatcher then they should just override it, like so: 

```
CoroutineScope(applicationScope.context + differentDispatcher)
```

This PR implements those changes.

### DIFF
--- a/core/common/src/main/java/com/google/samples/apps/nowinandroid/core/network/di/CoroutineScopesModule.kt
+++ b/core/common/src/main/java/com/google/samples/apps/nowinandroid/core/network/di/CoroutineScopesModule.kt
@@ -17,7 +17,7 @@
 package com.google.samples.apps.nowinandroid.core.network.di
 
 import com.google.samples.apps.nowinandroid.core.network.Dispatcher
-import com.google.samples.apps.nowinandroid.core.network.NiaDispatchers.IO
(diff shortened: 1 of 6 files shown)


manuelvicnt (approved):

SimonMarquis:
You'll probably need to update other instances of `@Dispatcher(IO) ioScope: CoroutineScope,` in the codebase.

dturner:
Getting a seemingly unrelated error when running local tests after submitting this change: 

```
> Task :core:datastore-test:compileDemoDebugKotlin FAILED
e: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class com.google.samples.apps.nowinandroid.core.datastore.UserPreferences, unresolved supertypes: com.google.protobuf.GeneratedMessageLite
    class com.google.samples.apps.nowinandroid.core.datastore.UserPreferencesOrBuilder, unresolved supertypes: com.google.protobuf.MessageLiteOrBuilder
Adding -Xextended-compiler-checks argument might provide additional information.
```

Let's see if CI throws the same error

PR #752, opened 2023-05-25
Kotlin 1.9.0

SimonMarquis:
This PR is testing future support for Kotlin 1.9.0:

- https://github.com/JetBrains/kotlin/releases/tag/v1.9.0
- https://github.com/google/ksp/releases/tag/1.9.0-1.0.11
- https://github.com/google/dagger/releases/tag/dagger-2.47
- Compose compiler [`1.5.0`](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.0)
- https://github.com/protocolbuffers/protobuf/releases/tag/v23.4

> **Warning**
> Had to force register generated protobuf sources to Android sourceSets otherwise Hilt fails to resolve `UserPreferences` generated type.
> ```kotlin
> androidComponents.beforeVariants {
>     android.sourceSets.register(it.name) {
>         java.srcDir(buildDir.resolve("generated/source/proto/${it.name}/java"))
>         kotlin.srcDir(buildDir.resolve("generated/source/proto/${it.name}/kotlin"))
>     }
> }
> ```

### Remaining issues

- [x] https://github.com/android/nowinandroid/actions/runs/5083859526/jobs/9135396287?pr=752#step:8:971 Unexpected lint error 🤯
  Adding an explicit `return@remember` fixes the issue.
  ```
  nowinandroid\app\src\androidTest\java\com\google\samples\apps\nowinandroid\ui\NiaAppStateTest.kt:186: Error: remember calls must not return Unit [RememberReturnType from androidx.compose.runtime]
    val navController: TestNavHostController = remember {
                                               ~~~~~~~~

   Explanation for issues of type "RememberReturnType":
   A call to remember that returns Unit is always an error. This typically
   happens when using remember to mutate variables on an object. remember is
   executed during the composition, which means that if the composition fails
   or is happening on a separate thread, the mutated variables may not reflect
   the true state of the composition. Instead, use SideEffect to make deferred
   changes once the composition succeeds, or mutate MutableState backed
   variables directly, as these will handle composition failure for you.
  ```
- [x] https://github.com/android/nowinandroid/actions/runs/5083859526/jobs/9135396287?pr=752#step:8:849
  "Fixed" by forcing a more recent version of Lint to AGP.
  `.../transformed/out/jars/classes.jar!/META-INF/....kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.`

- [x] ~Enabling K2 leads to the following errors~ out of scope
  ```
  [Hilt] All modules must be static and use static provision methods or have a visible, no-arg constructor. Found: [...]Module - com.google.samples.apps.nowinandroid.NiaApplication
  ```


lihenggui:
Now Kotlin 1.9.0 was officially released. We can upgrade the dependencies:

```
androidxComposeCompiler = "1.5.0-dev-k1.9.0-6a60475e07f"
kotlin = "1.9.0"
ksp = "1.9.0-1.0.11"
protobuf = "3.23.4"
```
So far, I don't see issues after upgrading these dependencies. 
And I have a question, disabling the build cache is for testing, it that correct?

SimonMarquis:
I'll update the dependencies, but there is still an ongoing issue with Lint.
AGP 8.0.2 uses internally Lint 31.0.2 which is compiled with metadata in version 1.7.1. This makes the Lint analyse tasks print errors.
A temporary solution is to bump the lint version to the most recent, and most stable, and compatible version which is 31.1.0-rc, or simply just wait for the next AGP stable version.

SimonMarquis:
Also, Lint errors similar to the `return@remember` still exists in my work app even with 1.9.0, so 🤷

SimonMarquis:
And we probably will wait for a stable compose compiler version anyway. (not relying on `1.5.0-dev-k1.9.0-6a60475e07f`)

zsmb13:
Bumped Compose to stable 1.5.0, added the Dagger version bump to get kotlinx.metadata up-to-date, and bumped lint to 31.0.2 to hopefully clean up the errors that it was spitting out earlier.

With that we should be good to merge this once CI is done.

SimonMarquis:
This bump in lint version will not impact the issue, which is related to the lint version used by AGP.

zsmb13:
I see, do you know if these errors cause any practical problems? I see that lint tasks still succeed and apparently create reports. They are also able to fail on lint errors in Kotlin code (like that false positive on `remember` if I remove the fix 😅).

SimonMarquis:
To be honest, I'm not sure about that. It will probably fail to parse and interpret some specific Lint issues for sure. But to me the main problem is that it pops up in the IDE and "look like" an error, which means, you'll probably have many reports saying that it "fails to run lint". If we want to still merge this update, I would instead merge the AGP's Lint version to the most recent `31.1.0-rc` and revert this once AGP updates. WDYT?

gradle.properties
```diff
@@ -38,3 +38,7 @@ android.nonTransitiveRClass=true
 # https://developer.android.com/build/releases/gradle-plugin#default-changes
 android.defaults.buildfeatures.resvalues=false
 android.defaults.buildfeatures.shaders=false
+
+# Use newer lint version
```
SimonMarquis, line 42: nit: we should probably explain the "why" in the comment, wdyt?

```suggestion
# Use newer lint version to support Kotlin 1.9 and corresponding kotlinx-metadata-jvm
```

SimonMarquis:
Thanks for the updates @dturner & @zsmb13.
I've updated the PR description, and I guess we are now ready to merge this once the CI validates it.
🚀

gradle.properties
dturner: Yes, much better, thanks.

dturner (approved):
Thanks very much for doing this.

SimonMarquis:
Workaround removed in
- #854

PR #753, opened 2023-05-25
Fix lint warnings `ComposableModifierFactory` and `ModifierFactoryExtensionFunction`

SimonMarquis:
- `ComposableModifierFactory`: Modifier factory functions that need to be aware of the composition should use androidx.compose.ui.composed {} in their implementation instead of being marked as @Composable. This allows Modifiers to be referenced in top level variables and constructed outside of the composition.

- `ModifierFactoryExtensionFunction`: Modifier factory functions should be defined as extension functions on Modifier to allow modifiers to be fluently chained.


alexvanyo (approved):

SimonMarquis:
Closed, superseded by #765

PR #754, opened 2023-05-26
Compose compiler metrics saved into one folder

mlykotom:
I found it difficult to navigate to each module to find compose metrics.
This way, it's saved directly in the root `build/compose-metrics` folder from all modules, so it's easier to find the files when investigating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,14 @@ The app uses adaptive layouts to
 
 Find out more about the [UI architecture here](docs/ArchitectureLearningJourney.md#ui-layer).
 
-# Baseline profiles
+# Performance
+
+## Benchmarks
+
+Find all tests written using [`Macrobenchmark`](https://developer.android.com/topic/performance/benchmarking/macrobenchmark-overview)
+in the `benchmarks` module. This module also contains the test to generate the Baseline profile.
+
+## Baseline profiles
 
 The baseline profile for this app is located at [`app/src/main/baseline-prof.txt`](app/src/main/baseline-prof.txt).
 It contains rules that enable AOT compilation of the critical user path taken during app launch.
@@ -143,6 +150,16 @@ For more information on baseline profiles, read [this document](https://develope
 To generate the baseline profile, select the `benchmark` build variant and run the
 `BaselineProfileGenerator` benchmark test on an AOSP Android Emulator.
 Then copy the resulting baseline profile from the emulator to [`app/src/main/baseline-prof.txt`](app/src/main/baseline-prof.txt).
+
+## Compose compiler metrics
+
+Run the following command to get and analyse compose compiler metrics:
+
+```
+./gradlew assembleRelease -PenableComposeCompilerMetrics=true -PenableComposeCompilerReports=true
+```
+
+For more information on Compose compiler metrics, see [this blog post](https://medium.com/androiddevelopers/jetpack-compose-stability-explained-79c10db270c8).
 
 # License
 
--- a/README.md
+++ b/README.md
@@ -159,8 +159,8 @@ Run the following command to get and analyse compose compiler metrics:
 ./gradlew assembleRelease -PenableComposeCompilerMetrics=true -PenableComposeCompilerReports=true
 ```
 
-The reports files will be added to build/compose-reports in root folder. The metrics files will also be 
-added to build/compose-metrics in root folder.
+The reports files will be added to [build/compose-reports](build/compose-reports). The metrics files will also be 
+added to [build/compose-metrics](build/compose-metrics).
 
 For more information on Compose compiler metrics, see [this blog post](https://medium.com/androiddevelopers/jetpack-compose-stability-explained-79c10db270c8).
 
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -159,8 +159,8 @@ Run the following command to get and analyse compose compiler metrics:
 ./gradlew assembleRelease -PenableComposeCompilerMetrics=true -PenableComposeCompilerReports=true
 ```
 
-The reports files will be added to build/compose-reports in each module. The metrics files will be 
-added to build/compose-metrics in each module.
+The reports files will be added to build/compose-reports in root folder. The metrics files will also be 
+added to build/compose-metrics in root folder.
```
SimonMarquis, line 163: Let's use direct link instead, wdyt?

```suggestion
The reports files will be added to [build/compose-reports](build/compose-reports). The metrics files will also be 
added to [build/compose-metrics](build/compose-metrics).
```

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -61,7 +61,7 @@ private fun Project.buildComposeMetricsParameters(): List<String> {
     val enableMetricsProvider = project.providers.gradleProperty("enableComposeCompilerMetrics")
     val enableMetrics = (enableMetricsProvider.orNull == "true")
     if (enableMetrics) {
-        val metricsFolder = File(project.buildDir, "compose-metrics")
+        val metricsFolder = File(project.rootProject.buildDir, "compose-metrics")
```
SimonMarquis, line 64:
```suggestion
        val metricsFolder = rootProject.buildDir.resolve("compose-metrics")
```

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -71,7 +71,7 @@ private fun Project.buildComposeMetricsParameters(): List<String> {
     val enableReportsProvider = project.providers.gradleProperty("enableComposeCompilerReports")
     val enableReports = (enableReportsProvider.orNull == "true")
     if (enableReports) {
-        val reportsFolder = File(project.buildDir, "compose-reports")
+        val reportsFolder = File(project.rootProject.buildDir, "compose-reports")
```
SimonMarquis, line 74:
```suggestion
        val reportsFolder = rootProject.buildDir.resolve("compose-reports")
```

SimonMarquis:
One potential downside of this is that thoses files are using the project name and not their complete path. This means one report would be overwritten by another if it has the same name (and not the same path).
For instance: `:feature:foo` and `:core:foo` would collide.

mlykotom:
> One potential downside of this is that thoses files are using the project name and not the complete project path. This means one report would be overwritten by another if it has the same name (and not the same path). For instance: `:feature:foo` and `:core:foo` would collide.

Added `project.path` as a subfolder. I think this way, it's even more clear
image

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -59,9 +58,11 @@ internal fun Project.configureAndroidCompose(
 private fun Project.buildComposeMetricsParameters(): List<String> {
     val metricParameters = mutableListOf<String>()
     val enableMetricsProvider = project.providers.gradleProperty("enableComposeCompilerMetrics")
+    val projectPath = project.path.replace(':', '_').drop(1)
+
     val enableMetrics = (enableMetricsProvider.orNull == "true")
     if (enableMetrics) {
-        val metricsFolder = File(project.buildDir, "compose-metrics")
+        val metricsFolder = rootProject.buildDir.resolve("compose-metrics/$projectPath")
```
SimonMarquis, line 65: `projectPath` will contain invalid characters on Windows: `:`.

Does this work?

```suggestion
        val metricsFolder = rootProject.buildDir.resolve(projectDir.relativeTo(rootDir))
```

SimonMarquis: Ooops, I did not see the `project.path.replace(':', '_').drop(1)` at first.
Both solutions are valid here.
One potential issue (that does not appear to happen in this project) is that replacing path chars could also lead to collision, but is very unlikely. (e.g. `:foo:bar_baz` and `:foo_bar:baz`)

This would not happen with the relative directory method, but will add maybe cumbersome directories 🤷 
```kotlin
val relativePath = projectDir.relativeTo(rootDir)
val metricsFolder = rootProject.buildDir.resolve("compose-metrics").resolve(relativePath)
val reportsFolder = rootProject.buildDir.resolve("compose-reports").resolve(relativePath)
```

And the output:
image

mlykotom: I agree it's cumbersome, but since people might be copying the solution from this sample, let's make it safe, so it doesn't create confusion.

I think it's still beneficial to have it in one folder with subfolders than finding the metrics in each module.

dturner (approved):

PR #756, opened 2023-05-27
Cleanup proguard configuration

SimonMarquis:
- Remove custom KotlinX Serialization config as it's already provided, see https://github.com/Kotlin/kotlinx.serialization#android.
- Delete empty config files.

image


dturner:
@SimonMarquis Always good to see redundant code being deleted, thanks. 

For any other readers, the `kotlinx.serialization` rules are here so don't need to be in NiA: https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro

dturner (approved):

PR #759, opened 2023-05-27
Android Lint improvements

SimonMarquis:
- Create `LintConventionPlugin` to configure Lint on all compatible modules: Android applications, Android libraries and JVM modules.
- Run `lintProdRelease` in CI instead of the default `lintDemoDebug` which is less important compared to production code.
- Rearrange CI steps to make it more clear that Lint should is an additional step after build (build -> test -> lint).
- Enable SARIF support and upload results to GitHub's CodeQL to get inline feedback on PRs.

If we really need better perfs, we could restore `lint.checkDependencies = true` on the `:app` module and only execute `:app:lintProdRelease`. But in practice, this does not change the total build time on this project.


SimonMarquis:
```
Error: Code Scanning could not process the submitted SARIF file:
rejecting SARIF, as there are more runs than allowed (23 > 20)
```

We could either:
- merge multiple reports into a single one (https://github.com/ZacSweers/CatchUp/blob/77bbfe95fc1256203b7fb614b70dd4762b8cb921/.github/workflows/ci.yml#L97-L100 & https://github.com/ZacSweers/CatchUp/blob/main/.github/workflows/sarif-filter.main.kts)
- limit to the first 20 by size
- split into two steps and upload the first 20 and then the remaining ones
- use `lint.checkDependencies = true` on the `app` module instead of executing all module independently.

SimonMarquis:
Here are some examples of warnings/errors provided by CodeQL, wdyt?

dturner:
This sure looks pretty :) but not sure of the benefit vs increased complexity (another thing to go wrong and maintain). Will discuss in our next standup.

SimonMarquis:
I've removed the SARIF support.
Do you want me to create a new dedicated PR for it? (if so, I'll do once this one is merged)

SimonMarquis:
@dturner PTAL 😉

mmoczkowski (approved):
LGTM, thank you for your contribution!

SimonMarquis:
> Could we split out the SERIF part of this PR for further evaluation?

FYI @mmoczkowski I've created a dedicated PR (currently in draft) that specifically enables SARIF support
- #921

Issue #760, opened 2023-05-27
[FR]: set resource independence at multi module


riflockle7:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

I always get insights by looking at the nowinandroid code, and I personally use the code. Thank you every time.
I found an obscure problem while using it and left it as a "github issue".

---

Currently in nowinandroid, "top_app_bar_title" is used in the ForYou tab, Saved tab, and Interests tab.
Looking at the Find in files, it looks like this.

<image src=https://github.com/android/nowinandroid/assets/13694046/639f6b32-5688-4991-a8cc-c223e09d1423 width=480>

I tested each module in nowinandroid's feature.

### 1. First, I tried putting the code below in the Screen composable of each tab.
```kotlin

Text(stringResource(id = R.string.top_app_bar_title))
```
so I added `BookmarksScreen.kt`, `ForYouScreen.kt`, `InterestsScreen.kt`.

### 2. So, the result we faced was as follows.
<img width="480" alt="image" src="https://github.com/android/nowinandroid/assets/13694046/5759cbcc-a2b5-4149-bb53-bad540b1a77d">

Instead of strings.xml defined in each module, we only use values ​​from a specific strings.xml 
(the interest module in this example).

### 3. Also tested in the search module (SearchScreen.kt)
I applied it once to the search module as well, but the result was the same.
<img width="480" alt="image" src="https://github.com/android/nowinandroid/assets/13694046/a5379ad3-c803-4b3d-b4f0-288501c6abea">

### Describe the solution

My guess is that `the resource is either merged from multiple modules into one` or `the last value found within the resource is selected`.

It would be nice if there was a guideline or corresponding code, but is there a solution?

Thank you

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SimonMarquis:
Unfortunately, `top_app_bar_title` is not used in the project.
These strings should have been removed.

The general guidance here could be to simply rely on `resourcePrefix` and lint validation:
For example, in the `:feature:search` module:

```kotlin
android {
    namespace = "com.google.samples.apps.nowinandroid.feature.search"
    resourcePrefix = "feature_search_"
}
```

This could enforce the following prefix for strings (and other resources):

```xml
<resources>
    <string name="feature_search_title">Search</string>
    <!--...-->
<resources/>
```

SimonMarquis:
@dturner is this something that you would want for this project to follow for each module?

[...]

PR #763, opened 2023-05-28
Cleanup unused string resources

SimonMarquis:
References #760


mmoczkowski (approved):
Thanks for keeping the codebase clean

PR #765, opened 2023-05-29
Notification dot improvements

mmoczkowski:
1. Refactor `notificationDot` function to use `composed` instead of `@Composable`
2. Refactor `notificationDot` function to be extension of `Modifier`
3. Add notification dots to the navRail (previously only bottom bar supported it)

Screenshot_20230529_122416


thagikura (approved):
Looking good!

PR #767, opened 2023-05-29
Enforce `resourcePrefix` on Android library modules

### LINKED ISSUE #760: [FR]: set resource independence at multi module

SimonMarquis:
See this issue for the rationale behind this change:

- https://github.com/android/nowinandroid/issues/760

```
:ui-test-hilt-manifest -> ui_test_hilt_manifest_
:core:analytics -> core_analytics_
:core:common -> core_common_
:core:data -> core_data_
:core:data-test -> core_data_test_
:core:database -> core_database_
:core:datastore -> core_datastore_
:core:datastore-test -> core_datastore_test_
:core:designsystem -> core_designsystem_
:core:domain -> core_domain_
:core:network -> core_network_
:core:notifications -> core_notifications_
:core:testing -> core_testing_
:core:ui -> core_ui_
:feature:bookmarks -> feature_bookmarks_
:feature:foryou -> feature_foryou_
:feature:interests -> feature_interests_
:feature:search -> feature_search_
:feature:settings -> feature_settings_
:feature:topic -> feature_topic_
:sync:sync-test -> sync_test_
:sync:work -> sync_work_
```

And Lint will fail as expected with these kind of errors:

```
nowinandroid\feature\interests\src\main\res\values\strings.xml:18: Error: Resource named 'interests' does not start with the project's resource prefix 'feature_interests_'; rename to 'feature_interests_interests' ? [ResourceName]
    <string name="interests">Interests</string>
            ~~~~~~~~~~~~~~~~

   Explanation for issues of type "ResourceName":
   In Gradle projects you can specify a resource prefix that all resources in
   the project must conform to. This makes it easier to ensure that you don't
   accidentally combine resources from different libraries, since they all end
   up in the same shared app namespace.
```

Closes #760


build-logic/convention/src/main/kotlin/AndroidLibraryConventionPlugin.kt
```diff
@@ -42,6 +42,7 @@ class AndroidLibraryConventionPlugin : Plugin<Project> {
                 defaultConfig.targetSdk = 33
                 configureFlavors(this)
                 configureGradleManagedDevices(this)
+                resourcePrefix = path.split("""\W""".toRegex()).drop(1).distinct().joinToString(separator = "_").lowercase() + "_"
```
dturner, line 45:
```suggestion
                // The resource prefix is derived from the module name,
                // so resources inside ":core:module1" must be prefixed with  "core_module1_" 
                resourcePrefix = path.replace(":([\\w\\-]+)".toRegex(), "$1_")
```

dturner: Really nice idea, thanks. 

I simplified your regex, the only difference being that dashes in the module name are retained, rather than being replaced for underscores.

Now all that needs to be done is to update all the resource names to match, which is a fairly large change. Would you be able to do this? There's a chance this might cause some breaking changes for in-progress PRs, but hopefully not.

SimonMarquis: @dturner I've update the PR with the new resource names. There was one "error" [here](https://github.com/android/nowinandroid/pull/767/commits/6c941ccd96e75eab2b812523197dd6c2c88906dd#diff-9df87753608fc53600f9e148c0e20ee6f0a40520c39283a2f4be7f5cf64ca16fL24-R24).
And I've taken the liberty to rename some resources where it would have produce duplicated words, as well as "title" strings to be more explicit.
Should we keep the current regex or should we find a "simpler" (smaller) regex?

app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
```diff
@@ -51,7 +51,7 @@ import javax.inject.Inject
 import kotlin.properties.ReadOnlyProperty
 import com.google.samples.apps.nowinandroid.feature.bookmarks.R as BookmarksR
 import com.google.samples.apps.nowinandroid.feature.foryou.R as FeatureForyouR
-import com.google.samples.apps.nowinandroid.feature.interests.R as FeatureInterestsR
+import com.google.samples.apps.nowinandroid.feature.search.R as FeatureSearchR
```
MercuryLiRC, line 54: Should start with uppercase or lowercase?

SimonMarquis:
🔔 @dturner 🙈
Were there any remaining changes you wanted before merging this?

build-logic/convention/src/main/kotlin/AndroidLibraryConventionPlugin.kt
dturner: Let's keep your version, it just took me a while to figure out what it was doing :)

dturner (approved):

PR #768, opened 2023-05-29
Unify const naming conventions

mmoczkowski:
https://developer.android.com/kotlin/style-guide#constant_names
https://kotlinlang.org/docs/coding-conventions.html#property-names


thagikura:
I think NiA should follow the same naming convention for Jetpack Compose API guideline as in the document?

https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#singletons-constants-sealed-class-and-enum-class-values


If so, I agree it's better to rename some of the variables that don't follow the guideline, e.g. `bookmarksRoute` to `BookmarksRoute`. But not sure if we should follow the CAPITALS_AND_UNDERSCORES case.
The guideline notes following, but I think NiA doesn't so heavily invest in CAPITALS_AND_UNDERSCORES.
```
Library and app code with a strong existing investment in CAPITALS_AND_UNDERSCORES in their codebase MAY opt for local consistency with that pattern instead.
```

PR #775, opened 2023-05-30
Fix benchmarks

mlykotom:
- fix ScrollForYou benchmark (needs to be WARM)
- fix BaselineProfileGenerator to accept notification
- removes startup benchmarks abstraction to be able to run just one benchmark


mmoczkowski (commented):
It looks ok, but I have limited experience with benchmarks so I would like someone else to review it

keyboardsurfer (approved):

PR #782, opened 2023-06-02
Update default build features value in `gradle.properties`

SimonMarquis:
Default values have changed as you can see in https://developer.android.com/studio/releases/gradle-plugin#buildFeatures since AGP 8.

Personnaly, I would remove the last two remaining settings from this config altogether, as they don't bring any negative build time impact as long as you don't use the feature.

WDYT? Would you remove them or keep them?


dturner:
> Personnaly, I would remove the last two remaining settings from this config altogether, as they don't bring any negative build time impact as long as you don't use the feature.

I'd prefer to keep them explicitly disabled to avoid a situation where our (already long) build time increases unintentionally because someone started using these features (e.g. reading/writing resource values) without evaluating the negative impact on build time first.

dturner (approved):

PR #783, opened 2023-06-02
Create `externalDependencies.xml`

SimonMarquis:
This will ensure IntelliJ IDEA will prompt the user to install these plugins when contributing to the project.
https://www.jetbrains.com/help/idea/managing-plugins.html#required-plugins

### DIFF
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalDependencies">
+    <plugin id="com.intellij.copyright" />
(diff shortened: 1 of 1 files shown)


dturner:
I'm not in favour of adding this as it implies that we support IntelliJ IDEA for development. Obviously external contributors are welcome to use whatever IDE they like to contribute as long as the code conforms to our guidelines, but Android Studio is the only officially supported IDE.

I will leave open and raise with the team though to see if there are other opinions.

SimonMarquis:
When I said IntelliJ IDEA, I also meant Android Studio. Both share the same settings.
`externalDependencies.xml` is developped on the IntelliJ IDE but is compatible with all IDE based on it.

PR #758, opened 2023-05-27
Remove `ExperimentalCoroutinesApi` and `FlowPreview` opt-in

SimonMarquis:
```
> Task :core:model:compileKotlin
w: Opt-in requirement marker kotlinx.coroutines.ExperimentalCoroutinesApi is unresolved. Please make sure it's present in the module dependencies
w: Opt-in requirement marker kotlinx.coroutines.FlowPreview is unresolved. Please make sure it's present in the module dependencies
```

PR #784, opened 2023-06-02
Revert "Remove `ExperimentalCoroutinesApi` and `FlowPreview` opt-in"

SimonMarquis:
Reverts android/nowinandroid#758

Unfortunately, the "real" issue was that `:core:model` could not resolve these Opt-in requirement markers because it does not use Kotlin coroutines.
The opt-in could be declared only on modules using the dependencies, but then it becomes tedious to fix.

Sorry for that.


dturner (approved):
No problem, thanks for spotting and reverting

PR #785, opened 2023-06-02
Optimize AVD to fix long vector paths Lint warning

SimonMarquis:
... with SVGOM, but you'll have to check on Android Studio the AVD diff.

https://googlesamples.github.io/android-custom-lint-rules/checks/VectorPath.md.html


SimonMarquis:
🔔 @alexvanyo (sorry for the ping, you were the last to update these files)

lihenggui:
Hi @SimonMarquis Could you please share how to reduce the 'long path' in the source file? I often struggle with reducing the length of SVG paths as it's not very intuitive.

SimonMarquis:
@lihenggui good question! I don't exactly remembered what I did back in June 🙈, but I think for these kind of AVD files, when you don't have access to the source SVG paths, I "simply" recreated an SVG file with the same paths and run them through [SVGOMG](https://jakearchibald.github.io/svgomg/) and finally backported the compressed paths. Does that makes sense?

lihenggui:
> @lihenggui good question! I don't exactly remembered what I did back in June 🙈, but I think for these kind of AVD files, when you don't have access to the source SVG paths, I "simply" recreated an SVG file with the same paths and run them through [SVGOMG](https://jakearchibald.github.io/svgomg/) and compress them until there was noticable changes at the expected zoom level and finally backported the compressed paths to the AVD files. Does that makes sense?

Thank you! I will attempt to eliminate warnings in my projects.

dturner:
Before: 

image

After: 

image

dturner (approved):

PR #789, opened 2023-06-07
Fix ForYouScreen Compose Previews

amaurymedeiros:
Permissions should only be called in an Activity context, which is a layoutlib limitation. We need to avoid launching the permission request when in LocalInspectionMode, otherwise we'll have render errors on the ForYouScreen previews.


feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -406,6 +407,8 @@ fun TopicIcon(
 @Composable
 @OptIn(ExperimentalPermissionsApi::class)
 private fun NotificationPermissionEffect() {
+    // Permissions should be called from in an Activity Context, which is not present in previews
```
dturner, line 410:
```suggestion
    // Permissions should be only called from an Activity Context, which is not present in previews
```

amaurymedeiros (commented):
Fixed the inline comment as suggested.

PR #792, opened 2023-06-10
Update GMD workflow

SimonMarquis:
- Add `--no-parallel --max-workers=1` (was not able to trigger the error locally whit this combination).
- Remove useless `cleanManagedDevices` as devices are not persisted across runs.
- Rename test reports artifact to `test-reports-GMD` similar to `test-reports-${{ matrix.api-level }}`.
- Print disk space usage in case of failure.

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -138,13 +138,16 @@ jobs:
         run: ./gradlew packageDemoDebug packageDemoDebugAndroidTest
 
       - name: Run instrumented tests with GMD
-        run: ./gradlew cleanManagedDevices --unused-only &&
-          ./gradlew ciDemoDebugAndroidTest -Dorg.gradle.workers.max=1
+        run: ./gradlew ciDemoDebugAndroidTest --no-parallel --max-workers=1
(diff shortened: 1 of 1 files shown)


JoseAlcerreca:
This LGTM, can you mark it ready for review?

.github/workflows/Build.yaml
```diff
@@ -138,13 +138,16 @@ jobs:
         run: ./gradlew packageDemoDebug packageDemoDebugAndroidTest
 
       - name: Run instrumented tests with GMD
-        run: ./gradlew cleanManagedDevices --unused-only &&
-          ./gradlew ciDemoDebugAndroidTest -Dorg.gradle.workers.max=1
+        run: ./gradlew ciDemoDebugAndroidTest --no-parallel --max-workers=1
```
mmoczkowski, line 141: `--no-parallel --max-workers=1`

This looks like it might substantially increase the execution time.

SimonMarquis: Since we are not sharing in any way, I don't think this will have any impact.

mmoczkowski (approved):

PR #798, opened 2023-06-13
remove beforeVariants check for benchmark

yenerm:
Removes beforeVariants check for benchmark


mlykotom:
Now in Android doesn't use Baseline Profiles Gradle Plugin (yet). I think we should keep the build type until it's implemented. @keyboardsurfer WDYT?

anilu258 (commented):
> [](url)

keyboardsurfer:
This issue is considered stale. We are missing input to take further action.  
If the problem persists, please file a [new issue](new) with the missing information and link back to this issue.

PR #799, opened 2023-06-14
Add contentType to news feed

mlykotom:
Small change with some performance improvements when scrolling the feed.
Tested on Pixel 6 API 33.

Before
```
ScrollForYouFeedBenchmark_scrollFeedCompilationFull
frameDurationCpuMs   P50   4.9,   P90   6.9,   P95   7.6,   P99   9.5
frameOverrunMs       P50  -10.3,   P90  -6.8,   P95  -2.8,   P99  -2.1
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```

After
```
ScrollForYouFeedBenchmark_scrollFeedCompilationFull
frameDurationCpuMs   P50   4.6,   P90   6.3,   P95   7.2,   P99   9.2
frameOverrunMs       P50  -10.6,   P90  -8.7,   P95  -7.8,   P99  -4.7
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```

### DIFF
--- a/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
+++ b/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
@@ -58,7 +58,7 @@ fun LazyGridScope.newsFeed(
         NewsFeedUiState.Loading -> Unit
         is NewsFeedUiState.Success -> {
             items(
-                feedState.feed,
+                items = feedState.feed,
                 key = { it.id },
                 contentType = { "newsFeedItem" },
             ) { userNewsResource ->
--- a/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
+++ b/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
@@ -57,7 +57,11 @@ fun LazyGridScope.newsFeed(
     when (feedState) {
         NewsFeedUiState.Loading -> Unit
         is NewsFeedUiState.Success -> {
-            items(feedState.feed, key = { it.id }) { userNewsResource ->
+            items(
+                feedState.feed,
+                key = { it.id },
+                contentType = { "newsFeedItem" },
+            ) { userNewsResource ->
                 val resourceUrl by remember {
                     mutableStateOf(Uri.parse(userNewsResource.url))
                 }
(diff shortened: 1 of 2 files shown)


feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
```diff
@@ -182,7 +182,7 @@ internal fun ForYouScreen(
             onTopicClick = onTopicClick,
         )
 
-        item(span = { GridItemSpan(maxLineSpan) }) {
+        item(span = { GridItemSpan(maxLineSpan) }, contentType = { "bottomSpacing" }) {
```
alexvanyo, line 185: Nit: Why is this `item` `contentType` a lambda, but the `contentType` in the `item` later in this file just a `String`? Not sure if it matters at all, but seems inconsistent.

mlykotom: This is what the API requires.

When you use `items` (plural), it gives you lambda, because you can use the item inside the lambda to decide on `contentType`.

For `item` (singular)`, it takes directly the string, because you know what item you pass.

dturner: Possibly because `LazyGridScope.items` (with an s) has a `contentType` lambda rather than a simple type. There doesn't seem to be any need for it here.

dturner: @mlykotom Posted my previous comment before I saw that you'd replied. You're calling `item` not `items` so should use `contentType = "bottomSpacing"`.

core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
```diff
@@ -57,7 +57,11 @@ fun LazyGridScope.newsFeed(
     when (feedState) {
         NewsFeedUiState.Loading -> Unit
         is NewsFeedUiState.Success -> {
-            items(feedState.feed, key = { it.id }) { userNewsResource ->
+            items(
+                feedState.feed,
```
dturner, line 61: nit: use named argument for consistency
```suggestion
                items = feedState.feed,
```

dturner: I realise this wasn't your code but might as well fix it while you're here :)

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt
mlykotom: Ah, good catch. Fixed now.

dturner (approved):

PR #801, opened 2023-06-15
Fix HTML/Markdown typo in `ModularizationLearningJourney.md`

SimonMarquis:


dturner:
Good spot, thanks. Force of habit when writing markdown :)

PR #807, opened 2023-06-18
AdaptiveScaffold

mmoczkowski:
Adaptive Scaffold is a composable that simplifies the adaptation of different screen sizes and form factors for Android applications. This Adaptive Scaffold supports multiple navigation components including bottom bar, navigation rail, and permanent drawer, and automatically switches between them depending on the available screen size. Additionally, it supports a two-pane view with a custom navigator.

```kotlin
fun <T> AdaptiveScaffold(
    navigator: AdaptiveScaffoldNavigator,
    navigationItems: List<T>,
    navigationItemTitle: @Composable (item: T, isSelected: Boolean) -> Unit,
    navigationItemIcon: @Composable (item: T, isSelected: Boolean) -> Unit,
    isItemSelected: @Composable (item: T) -> Boolean,
    onNavigationItemClick: (item: T) -> Unit,
    topBar: @Composable () -> Unit = {},
    snackbarHost: @Composable () -> Unit = {},
    colors: AdaptiveScaffoldNavigationComponentColors = AdaptiveScaffoldNavigationComponentDefaults.colors(),
    contentWindowInsets: WindowInsets = ScaffoldDefaults.contentWindowInsets,
    content: @Composable (padding: PaddingValues) -> Unit,
)
```

Sample usage:

```kotlin
val adaptiveScaffoldNavigator = remember { AdaptiveScaffoldNavigator() }
val navController = rememberNavController(adaptiveScaffoldNavigator)

enum class NavDestination(
    val title: String,
    val icon: ImageVector
) {
    Home(title = "Home", icon = Icons.Default.Home),
    Profile(title = "Profile", icon = Icons.Default.Account),
    Settings(title = "Settings", icon = Icons.Default.Settings)
}

AdaptiveScaffold(
    navigator = adaptiveScaffoldNavigator,
    navigationItems = NavDestination.values().toList(),
    navigationItemTitle = { item, _ -> Text(text = item.title) },
    navigationItemIcon = { item, _ -> Icon(imageVector = item.icon) },
    onNavigationItemClick = { item ->
        val destination = when(item) {
            Home -> "home"
            Profile -> "profile"
            Settings -> "settings"
        }
        navController.navigate(destination)
    }
) { padding ->
    NavHost(
        navController = navController,
        startDestination = "home",
        modifier = Modifier.padding(padding)
    ) {
        composable("home") { ... }
        composable("profile") { ... }

        // This destination will be opened in the Scaffold's details pane
        detailsPane("settings) { ... }
    }
}
```
| Navigation component        | Details Pane Visible                                                                                           | Details Pane Hidden                                                                                           |
|-|-|-|
| Bottom Bar        | |                                 |
| Navigation Rail   | | |
| Permanent Drawer  | | |


app/src/main/java/com/google/samples/apps/nowinandroid/ui/AdaptiveScaffold.kt
```diff
@@ -0,0 +1,216 @@
...
+                                        unselectedTextColor = colors.unselectedTextColor,
+                                    ),
+                                )
+                            }
+
+                        }
+                    },
+                    modifier = Modifier.padding(padding),
+                ) {
+                    content(padding)
```
manuelvicnt, line 144: Because we're using different instances of `Scaffold` depending on the configuration, that means we'll also lose app state when the window size changes. 

To avoid that, should we wrapped the content in `movableContentOf`? I don't think we have a problem with `NavigationDrawerItem` and other components because they're stateless

app/src/main/java/com/google/samples/apps/nowinandroid/ui/AdaptiveScaffold.kt
```diff
@@ -0,0 +1,220 @@
...
+    topBar: @Composable () -> Unit = {},
+    snackbarHost: @Composable () -> Unit = {},
+    colors: AdaptiveScaffoldNavigationComponentColors = AdaptiveScaffoldNavigationComponentDefaults.colors(),
+    contentWindowInsets: WindowInsets = ScaffoldDefaults.contentWindowInsets,
+    content: @Composable (padding: PaddingValues) -> Unit,
+) {
+    val context = LocalContext.current
+    val metrics = WindowMetricsCalculator.getOrCreate().computeCurrentWindowMetrics(context)
+    val widthDp = metrics.bounds.width() / context.resources.displayMetrics.density
+    val movableContent = remember(content) {
```
alexvanyo, line 109: I don't think we want to be using `content` as a key here, since we don't want to wipe away the whole state if the `content` lambda from the caller changes slightly. Although I'm not really sure how `@Composable (padding: PaddingValues) -> Unit` equality works to begin with. If `AppState` changes or something else changes in what we pass to `AdaptiveScaffold`, will that cause `content` to be different?

mmoczkowski: Hey @alexvanyo, using content as a key is a recommendation from our docs:

https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#movableContentOf(kotlin.Function0)

juliewww:
@mmoczkowski Hi, why is it closed? Will use a new solution?

PR #809, opened 2023-06-18
Extract launcher icon tints and provide `-night` alternatives

### LINKED ISSUE #805: [FR]: App icon supporting light and dark themes
### Describe the problem

The app currently only provide a single version of its icon.
Adding support for a different icon when toggling between light and dark theme would be great.

Support on the IDEA project was done here:
- #780

And both icons are also described in the [Now-In-Android-Design-File.pdf](https://github.com/android/nowinandroid/blob/main/docs/Now-In-Android-Design-File.pdf) on page 16.

### Describe the solution

Simply use color references instead of hardcoded values, and override color in `values-night` to invert the colors.

### Additional context

If this is something that you would want to add to the codebase I can prepare a PR.
And this could also be used in
- #800

SimonMarquis:
Closes #805
Not sure which if these are the right dark/light (or the opposite) but it looks fine to me this way.

One question though:
- [ ] Should we also update pre-26 pngs to support dark/light theme or is can we ignore them?

| Light (default) | Dark |
|---|---|
| Screenshot_20230618_132811 | Screenshot_20230618_132720 |
| Screenshot_20230618_132550 | Screenshot_20230618_132755 |


SimonMarquis:
I agree for pre-26 pngs but they are pain to render :/

PR #810, opened 2023-06-18
Remove feature to feature dependencies

SimonMarquis:


app/src/main/java/com/google/samples/apps/nowinandroid/navigation/NiaNavHost.kt
```diff
@@ -41,6 +45,9 @@ fun NiaNavHost(
     appState: NiaAppState,
     onShowSnackbar: suspend (String, String?) -> Boolean,
     modifier: Modifier = Modifier,
+    bookmarksViewModel: BookmarksViewModel = hiltViewModel(),
```
mmoczkowski, line 48: This scopes view models to the activity. It seems like quite a fundamental change. I wonder what kind of effect will it have.

PR #811, opened 2023-06-19
fix: close tag in ModularizationLearningJourney.md

tsumuchan:
There was a display breakdown because the closing tag was missing.

| Before | After |
|--------|--------|
|スクリーンショット 2023-06-20 午前7 07 57|スクリーンショット 2023-06-20 午前7 08 03|


dturner (approved):

PR #812, opened 2023-06-21
Kotlin Coding conventions

Askeri-Muhendis:
Kotlin Class layout
The contents of a class should go in the following order:
- Property declarations and initializer blocks
-Secondary constructors
-Method declarations
-Companion object


mmoczkowski (approved):
Thanks for keeping the code clean

dturner (approved):

PR #817, opened 2023-06-22
Switch to new Activity.enableEdgeToEdge

alexvanyo:
This switches over to the new `Activity.enableEdgeToEdge` call in `androidx.compose`.

This allows a bit of cleanup:
- We no longer have as much `themes.xml` logic for the system bar color, as this is done programmatically
- The system bar color is backported even better across API versions
- We can remove `accompanist/systemuicontroller` as a dependency, since the only usage of it in this project is superceded by `Activity.enableEdgeToEdge`


yaraki (approved):

lihenggui:
There is a small UI change here
Before this change:
image
image


After the change: 
image
image


Previously, the navigation bar was transparent, but after implementing this update, it will overlay a new background. It appears that this layer cannot be removed by setting the `navigationBarStyle` to `SystemBarStyle.auto(android.graphics.Color.TRANSPARENT, android.graphics.Color.TRANSPARENT)`.

yaraki:
> Previously, the navigation bar was transparent, but after implementing this update, it will overlay a new background. It appears that this layer cannot be removed by setting the `navigationBarStyle` to `SystemBarStyle.auto(android.graphics.Color.TRANSPARENT, android.graphics.Color.TRANSPARENT)`.

That's what API29+ does by default for 3-button nav. If you want transparent nav bar even for 3-button nav, you can do `SystemBarStyle.dark(TRANSPARENT)` or `SystemBarStyle.light(TRANSPARENT, TRANSPARENT)`.

lihenggui:
> > Previously, the navigation bar was transparent, but after implementing this update, it will overlay a new background. It appears that this layer cannot be removed by setting the `navigationBarStyle` to `SystemBarStyle.auto(android.graphics.Color.TRANSPARENT, android.graphics.Color.TRANSPARENT)`.
> 
> That's what API29+ does by default for 3-button nav. If you want transparent nav bar even for 3-button nav, you can do `SystemBarStyle.dark(TRANSPARENT)` or `SystemBarStyle.light(TRANSPARENT, TRANSPARENT)`.

Thank you for your answer. : )

dturner (approved):

PR #820, opened 2023-06-27
Added ProfileVerifier to log the app's Baseline Profile Compilation Status

yenerm:
Added ProfileVerifier to log the app's Baseline Profile Compilation Status
To verify profile installation locally, you need to either compile the app  with the speed-profile like so:
         `adb shell cmd package compile -f -m speed-profile com.example.macrobenchmark.target`
Or trigger background dex optimizations manually like so:
        `adb shell pm bg-dexopt-job`

As these run in the background it might take a while for this to complete.


app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -133,12 +138,45 @@ class MainActivity : ComponentActivity() {
...
+         `adb shell cmd package compile -f -m speed-profile com.example.macrobenchmark.target`
+
+        Or trigger background dex optimizations manually like so:
+        `adb shell pm bg-dexopt-job`
+        As these run in the background it might take a while for this to complete.
+
+        To see quick turnaround of the ProfileVerifier, we recommend using `speed-profile`.
+        If you don't do either of these steps, you might only see the profile being enqueued for
+        compilation when running the sample locally.
+         */
```
keyboardsurfer, line 165:
```suggestion
        When delivering through Google Play, the baseline profile is compiled during installation.
        In this case you will see the correct state logged without any further action necessary.
        
        To verify baseline profile installation locally, you need to manually trigger baseline profile installation.
        For immediate compilation, call:
         `adb shell cmd package compile -f -m speed-profile com.example.macrobenchmark.target`

        You can also trigger background optimizations:
        `adb shell pm bg-dexopt-job`
        Both jobs run asynchronously and might take some time complete.

        To see quick turnaround of the ProfileVerifier, we recommend using `speed-profile`.
        If you don't do either of these steps, you might only see the profile status reported as "enqueued for
        compilation" when running the sample locally.
         */
```

keyboardsurfer (approved):

app/src/main/java/com/google/samples/apps/nowinandroid/MainActivity.kt
yenerm: done

PR #824, opened 2023-06-29
Track when deep link notifications are opened

tunjid:
This PR tracks an deep link for a news resource is opened.

### DIFF
--- a/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
+++ b/feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
@@ -131,17 +131,7 @@ class ForYouViewModel @Inject constructor(
         if (newsResourceId == deepLinkedNewsResource.value?.id) {
             savedStateHandle[LINKED_NEWS_RESOURCE_ID] = null
         }
-        analyticsHelper.logEvent(
-            AnalyticsEvent(
-                type = "news_deep_link_opened",
-                extras = listOf(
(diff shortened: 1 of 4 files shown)


dturner (approved):

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
```diff
@@ -127,6 +131,17 @@ class ForYouViewModel @Inject constructor(
         if (newsResourceId == deepLinkedNewsResource.value?.id) {
             savedStateHandle[LINKED_NEWS_RESOURCE_ID] = null
         }
+        analyticsHelper.logEvent(
```
mmoczkowski, line 134: would it be a good idea to encapsulate distinct events in an extension function?

mmoczkowski (approved):
Just a minor suggestion

feature/foryou/src/main/java/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModel.kt
dturner: Good idea. In other modules we added an `AnalyticsExtensions.kt` file which contained `logX` methods specific to that module. 

Example: https://github.com/android/nowinandroid/blob/2a4d0434a5c8a794f86410850e25d66e846eb97f/core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/repository/AnalyticsExtensions.kt#L69

PR #827, opened 2023-06-29
Adds PowerMetric benchmark to measure Light vs Dark theme

yenerm:
Adds PowerMetric benchmark which scrolls through the topics list in Light vs Dark theme

### DIFF
--- a/benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
+++ b/benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
@@ -48,3 +48,17 @@ fun MacrobenchmarkScope.interestsToggleBookmarked() {
     checkable.click()
     device.waitForIdle()
 }
+
+fun MacrobenchmarkScope.setAppTheme(isDark: Boolean) {
+    when (isDark){
+        true -> device.findObject(By.text("Dark")).click()
+        false -> device.findObject(By.text("Light")).click()
+    }
+    device.waitForIdle()
+    device.findObject(By.text("OK")).click()
+
+    // Wait until interests are shown on screen
+    device.wait(Until.hasObject(By.res("niaTopAppBar")), 2_000)
+    val topAppBar = device.findObject(By.res("niaTopAppBar"))
+    topAppBar.wait(Until.hasObject(By.text("Now in Android")), 2_000)
+}
--- a/benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
+++ b/benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
@@ -48,17 +48,3 @@ fun MacrobenchmarkScope.interestsToggleBookmarked() {
     checkable.click()
     device.waitForIdle()
 }
-
-fun MacrobenchmarkScope.setAppTheme(isDark: Boolean) {
-    when (isDark) {
-        true -> device.findObject(By.text("Dark")).click()
-        false -> device.findObject(By.text("Light")).click()
-    }
-    device.waitForIdle()
-    device.findObject(By.text("OK")).click()
-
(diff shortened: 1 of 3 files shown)


benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/ScrollTopicListPowerMetricsBenchmark.kt
```diff
@@ -0,0 +1,82 @@
...
+            metrics = listOf(FrameTimingMetric(), PowerMetric(PowerMetric.Energy(categories))),
+            compilationMode = compilationMode,
+            iterations = 2,
+            startupMode = StartupMode.WARM,
+            setupBlock = {
+                // Start the app
+                pressHome()
+                startActivityAndWait()
+                allowNotifications()
+                // Navigate to interests screen
```
dturner, line 70: The following code doesn't appear to navigate to the interests screen, it just opens the app and clicks the Settings button. Either the comment is wrong or the code needs to be updated.

benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
```diff
@@ -48,3 +48,17 @@ fun MacrobenchmarkScope.interestsToggleBookmarked() {
     checkable.click()
     device.waitForIdle()
 }
+
+fun MacrobenchmarkScope.setAppTheme(isDark: Boolean) {
```
dturner, line 52: I'm not sure why this function is inside `InterestsActions`. The Settings button can be found on every screen in the app inside the top app bar. It might make more sense to have this in a `ForYouActions` file (because ForYou is the first screen in the app), or just `AppActions`.

benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
```diff
@@ -48,3 +48,17 @@ fun MacrobenchmarkScope.interestsToggleBookmarked() {
...
+    when (isDark) {
+        true -> device.findObject(By.text("Dark")).click()
+        false -> device.findObject(By.text("Light")).click()
+    }
+    device.waitForIdle()
+    device.findObject(By.text("OK")).click()
+
+    // Wait until interests are shown on screen
```
dturner, line 60: It looks like this code is waiting for the top app bar to be shown, rather than any interests/topics.

benchmarks/src/main/java/com/google/samples/apps/nowinandroid/interests/ScrollTopicListPowerMetricsBenchmark.kt
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
dturner, line 2: nit: new files should have the current year for copyright

dturner (changes requested):
I'm not sure if this is relevant but the way the app looks is defined by: 

- the brand which the user chooses, this can be default or Android green
- if the default brand is chosen then this will be pink on API < 31 or will use dynamic color on API 31+ if the user has selected that option  
- lastly, once the brand has been chosen, dark mode determines the actual colours used. 

I'm guessing you just want to test light vs dark mode energy consumption but there might be some variance with the brand configurations as well.

yenerm:
Thanks Don! I addressed the comments. Also good points on variance with the brand configurations. I think the benchmark still makes sense as a sample in NiA. 
Eventually (no the first priority) I am planning to add similar sample to performance samples but it doesn't currently have a theme for dark mode. Let me know wdyt?

dturner (approved):

dturner:
Totally fine for this to be in this project. Implementing the theme wasn't trivial so probably makes sense to benchmark it here, rather than introducing that complexity into a simpler sample just to test it.

PR #834, opened 2023-07-13
Rename LazyListScope.TopicBody LazyListScope.topicBody

takahirom:
It appears that other extensions for LazyListScope start with a lowercase letter, but TopicBody starts with an uppercase letter. Therefore, I have changed it to start with a lowercase letter.


hoc081098:
extensions which start with lowercase are not Composables.
Composables emitting UI elements should start with uppercase letter

hoc081098:
It is better to rename `userNewsResourceCard` to `UserNewsResourceCard`

alexvanyo (approved):
Thanks for cleaning this up to be consistent!

PR #835, opened 2023-07-13
Call getNewsResourceIds instead of getNewsResources

blackbracken:
By calling `getNewsResources`, we were unintentionally fetching extra entities. To resolve this issue, I defined a new method (from `getNewsResources`) called `getNewsResourceIds` that retrieves only the IDs.

### DIFF
--- a/core/data/src/test/java/com/google/samples/apps/nowinandroid/core/data/testdoubles/TestNewsResourceDao.kt
+++ b/core/data/src/test/java/com/google/samples/apps/nowinandroid/core/data/testdoubles/TestNewsResourceDao.kt
@@ -67,6 +67,33 @@ class TestNewsResourceDao : NewsResourceDao {
                 result
             }
 
+    override fun getNewsResourceIds(
+        useFilterTopicIds: Boolean,
+        filterTopicIds: Set<String>,
+        useFilterNewsIds: Boolean,
+        filterNewsIds: Set<String>,
+    ): Flow<List<String>> =
+        entitiesStateFlow
+            .map { newsResourceEntities ->
+                newsResourceEntities.map { entity ->
+                    entity.asPopulatedNewsResource(topicCrossReferences)
+                }
+            }
+            .map { resources ->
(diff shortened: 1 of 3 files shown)


core/data/src/test/java/com/google/samples/apps/nowinandroid/core/data/testdoubles/TestNewsResourceDao.kt
```diff
@@ -67,6 +67,33 @@ class TestNewsResourceDao : NewsResourceDao {
...
+                        resource.topics.any { it.id in filterTopicIds }
+                    }
+                }
+                if (useFilterNewsIds) {
+                    result = result.filter { resource ->
+                        resource.entity.id in filterNewsIds
+                    }
+                }
+                result.map { it.entity.id }
+            }
```
blackbracken, line 95: Although we could have directly called the existing implementation of `getNewsResources`, I decided to avoid doing so to prevent excessive generalization.

tunjid: I think its fine to here.

tunjid (changes requested):
Thanks for doing this! Just a few changes requested.

tunjid:
Also seems you have a few small formatting issues.

core/database/src/main/java/com/google/samples/apps/nowinandroid/core/database/dao/NewsResourceDao.kt
```diff
@@ -65,6 +65,33 @@ interface NewsResourceDao {
         filterNewsIds: Set<String> = emptySet(),
     ): Flow<List<PopulatedNewsResource>>
 
+    @Transaction
```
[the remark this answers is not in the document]
blackbracken: Added method overview as kdoc.

tunjid (approved):

PR #852, opened 2023-07-23
Improved Result class for less need of boilerplate code

mobilekosmos:


jerryOkafor (commented):
Looks good to me.

core/common/src/main/java/com/google/samples/apps/nowinandroid/core/result/Result.kt
```diff
@@ -27,11 +27,20 @@ sealed interface Result<out T> {
     object Loading : Result<Nothing>
 }
 
-fun <T> Flow<T>.asResult(): Flow<Result<T>> {
+fun <T, R> Flow<T>.mapToResultUiState(
```
dturner, line 30: Not really a fan of this as UI state isn't something which the Result class should care about.

core/common/src/main/java/com/google/samples/apps/nowinandroid/core/result/Result.kt
```diff
@@ -27,11 +27,20 @@ sealed interface Result<out T> {
...
+    onSuccess: (T) -> R,
+    onLoading: () -> R,
+    onError: (Throwable?) -> R,
```
dturner, line 33: The added complexity of having 3 lambdas as parameters, plus an additional generic type, is difficult to justify, even if the call site has less boilerplate.

dturner:
Thanks for submitting this. It's an interesting idea, however, I think in this case it's probably not worth merging. Feel free to convince me otherwise :)

core/common/src/main/java/com/google/samples/apps/nowinandroid/core/result/Result.kt
mobilekosmos: What's your reasoning behind prioritizing a "hidden" code (for lack of a better adjective) over a simplified code on the calling site? 2 Million Developers should write and read boilerplate code because otherwise, the code behind would be more complex?

dturner: Sorry, I should've added this comment at the call site. I agree that if the call site is simpler then it doesn't matter if there's extra complexity behind the scenes. 
 
However, my point was: the call site isn't simpler. Consider the two options: 

Before: 
```
getSearchContentsUseCase(query).asResult().map {
    when (it) {
        is Result.Success -> {
            SearchResultUiState.Success(
                topics = it.data.topics,
                newsResources = it.data.newsResources,
             )
         }
         is Result.Loading -> { SearchResultUiState.Loading }
         is Result.Error -> { SearchResultUiState.LoadFailed }
     }
}
```

After:
```
getSearchContentsUseCase(query)
    .mapToResultUiState(
        onSuccess = {
            SearchResultUiState.Success(
                topics = it.topics,
                newsresources = it.newsResources,
            )
        },
        onLoading = { SearchResultUiState.Loading },
        onError = { SearchResultUiState.LoadFailed },
    )
```

I find the "Before" easier to understand than the "After". A type comparison `is something` is easier to read than `onSomething = `. The `on*` parameters add an extra level of abstraction (and therefore cognitive burden) between the Result types and methods.  

Also the "Before" approach is more flexible as it allows the same state for multiple Result types (e.g. `is Result.Success, is Result.Loading -> { ... }`).

mmoczkowski (changes requested):
IMO this is a wrapper around a simple map call. I agree with Don that it's quite unnecessary.

mobilekosmos:
I see the positive looks of when+is in other circumstances, but with that solution, you need 3 calls instead of just one and, overall many more characters to write each time. You could also change the function names like whenResultIsSuccess, whenResultIsLoading, etc. In the end, it looks like these would be a taste thing, but if you try to measure the difference, one solution would need less than the other.

image

PR #854, opened 2023-07-25
Update AGP 8.1.0

SimonMarquis:
And remove AGP lint experimental workaround to support Kotlin 1.9 metadata.


alexvanyo (approved):
Thanks for the update!

PR #861, opened 2023-07-27
Fix for Retrofit R8 issue, bump version to 0.1.2

dturner:
This fixes an issue where data would not load from the backend because of an exception thrown by retrofit. Users would see a loading spinner for a short time but no new data would be loaded from the server. The following would be seen in logcat: 

```
2023-07-26 15:59:15.186 12572-12678 suspendRunCatching      com...gle.samples.apps.nowinandroid  I  Failed to evaluate a suspendRunCatchingBlock. Returning failure Result
                                                                                                    java.lang.ClassCastException: java.lang.Class cannot be cast to java.lang.reflect.ParameterizedType
```

Cause: https://github.com/square/retrofit/issues/3751
Solution: Add proguard rules to stop Retrofit classes from being removed

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -29,8 +29,8 @@ plugins {
 android {
     defaultConfig {
         applicationId = "com.google.samples.apps.nowinandroid"
-        versionCode = 7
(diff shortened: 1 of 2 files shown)


SimonMarquis:
For reference
- https://github.com/square/retrofit/pull/3910

PR #862, opened 2023-07-27
Remove hardcoded topic names in `NavigationTest.kt`

SimonMarquis:
This test now uses ~`NiaNetworkDataSource`~ `TopicsRepository` (provided by Hilt) to get a random topic.
This also removes potential flakiness by really checking the absence of the topic list instead of searching for text (that could match in the opened topic page).

### DIFF
--- a/app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
+++ b/app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
@@ -33,7 +33,7 @@ import androidx.test.espresso.Espresso
 import androidx.test.espresso.NoActivityResumedException
 import com.google.samples.apps.nowinandroid.MainActivity
 import com.google.samples.apps.nowinandroid.R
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
```diff
@@ -251,11 +263,14 @@ class NavigationTest {
...
     @Test
-    fun navigationBar_multipleBackStackInterests() {
+    fun navigationBar_multipleBackStackInterests() = runTest {
         composeTestRule.apply {
             onNodeWithText(interests).performClick()
-            // TODO: Grab string from fake data
-            onNodeWithText("Android Studio & Tools").performClick()
+
+            // Select a random topic
+            val topic = datasource.getTopics().random().name
```
dturner, line 271: It's not good practice to use anything random in tests as it can make them flaky. Select the last element instead.

dturner (approved):

PR #868, opened 2023-07-28
Handle topic clicks from all destinations

### LINKED ISSUE #865: [Bug] Tapping on a topic chip doesn't open the topic detail screen
Version tested: 0.1.2

Steps to repro: 

- Open the For You screen
- Tap on any topic chip

Expected result

- Topic detail screen for that topic is shown

Actual result

- Nothing happens, although the tap animation is shown

Other notes
If you bookmark a new resource, then open the Saved screen, then tap on the topic chip there, it does work as expected.

SimonMarquis:
Closes #865

### DIFF
--- a/app/src/main/java/com/google/samples/apps/nowinandroid/navigation/NiaNavHost.kt
+++ b/app/src/main/java/com/google/samples/apps/nowinandroid/navigation/NiaNavHost.kt
@@ -49,8 +49,7 @@ fun NiaNavHost(
         startDestination = startDestination,
         modifier = modifier,
     ) {
-        // TODO: handle topic clicks from each top level destination
-        forYouScreen(onTopicClick = {})
+        forYouScreen(onTopicClick = navController::navigateToTopic)
         bookmarksScreen(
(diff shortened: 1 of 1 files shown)


SimonMarquis:
There are probably more instances of this callback to be fixed 👀

mmoczkowski (approved):
LGTM. Now sure why those callbacks were left empty. Thank you for your contribution!

lelelongwang:
I found two bugs:

- bug 1: When you execute onTopicClick in foryouScreen to enter the topicScreen, and then click foryouItem on the navigationbaritem, you cannot jump to foryouScreen.

> Solution 1: It can be consistent with "Interests", when jumping from foryouScreen to topicScreen, "Foryou" is still selected

> Solution 2: After jumping from foryouScreen to topicScreen, if "Interests" is selected, clicking "Foryou" should enter foryouScreen

- bug 2: When I execute onTopicClick again in topicScreen, I do not enter the new topicScreen according to the new topic ID, which means that in the topicScreen interface, the onTopicClick interface does not change

PR #869, opened 2023-07-28
Migrate to baseline profile plugin

SimonMarquis:


SimonMarquis:
ℹ️ Running the default `gradlew :app:generateProdReleaseBaselineProfile` task leads run a Gradle error where it says the target app is not installed on the device 🤯 
Unfortunately I did not saved the stacktrace.
But running `gradlew :app:installProdRelease :app:generateProdReleaseBaselineProfile` seems to have fixed it.

SimonMarquis:
@keyboardsurfer can you share with us what is the expected run configuration for this project?

**BuildType/Flavor:**
- `ProdRelease`: unable to reproduce since production API keys are not provided here (although we can extract them from the PlayStore apk) and therefore the listings would be empty
- `DemoRelease`: not necessarily representative since it does not go through the whole network stack (data is fetched from a local JSON file).

**Device:**
- `pixel6api31aosp`: anyone can "reproduce" since it's a Gradle Managed Device, but it comes with its own downsides that can be ignored with `testInstrumentationRunnerArguments["androidx.benchmark.suppressErrors"] = "EMULATOR"`:
  <details><summary><b>Click to show warning message</b></summary>

  ```
  /**
   * ERROR: Running on Emulator
   *     Benchmark is running on an emulator, which is not representative of
   *     real user devices. Use a physical device to benchmark. Emulator
   *     benchmark improvements might not carry over to a real user's
   *     experience (or even regress real device performance).
   *
   * While you can suppress these errors (turning them into warnings)
   * PLEASE NOTE THAT EACH SUPPRESSED ERROR COMPROMISES ACCURACY
   */
  ```

  </details>
- Custom rooted device: not able to reproduce (for obvious reasons)

---

It would be really great if it could be reproduced on CI builds before releases.

SimonMarquis:
Although the `BaselineProfileGenerator` test succeeded, I got the following error:

```

* What went wrong:
Execution failed for task ':benchmarks:collectDemoNonMinifiedReleaseBaselineProfile'.
> kotlin.io.NoSuchFileException: nowinandroid\benchmarks\build\intermediates\managed_device_android_test_additional_output\demoNonMinifiedRelease\pixel6api31aospDemoNonMinifiedReleaseAndroidTest\BaselineProfileGenerator_generate-baseline-prof-2023-07-28-19-46-26.txt: The source file doesn't exist.
```

Could this be related to running baseline profile tests on emulator?

keyboardsurfer:
Re #issuecomment-1656245386: This means the baseline profile output file wasn't found after the test was created. Can you try again with `saveInSource` disabled? If that works we might have to [file a bug](https://issuetracker.google.com/issues/new?component=1071684&template=1592279) against the plugin.

You can create a Baseline Profile on an emulator and should not see any errors along the way.
Measuring its effectiveness requires a physical device. The "running on emulator" error above should only pop up when you want to measure performance.

Paging @dturner for advice on the variant to build.

SimonMarquis:
I guess you made huge progress on #880 making this PR a bit outdated 🙈

keyboardsurfer:

PR #872, opened 2023-07-29
Convert 'sealed' sub-object to 'data object'

SimonMarquis:


alexvanyo (approved):
Yay `data object`s!

dturner:
Just in case anyone is interested in why this is a good thing: https://kotlinlang.org/docs/whatsnew1720.html#improved-string-representations-for-singletons-and-sealed-class-hierarchies-with-data-objects

PR #874, opened 2023-07-29
Replace StringDecoder injection with simpler API

SimonMarquis:
The initial reason for this `StringDecoder` was to make sure the AndroidX navigation route would not use incompatible chars in the resulting String.

For that, `Uri.encode(String)` was used, but this is an Android API (`android.net.Uri`). Unit tests would fail because Uri is not mocked (since we are not using Robolectric). Therefore, a `StringDecoder` interface was introduced with 2 implementations: `UriDecoder` and `FakeStringDecoder`, Hilt modules were added, etc. FWIW, the naming of the API was misleading: `StringDecoder.decodeString(encodedString: String)` does not inform how encoded the input is, so `UriDecoder` could "decode" something that was not necessarily uri "encoded".

The solution to this problem was to simply use regular Java APIs:
- `URLDecoder.decode(urlEncodedTopicId, "UTF-8")`
- `URLEncoder.encode(topicId, "UTF-8")`


feature/topic/src/main/java/com/google/samples/apps/nowinandroid/feature/topic/navigation/TopicNavigation.kt
```diff
@@ -16,27 +16,27 @@
...
 internal class TopicArgs(val topicId: String) {
-    constructor(savedStateHandle: SavedStateHandle, stringDecoder: StringDecoder) :
-        this(stringDecoder.decodeString(checkNotNull(savedStateHandle[topicIdArg])))
+    constructor(savedStateHandle: SavedStateHandle) :
+        this(URLDecoder.decode(checkNotNull(savedStateHandle[topicIdArg]), "UTF-8"))
 }
 
 fun NavController.navigateToTopic(topicId: String) {
-    val encodedId = Uri.encode(topicId)
+    val encodedId = URLEncoder.encode(topicId, "UTF-8")
```
dturner, line 39: nit: please move "UTF-8" to a constant

dturner (approved):
This is great, much simpler. One very minor nit.

feature/topic/src/main/java/com/google/samples/apps/nowinandroid/feature/topic/navigation/TopicNavigation.kt
SimonMarquis: Should we really extract these into constants?
Do you prefer a 

```kotlin
private const val UTF_8 = "UTF-8"

URLEncoder.encode(topicId, UTF_8)
```

or directly use Kotlin's already defined `Charsets.UTF_8`?

```kotlin
URLEncoder.encode(topicId, UTF_8.name())
```

dturner: The constant name should express the meaning of its value, rather than the actual value, so maybe: 

```
private const val URL_CHARACTER_ENCODING  = UTF_8.name()
```

SimonMarquis: Well, if we go the route of expressiveness, I think there is nothing that can beat `URLEncoder.encode(topicId, "UTF-8")`, but OK I'll update that code :)

SimonMarquis: Although, since `.name()` is a function, the `val` can no longer be a `const`.

dturner (approved):

PR #876, opened 2023-08-01
Adds Screenshot testing with Roborazzi

JoseAlcerreca:


core/testing/src/main/java/com/google/samples/apps/nowinandroid/core/testing/util/ScreenshotHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+fun <A : ComponentActivity> AndroidComposeTestRule<ActivityScenarioRule<A>, A>.captureMultiDevice(
+    screenshotName: String,
+    body: @Composable () -> Unit,
+) {
+    listOf(
+        "phone" to "spec:shape=Normal,width=640,height=360,unit=dp,dpi=480",
+        "foldable" to "spec:shape=Normal,width=673,height=841,unit=dp,dpi=480",
+        "tablet" to "spec:shape=Normal,width=1280,height=800,unit=dp,dpi=480",
+    ).forEach {
+        this.captureForDevice(it.first, it.second, screenshotName, body)
```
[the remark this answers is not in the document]
JoseAlcerreca: Yeah we'll do it properly when it's ready

core/testing/src/main/java/com/google/samples/apps/nowinandroid/core/testing/util/ScreenshotHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+    this.onRoot()
+        .captureRoboImage(
+            "src/test/screenshots/${screenshotName}_$deviceName.png",
+            roborazziOptions = roborazziOptions,
+        )
+}
+
+private fun extractSpecs(deviceSpec: String): List<Int> {
```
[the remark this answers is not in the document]
JoseAlcerreca: The idea is to have a clear migration path between this and the preview-based testing tool and to have a story that uses previews for manual testing and roborazzi for automated testing. By using the same spec we make that easier. The hope is to remove this once we can use the preview-based testing tool.

build.gradle.kts
```diff
@@ -40,4 +40,5 @@ plugins {
     alias(libs.plugins.hilt) apply false
     alias(libs.plugins.ksp) apply false
     alias(libs.plugins.secrets) apply false
-}
\ No newline at end of file
+    id("io.github.takahirom.roborazzi") version "1.2.0-alpha-1" apply false
```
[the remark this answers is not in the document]
JoseAlcerreca: Woops! not a nit at all :D

gradle/libs.versions.toml
```diff
@@ -125,6 +127,8 @@ protobuf-kotlin-lite = { group = "com.google.protobuf", name = "protobuf-kotlin-
 protobuf-protoc = { group = "com.google.protobuf", name = "protoc", version.ref = "protobuf" }
 retrofit-core = { group = "com.squareup.retrofit2", name = "retrofit", version.ref = "retrofit" }
 retrofit-kotlin-serialization = { group = "com.jakewharton.retrofit", name = "retrofit2-kotlinx-serialization-converter", version.ref = "retrofitKotlinxSerializationJson" }
+robolectric = { module = "org.robolectric:robolectric", version.ref = "robolectric" }
+roborazzi = { module = "io.github.takahirom.roborazzi:roborazzi", version.ref = "roborazzi" }
```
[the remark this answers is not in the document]
JoseAlcerreca: Done

core/testing/src/main/java/com/google/samples/apps/nowinandroid/core/testing/util/ScreenshotHelper.kt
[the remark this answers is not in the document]
alexvanyo: SGTM. Only nit I have is that we aren't applying everything from the device spec to configure Robolectric. For instance, we aren't overriding Roboelctric's default `160dpi` value with the `480dpi` value from the device specs.

alexvanyo (approved):

SimonMarquis:
Hi @JoseAlcerreca, awesome work!
I was just wondering if you can comment on why you chose Roborazzi over Paparazzi?
Adding the Robolectric to this codebase has been delayed for a long time.
And to me, it sometimes unintentionally pushes to write unit tests at the wrong layer of abstraction.
Also, I don't see in these tests any access to the Android framework (which would be the main reason to use Roborazzi over Paparazzi).

.github/workflows/Build.yaml
```diff
@@ -13,7 +13,7 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 90
```
SimonMarquis, line 16: Why do we need to increase the timeout of this job if we also removed the `testDemoDebug testProdDebug` tasks?

.github/workflows/Build.yaml
```diff
@@ -59,6 +56,65 @@ jobs:
           name: lint-reports
           path: '**/build/reports/lint-results-*.html'
 
+  test:
```
SimonMarquis, line 59: Shouldn't we also add `needs: build`?

.github/workflows/Build.yaml
```diff
@@ -59,6 +56,65 @@ jobs:
...
+
+      - name: Run all local screenshot tests (Roborazzi)
+        id: screenshotsverify
+        continue-on-error: true
+        run: ./gradlew verifyRoborazziDemoDebug
+
+      - name: Prevent pushing new screenshots if this is a fork
+        id: checkfork
+        continue-on-error: false
+        if: steps.screenshotsverify.outcome == 'failure' && github.event.pull_request.head.repo.full_name != github.repository
```
SimonMarquis, line 94: nit: there is a `github.event.pull_request.head.repo.fork` boolean value

```suggestion
        if: steps.screenshotsverify.outcome == 'failure' && github.event.pull_request.head.repo.fork
```

.github/workflows/Build.yaml
```diff
@@ -59,6 +56,65 @@ jobs:
...
+        id: screenshotsrecord
+        if: steps.screenshotsverify.outcome == 'failure' && github.event_name == 'pull_request'
+        run: |
+          ./gradlew recordRoborazziDemoDebug
+
+      - name: Push new screenshots if available
+        uses: stefanzweifel/git-auto-commit-action@v4
+        if: steps.screenshotsrecord.outcome == 'success'
+        with:
+          file_pattern: '*/*.png'
```
SimonMarquis, line 109: nit: is there a more precise path for generated screenshots?
Something like `*/screenshots/*.png`

SimonMarquis:
FWIW, Windows devs won't be able to run the whole test suite (at the moment) because of:
- https://github.com/robolectric/robolectric/issues/8312

app/src/testDemo/java/com/google/samples/apps/nowinandroid/ui/NiaAppScreenSizesScreenshotTests.kt
```diff
@@ -0,0 +1,233 @@
...
+            InstrumentationRegistry.getInstrumentation().context,
+            config,
+        )
+
+        hiltRule.inject()
+
+        // Configure user data
+        runBlocking {
+            userDataRepository.setShouldHideOnboarding(true)
+            userDataRepository.setFollowedTopicIds(setOf(topicsTestData.first().id))
```
SimonMarquis, line 119: Since we are using Hilt, shouldn't we use an injected `TopicsRepository` instead of accessing fake test data?

JoseAlcerreca:
> nt on why you chose Roborazzi over Paparazzi?

Roborazzi uses Robolectric Native Graphics which is a supported API. Currently, paparazzi uses LayoutLib in an unsupported way.

> Adding the Robolectric to this codebase has been delayed for a long time.
> And to me, it sometimes unintentionally pushes to write unit tests at the wrong layer of abstraction.

We're not changing unit tests or the test strategy with this change.

> Also, I don't see in these tests any access to the Android framework (which would be the main reason to use Roborazzi over Paparazzi).

It doesn't access the framework but it uses RNG.

JoseAlcerreca:
> FWIW, Windows devs won't be able to run the whole test suite (at the moment) because of:
> 
> * [Robolectric NATIVE GraphicsMode fails when running on Windows machine robolectric/robolectric#8312](https://github.com/robolectric/robolectric/issues/8312)

Yeah, known issue. Running `./gradlew test` shouldn't be affected by that and we're taking the screenshots on CI, so it's independent of the developing platform.

.github/workflows/Build.yaml
JoseAlcerreca: ack

.github/workflows/Build.yaml
JoseAlcerreca: No, that would delay the execution of this step and we want quick feedback over lower use of resources.

.github/workflows/Build.yaml
JoseAlcerreca: Dealing with `file_pattern` is a pain that requires multiple tries every time :) so I'd rather keep it simple. Also, only pngs should be untracked at this point so this is unnecessary really.

core/testing/src/main/java/com/google/samples/apps/nowinandroid/core/testing/util/ScreenshotHelper.kt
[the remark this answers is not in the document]
JoseAlcerreca: Refactored everything to use Robolectric's size/width/dpi qualifiers instead of the TestHarness for the `foryou` tests.

app/src/testDemo/java/com/google/samples/apps/nowinandroid/ui/NiaAppScreenSizesScreenshotTests.kt
JoseAlcerreca: Good catch! I was being lazy here. Now I'm accessing the topic ID correctly.

.github/workflows/Build.yaml
JoseAlcerreca: Reverted this. Turns out it doesn't do the same thing. https://stackoverflow.com/questions/70775445/determining-the-origin-repo-of-pull-request-in-workflow-yaml#comment125127165_70776678

SimonMarquis: Oh, interesting 👀

JoseAlcerreca:
Currently blocked by #909

mmoczkowski (changes requested):
scr

Screenshot 2023-08-21 at 17 06 27

`NewsResourceCard.kt:246`

Reference screenshots on my machine had a different date (off by one day). The dates are mapped to local time zone in the UI layer. This makes the tests non-deterministic.

JoseAlcerreca:
Good catch @mmoczkowski, forced UTC timezone in screenshot tests.

mmoczkowski (approved):
LGTM

alexvanyo (approved):
Looks great! 🖼️

Issue #867, opened 2023-07-28
[Bug]: Topic screen remains displayed instead of bookmarks screen


SimonMarquis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

Steps to reproduce:

- Bookmark a news
- Open the bookmarks tab
- Click on the corresponding topic
- The interest screen open
- Open the bookmarks tab 🐛 

At this point, opening the bookmarks tab always return to the selected topic instead of poping-up the backstack.  

https://github.com/android/nowinandroid/assets/1921278/2fd45734-1438-4ce2-ac65-a4220f9e9d59

(tested on versions 0.1.1 and 0.1.2)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adityaSharma6356:
Not sure what the actual problem was but the NavController is unable to navigate to destinations already in the back stack.
The same bug can be seen if we go to interests screen through search screen and makes it impossible to navigate back through navigation bar but we still can press back button and go back.

Solution: This can be fixed if we pop all the way back to starting destination before navigating and restore the state too if the user goes to a previously selected item.

I would like to create a PR if no one is working on this :D

tunjid:
@adityaSharma6356

I believe this is already done using `navigateToTopLevelDestination`. When I looked into it, the cause of the bug was:

* Tap on the bookmarks tab
* Enter the bookmarks nav graph
* Tap on a topic
* Navigate to the interests nav graph

The issue therefore is the logic in `NiaBottomBar`, specifically the code to check if a tab is selected:

```kotlin
val selected = currentDestination.isTopLevelDestinationInHierarchy(destination)
```

The navigation isn't actually stuck, the bottom bar just indicates the wrong thing. You can verify this by:

* Bookmarking an item
* Navigating to the bookmarks tab (Bookmarks tab is highlighted)
* Tapping on a topic
* Topic is shown (Interests tab is highlighted)
* Tapping on the already highlighted interests tab
* Topics are shown.
* Tap on a topic that is different from the topic that was first clicked in bookmarks.
* News for that topic are shown.
* Tap on the bookmarks tab even though it is not highlighted.
* You will be taken to news for that topic although the interests tab is still highlighted.
* Tap on the interests tab, notice that news for the second topic is highlighted.
* Alternate between tapping the bookmarks and interests tab, not the tabs are independent top levels.

[...]

PR #887, opened 2023-08-11
Use the back stack to determine nav top level ownership

tunjid:
Addresses #867.


gradle/libs.versions.toml
```diff
@@ -16,7 +16,7 @@ androidxHiltNavigationCompose = "1.0.0"
 androidxLifecycle = "2.6.1"
 androidxMacroBenchmark = "1.2.0-alpha16"
 androidxMetrics = "1.0.0-alpha03"
-androidxNavigation = "2.5.3"
+androidxNavigation = "2.6.0"
```
alexvanyo, line 19: Any reason not to bump to latest stable `2.7.0`?

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
```diff
@@ -97,6 +105,37 @@ class NiaAppState(
             else -> null
         }
 
+    val backStackEntries: List<NavBackStackEntry>
+        @Composable get() {
+            // TODO: Read backStack directly from the navController when b/295553995 is resolved.
```
alexvanyo, line 110: Nit: Can this be the public issue link? https://issuetracker.google.com/issues/295553995

alexvanyo (commented):
Two nits, and can we add a regression test verifying that #867 is fixed?

tunjid:
Will need to rework this, the better solution is to track the last tab tapped to and use that to drive the indicator.

PR #857, opened 2023-07-25
Update kotlinx.coroutines to version 1.7.3

SimonMarquis:
Changelog:
- https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.7.0
- https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.7.1
- https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.7.2
- https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.7.3

PR #890, opened 2023-08-11
Update turbine 1.0.0

SimonMarquis:
Changelog: https://github.com/cashapp/turbine/releases/tag/1.0.0


SimonMarquis:
This PR needs to be applied on top of the KotlinX coroutines update (`org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.x`)

- #857

lihenggui:
https://github.com/android/nowinandroid/blob/main/gradle/libs.versions.toml
Turbine is already updated to the latest version (1.0.0) now.

PR #899, opened 2023-08-14
Create `TestMethodDetector` Lint to detect common naming patterns we want to avoid

### LINKED ISSUE #345: [Bug]: Excessive underscores in test function names

### What happened?

For example, in https://github.com/android/nowinandroid/blob/607c24e7f7399942e278af663ea4ad350e5bbc3a/core/data/src/test/java/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstUserDataRepositoryTest.kt

Check all codebase and use maximum of 2 underscores per test in the format: `given_when_then`

### Relevant logcat output

_No response_

### Code of Conduct

SimonMarquis:
This will need to be applied on top of this PR to run correctly on all modules:
- #759

Closes #345

image
image
image


mmoczkowski (commented):
Hey Simon, can you solve the conflicts please? Thanks!

SimonMarquis:
Hopefully these 3 new Lint checks makes sense for this repository.
And if they do, we can proceed to fix the corresponding warnings 🤓.

SimonMarquis:
🛎️ Ping

SimonMarquis:
🛎️ Ping 🥺

lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodDetector.kt
```diff
@@ -0,0 +1,160 @@
...
+import java.util.EnumSet
+import kotlin.io.path.Path
+
+/**
+ * A detector that checks for common patterns in naming the test methods:
+ * - [detectPrefix] removes unnecessary "test" prefix in all unit test.
+ * - [detectUnderscore] removes underscores in JVM unit test (and add backticks if necessary).
+ * - [detectFormat] Checks the `given_when_then` format of Android instrumented tests (backticks are not supported).
+ */
+class TestMethodDetector : Detector(), SourceCodeScanner {
```
dturner, line 43:
```suggestion
class TestMethodNameDetector : Detector(), SourceCodeScanner {
```

dturner (approved):
This is really cool. Thanks for adding, and sorry for the delay in reviewing.

dturner (changes requested):
Was a bit quick to approve there. Looking a bit more closely, the `UNDERSCORE` check isn't valid. 

We want tests follow this format: `given_when_then` where `given`, `when` and `then` are in camel case e.g. `uiState_whenFollowedTopicsAreLoading_thenShowLoading`. 

What we don't want is for underscores to be used to replace spaces between words e.g. `network_news_resource_can_be_mapped_to_news_resource_entity`

or for no underscores to be used e.g. `deepLinkedNewsResourceIsFetchedAndResetAfterViewing`

I'm not sure if it's possible to create lint checks for this. Or whether it'd be simpler to just rename the existing tests which don't follow the format.

SimonMarquis:
We can definitely enforce this through Lint checks. We simply need to find the right regex.
But from the examples you give, I'm not sure what is expected. I can see the words `when` and `then` but no `given`.

So what you actually want is something like: `[^\W_]+?_when[^\W_]+?_then[^\W_]+?`?

dturner:
Looking through some of the test, the clearest test names seem to be those that follow these formats: 

`when_then` e.g. `whenFilteredByFollowedTopics_matchingNewsResourcesAreReturned`
`given_when_then` e.g `oneBookmark_whenRemoving_removesFromFeed`

In which case the format should be: 

- one underscore minimum
- two underscores maximum
- words before and after each underscore

My regex isn't great but I think it's: `[\w]([_\w])([_\w])?` 

Here's some methods which will require a rename: 

`stateIsInitiallyLoading` => `whenCreated_stateIsLoading`
`stateIsLoadingWhenFollowedTopicsAreLoading` => `whenFollowedTopicsAreLoading_stateIsLoading`

I haven't had a chance to look through all the tests to see whether there are other cases which don't fit this format. Feedback or other suggestions welcome.

SimonMarquis:
The detector has been updated to conform to your specs: `[^\W_]+(_[^\W_]+){1,2}` and matches the `when_then` or `given_when_then` format.
Tests have bee updated as well.

lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodNameDetector.kt
```diff
@@ -0,0 +1,160 @@
...
+            priority = 5,
+            severity = WARNING,
+            implementation = Implementation(
+                TestMethodNameDetector::class.java,
+                EnumSet.of(JAVA_FILE, TEST_SOURCES),
+            ),
+        )
+
+        @JvmField
+        val UNDERSCORE: Issue = issue(
```
dturner: The `UNDERSCORE` check is incompatible with the `FORMAT` check since the method name must contain 1-2 underscores. Suggest removing the `UNDERSCORE` check entirely.

lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/NiaIssueRegistry.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 The Android Open Source Project
+ * Copyright 2023 The Android Open Source Project
```
dturner, line 2: nit: Copyright dates don't need to be updated. No need to change it back but might save you from some extra work for future changes.

lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodNameDetector.kt
dturner: I submitted this change already.

SimonMarquis: This was on purpose, the `UNDERSCORE` only applies to non-Android tests, where we should use backticks instead of undescores. 🙈

SimonMarquis: @dturner Unless you really want to force this ugly snake_case format on regular jvm unit tests.
To me, the underscores workaround was only there for Android instrumented tests that do not support (for whatever reason) spaces in their names, and to make these test names more consistent and readable.

dturner: I see, sorry, the intention wasn't clear from the PR description. I'd prefer to have a single convention, even if it's ugly.

SimonMarquis: Ok

lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/NiaIssueRegistry.kt
SimonMarquis: This was automatically update by my IDE I suppose...

```suggestion
 * Copyright 2022 The Android Open Source Project
```

dturner (approved):
Thanks for bearing with me on this one. I've never written a lint check so it took me a while to figure out whether this was the correct approach.

SimonMarquis:
No worries.
Seems like this issue still happens here:
- https://github.com/android/nowinandroid/issues/993

PR #907, opened 2023-08-17
Replace identical `InstantSerializer` with KotlinX datetime default serializer

SimonMarquis:
The default `Instant` serializer is already configured in:

https://github.com/Kotlin/kotlinx-datetime/blob/94bcc6ff1733c22ef4f937a25a276d3fd728a301/core/jvm/src/Instant.kt#L23-L24

And defined as `InstantIso8601Serializer` here:

https://github.com/Kotlin/kotlinx-datetime/blob/94bcc6ff1733c22ef4f937a25a276d3fd728a301/core/common/src/serializers/InstantSerializers.kt#L21-L33


tunjid:
Thanks for pointing this out!

PR #913, opened 2023-08-22
Refactor Search Feature

hoangchungk53qx1:
replace object with data object in search feature, simple format code…

### DIFF
--- a/feature/search/src/main/java/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
+++ b/feature/search/src/main/java/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
@@ -312,7 +312,11 @@ private fun SearchResultBody(
             state = state,
         ) {
             if (topics.isNotEmpty()) {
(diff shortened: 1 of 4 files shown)


feature/search/src/main/java/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
```diff
@@ -312,11 +312,7 @@ private fun SearchResultBody(
             state = state,
         ) {
             if (topics.isNotEmpty()) {
-                item(
-                    span = {
-                        GridItemSpan(maxLineSpan)
-                    },
-                ) {
+                item(span = { GridItemSpan(maxLineSpan) }) {
```
tunjid, line 315: For consistency, the `span` argument should probably be on the next line

hoangchungk53qx1: I solved this problem, thank you very much

tunjid (approved):

PR #919, opened 2023-08-23
Interests screen: replaces UI tests with screenshot tests

JoseAlcerreca:
This is a controversial one. The original instrumented tests were painstakingly checking for specific strings whereas the new screenshot tests are simple and verify a lot more.

We could add a @Ignore to the instrumented test in case we need the tests in the future, but I think we should go all in and have the goal to remove all `assertIsDisplayed`s if they're not testing any UI logic to reduce build and emulator times.


feature/interests/src/test/java/com/google/samples/apps/nowinandroid/interests/InterestsScreenScreenshotTests.kt
```diff
@@ -0,0 +1,85 @@
...
+    fun interestsScreen_empty() {
+        composeTestRule.captureMultiDevice("interestsScreen_empty") {
+            InterestsScreen(InterestsUiState.Empty)
+        }
+    }
+
+    @Test
+    fun interestsScreen_loading() {
+        composeTestRule.captureMultiDevice("interestsScreen_loading") {
+            InterestsScreen(InterestsUiState.Empty)
```
mmoczkowski, line 62:
```suggestion
            InterestsScreen(InterestsUiState.Loading)
```

mmoczkowski (changes requested):

feature/interests/src/test/java/com/google/samples/apps/nowinandroid/interests/InterestsScreenScreenshotTests.kt
JoseAlcerreca: Oops good catch!

mmoczkowski:
Existing instrumented tests to be replaced are:

```
niaLoadingWheel_inTopics_whenScreenIsLoading_showLoading
interestsWithTopics_whenTopicsFollowed_showFollowedAndUnfollowedTopicsWithInfo
topicsEmpty_whenDataIsEmptyOccurs_thenShowEmptyScreen
```

It seems to me, that everything that was evaluated by those tests is checked by the new screenshot tests. This is fine to me. I like the simplicity and comprehensiveness of screenshot tests. However, instead of removing instrumented tests completely, can we think of an instrumented test that cannot be replaced by a screenshot test?

JoseAlcerreca:
Good question! Yes, the navigation tests (NavigationTest.kt) have user actions and even though we could replace the assertions with screenshots, there would be a lot of them. 

There are also some tests that use APIs like `performScrollToNode`. This would be easy to do with Roborazzi, but we're trying to keep user actions out of screenshot tests for compatibility with the preview-based tool.

mmoczkowski (commented):
I like it. Let's see what @dturner has to say about this.

dturner:
Can the screenshot tests be run locally, like instrumented tests? 

Running the instrumented tests locally is pretty important as it's far quicker at catching issues that way than solely relying on our painfully slow CI.

JoseAlcerreca:
1- They can be run locally, but you should expect small anti-aliasing differences, especially on mac. This is annoying, but not a big deal once you understand why it happens.

2- Screenshot tests are the first thing that we run on CI, so you have results in ~15 mins iirc.

keyboardsurfer:

PR #923, opened 2023-08-24
Prep work for JVM alternatives to modules/plugins

SimonMarquis:
Some modules, like `:core:datastore` probably don't need to be "Android" aware.
And this PR prepares the work to transform them into simpler JVM modules.

I've decided to go with an abstract `HiltConventionPlugin` and implements two concrete classes:
- `AndroidHiltConventionPlugin`
- `JvmHiltConventionPlugin`

But this could also be solved with a if/else check based on `pluginManager.hasPlugin("com.android.base")`, tell me if you prefer this solution.


build-logic/convention/src/main/kotlin/HiltConventionPlugin.kt
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
lihenggui, line 2:
```suggestion
 * Copyright 2023 The Android Open Source Project
```

build-logic/convention/src/main/kotlin/JvmHiltConventionPlugin.kt
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
lihenggui, line 2:
```suggestion
 * Copyright 2023 The Android Open Source Project
```

build-logic/convention/src/main/kotlin/HiltConventionPlugin.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import com.google.samples.apps.nowinandroid.libs
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.artifacts.VersionCatalog
+import org.gradle.kotlin.dsl.DependencyHandlerScope
+import org.gradle.kotlin.dsl.dependencies
+
+abstract class HiltConventionPlugin : Plugin<Project> {
```
tunjid, line 24: Could this be implemented as a delegate instead of inheriting?

Something like:

```
class HiltConventionPlugin(
    val basePluginId: String? = null, 
    dependencyHandler: DependencyHandlerScope.(libs: VersionCatalog) -> Unit
): Plugin<Project> {
 
        with(target) {
            with(pluginManager) {
                basePluginId?.let(::apply)
                // KAPT must go last to avoid build warnings.
                // See: https://stackoverflow.com/questions/70550883/warning-the-following-options-were-not-recognized-by-any-processor-dagger-f
                apply("org.jetbrains.kotlin.kapt")
            }
            dependencies {
                dependencyHandler()
            }
        }
    }
}
```

tunjid (approved):
LGTM, might need some rework when the KSP migration in #933 merged. Will postpone merging this till that's in.

SimonMarquis:
@tunjid now that Hilt KSP has been merged, I've update this review to take advantage of it.
As a result, `:core:common` is now a pure JVM module.

SimonMarquis:
@tunjid @dturner this PR is ready to be reviewed 🙇

SimonMarquis:
@tunjid @dturner any feedback on this PR?

build-logic/convention/src/main/kotlin/AndroidHiltConventionPlugin.kt
```diff
@@ -14,25 +14,13 @@
...
-                apply("dagger.hilt.android.plugin")
-            }
-
-            dependencies {
-                "implementation"(libs.findLibrary("hilt.android").get())
-                "ksp"(libs.findLibrary("hilt.compiler").get())
-            }
-
-        }
+class AndroidHiltConventionPlugin : Plugin<Project> by HiltConventionPlugin(
```
dturner, line 20: Why not just `AndroidHiltConventionPlugin : HiltConventionPlugin`? And make `HiltConventionPlugin` an `abstract class`?

build-logic/convention/src/main/kotlin/HiltConventionPlugin.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import com.google.samples.apps.nowinandroid.libs
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.artifacts.VersionCatalog
+import org.gradle.kotlin.dsl.DependencyHandlerScope
+import org.gradle.kotlin.dsl.dependencies
+
+class HiltConventionPlugin(
```
dturner, line 24:
```suggestion
abstract class HiltConventionPlugin(
```

build-logic/convention/src/main/kotlin/JvmHiltConventionPlugin.kt
```diff
@@ -0,0 +1,24 @@
...
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+
+class JvmHiltConventionPlugin : Plugin<Project> by HiltConventionPlugin(
```
dturner, line 20:
```suggestion
class JvmHiltConventionPlugin : HiltConventionPlugin(
```

dturner:
> But this could also be solved with a if/else check based on pluginManager.hasPlugin("com.android.base"), tell me if you prefer this solution.

I added my review just in case we decide to keep the "2 concrete classes -> 1 abstract class" approach, however, I'm leaning towards your suggestion of an if/else check and a single `HiltConventionPlugin` just to keep things simple. The current approach requires quite a bit of effort to figure out what plugins and dependencies are being applied.

build-logic/convention/src/main/kotlin/AndroidHiltConventionPlugin.kt
SimonMarquis: It's just that I always try to favor composition over inheritence.
But I can switch the current code to abtract class if you prefer, let me know 👍

SimonMarquis:
Let's try the if/else version then :)

dturner (approved):

dturner:
You didn't even need an if/else :)

PR #898, opened 2023-08-14
Create `spotlessApply` permanent run configuration

SimonMarquis:
This will help devs run the right command, and it's easier to find.

image

| Run Anything | Run Configuration |
|---|---|
| Double <kbd>Ctrl</kbd> | <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>F10</kbd> |
| image | image |

PR #927, opened 2023-08-27
Create `Compose compiler metrics` permanent run configuration

SimonMarquis:
This will help devs run the right command, and it's easier to find.

image

### DIFF
--- a/.run/Compose compiler metrics.run.xml
+++ b/.run/Compose compiler metrics.run.xml
@@ -1,0 +1,23 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Compose compiler metrics" type="GradleRunConfiguration" factoryName="Gradle">
+    <ExternalSystemSettings>
+      <option name="executionName" />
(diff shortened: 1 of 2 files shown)


SimonMarquis:
I'm also wondering if it really makes sense to compile both `demo` and `prod` build types.
Shouldn't it focus only on the `prodRelease` variant?

SimonMarquis:
🔔 @mmoczkowski this is a followup on
- #898

Also, do you have any opinion on my previous comment? (`assembleProdRelease`/`assembleDemoRelease` vs `assembleRelease`)

dturner:
I'm not sure that an extra run configuration is necessary here but I'll defer to the Compose team. @bentrengrove what do you think about this approach - is it overkill?

bentrengrove:
Unfortunately, I think this is overkill. There shouldn't be any issue just keeping it permanently enabled. Also, with strong skipping mode coming out soon we don't want to encourage manually annotating too many classes as stable, the compiler reports will still be useful but shouldn't be needed as much as now.

SimonMarquis:
Ok, let's close this then 🫡

PR #933, opened 2023-08-30
Update to Hilt 2.48 and use the KSP version

danysantiago:
* Moved protobuf to separate module to avoid having to wire KSP's Plugin with the Protobuf Plugin.
* `@Binds` function cannot be an extension function.


tunjid:
Thanks so much @danysantiago! Why can't `@Binds` be an extension function?

manuelvicnt (approved):

core/datastore-proto/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+ *   limitations under the License.
+ */
+
+plugins {
+    id("nowinandroid.android.library")
+    alias(libs.plugins.protobuf)
+}
+
+android {
+    namespace = "com.google.samples.apps.nowinandroid.core.datastore"
```
dturner, line 23:
```suggestion
    namespace = "com.google.samples.apps.nowinandroid.core.datastore.proto"
```

dturner:
Please fix the spotless issues by running: 

```
./gradlew spotlessApply --init-script gradle/init.gradle.kts --no-configuration-cache
```

core/datastore-proto/build.gradle.kts
danysantiago: Done

SimonMarquis:
😢 The current version of Hilt KSP (still experimental) triggers a lot of OOMs...

> e: java.lang.OutOfMemoryError: Metaspace

danysantiago:
> It seems like the current default is set by Gradle to [384m](https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory). Should we try to increase it on CI? (we are using 1g by default on local builds)

Oh noes! 😢 

It seems the project is already configured with 1g: https://github.com/android/nowinandroid/blob/main/gradle.properties#L11. Is this config not the same as the one in CI?

SimonMarquis:
OOF...

image

I confirm that ksp* tasks are responsible for these spikes in metaspace usage.
Removing `kotlin.compiler.execution.strategy=in-process` make these spikes less noticeable, even in the Kotlin compiler daemon which remains ~250m.

This Kotlin compiler setting was introduced in the first PR of this project #1, but I'm not sure why.

lihenggui:
Looks like Google is aware of this issue and mentioned in the release note:

> Also note that Dagger’s KSP processors are still in the alpha stage. So far we’ve focused mainly on trying to ensure correctness rather than optimize performance. Please apply due diligence when enabling ksp and report any bugs or performance issues at https://github.com/google/dagger/issues. The current list of known issues can be found [here](https://github.com/google/dagger/issues?q=is%3Aissue+is%3Aopen+label%3A%22area%3A+ksp%22+label%3A%22type%3A+bug%22).

Shall we report an issue to them?

core/datastore-proto/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+                    option("lite")
+                }
+            }
+        }
+    }
+}
+
+androidComponents.beforeVariants {
+    android.sourceSets.register(it.name) {
+        java.srcDir(buildDir.resolve("generated/source/proto/${it.name}/java"))
```
lihenggui, line 47: `buildDir` will be deprecated in Gradle 8.3: 'getter for buildDir: File!' is deprecated. Deprecated in Java

Consider using `layout.buildDirectory.get().asFile` instead

abalros2: Hi

core/datastore-proto/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+                }
+                register("kotlin") {
+                    option("lite")
+                }
+            }
+        }
+    }
+}
+
+androidComponents.beforeVariants {
```
lihenggui, line 45: We can remove this workaround since https://github.com/google/protobuf-gradle-plugin/issues/714 got fixed.

manuelvicnt:
While running locally, I could see an increase in memory metadata usage but it didn't go over 1GB.

### KAPT
kapt

### KSP
ksp

SimonMarquis:
@manuelvicnt did you get these results with `kotlin.compiler.execution.strategy=in-process`?

manuelvicnt:
@SimonMarquis Yeah, using the current state of this PR

danysantiago:
I've commented `kotlin.compiler.execution.strategy=in-process` for now to avoid the OOMs in due to increased metaspace usage. There has been reports that the spikes are due to changes in Kotlin 1.9.0 and not necessarily KSP + Hilt, the team is investigating internally but for now feel free to their hold on this PR or merge without the `in-process` if it works.

core/datastore-proto/build.gradle.kts
danysantiago: I'll leave this as-is for now, since its a copy & paste of what was in the `datastore` module.

core/datastore-proto/build.gradle.kts
danysantiago: Yup - I'll leave it like this for now since its a copy & paste of what was in the datastore module. It can be cleaned up in a follow-up.

JoseAlcerreca (changes requested):
holding until we investigate

jdkoren:
I'd like to look into this, what is the tool being used in this comment? https://github.com/android/nowinandroid/pull/933#issuecomment-1701170925

jdkoren:
@SimonMarquis which tool are you using in [this comment](https://github.com/android/nowinandroid/pull/933#issuecomment-1699695541)?

SimonMarquis:
This is [VisualVM](https://visualvm.github.io/)

SimonMarquis:
FWIW:

> the current implementation of the in-process strategy is suboptimal, however we have plans to improve it

https://kotlinlang.slack.com/archives/C19FD9681/p1696201557559239?thread_ts=1695989901.167479&cid=C19FD9681

dturner:
@SimonMarquis Thanks for sharing that thread on KotlinLang. Seems like we shouldn't be using `in-process` anyway and it's a hangover from the original project setup. 

Full context on why using `in-process` is suboptimal from [the same thread](https://kotlinlang.slack.com/archives/C19FD9681/p1696271009205129?thread_ts=1695989901.167479&cid=C19FD9681)
> Currently it’s creating a new class loader for each compilation task, that may cause higher memory consumption and much slower compilation because JVM shall load classes and perform JIT compilation for the compiler classes almost from scratch. So I won’t recommend using it for the main build as I consider it consists of at least a few modules. However, it’s ok to use it for buildSrc or the single build-logic module. It shouldn’t cause any problems in this case if you’re ok with not having incremental compilation for that module.


I'll summarise what JetBrains said about `in-process`:

gradle/libs.versions.toml
```diff
@@ -37,7 +37,7 @@ gmsPlugin = "4.4.0"
 googleOss = "17.0.1"
 googleOssPlugin = "0.10.6"
 hilt = "2.48.1"
-hiltExt = "1.0.0"
+hiltExt = "1.1.0-alpha01"
```
SimonMarquis, line 40: Stable version is ready live: https://developer.android.com/jetpack/androidx/releases/hilt#1.1.0
```suggestion
hiltExt = "1.1.0"
```

.github/ci-gradle.properties
```diff
@@ -19,7 +19,8 @@ org.gradle.parallel=true
 org.gradle.workers.max=2
 
 kotlin.incremental=false
-kotlin.compiler.execution.strategy=in-process
+# Commented for avoid OOMs due to metaspace usage in KSP with Kotlin 1.9.0
+# kotlin.compiler.execution.strategy=in-process
```
SimonMarquis, line 23: This can be entirely removed (see https://github.com/android/nowinandroid/pull/933#issuecomment-1804041711)
```suggestion
```

ffgiraldez:
I'm still having issues with hilt 1.1.0 and ksp due to OOM

gradle/libs.versions.toml
danysantiago: Done

dturner:
@ffgiraldez Please could you share more details of your project and reproduction steps for the OOM by filing a new issue here: https://github.com/google/dagger/issues

PR #937, opened 2023-08-31
ConventionPlugins: Make version catalog more type safe

AntsyLich:


build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/ProjectExtensions.kt
```diff
@@ -16,10 +16,8 @@
 
 package com.google.samples.apps.nowinandroid
 
+import org.gradle.accessors.dm.LibrariesForLibs
```
tunjid, line 19: Is this gradle API internal and subject to change? If so, we shouldn't use it.

build-logic/convention/build.gradle.kts
```diff
@@ -40,6 +40,9 @@ dependencies {
     compileOnly(libs.firebase.performance.gradlePlugin)
     compileOnly(libs.kotlin.gradlePlugin)
     compileOnly(libs.ksp.gradlePlugin)
+
+    // These dependency below allows us to get [LibrariesForLibs] the class provided by [libs] in the plugins
+    implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
```
alexvanyo, line 45: This looks like the same hack/workaround as #318 for the underlying issue of https://github.com/gradle/gradle/issues/15383.

alexvanyo (changes requested):
Similar to the comments in #318 , I'd prefer to avoid adding a brittle workaround to a Gradle limitation in the build setup. Looking at discussion in https://github.com/gradle/gradle/issues/15383, it's possible this will break upon AGP, JDK and Gradle updates, and then either this change will need to be reverted, or a new workaround would need to be found.

If you want to take on that responsibility in your own project for the type-safe benefit, feel free to, but I think we should hold off here. I'm definitely looking forward to being able to do this with a proper API provided by Gradle though in the future to fix #317 .

build-logic/convention/build.gradle.kts
AntsyLich: Right I missed #317 and #318 sorry for that. The motivation behind this was that the workaround is working for quite a time.

PR #942, opened 2023-09-11
NavigationSuiteScaffold

mmoczkowski:
1. Replace logic switching between `BottomBar` and `NavRail` navigation components with `NavigationSuiteScaffold` API
2. Use `NavigationSuiteScaffold` API to display `PermanentDrawer` on screen widths larger than `1240.dp`
3. Add new screenshot tests validating that `PermanentDrawer` is properly displayed: `largeScreenWidth_compactHeight_showsPermanentDrawer`, `largeScreenWidth_mediumHeight_showsPermanentDrawer`, `largeScreenWidth_expandedHeight_showsPermanentDrawer`
4. Remove `niaAppState_showBottomBar_compact`, `niaAppState_showNavRail_medium` and `niaAppState_showNavRail_large` integration tests. They are redundant and already covered by screenshot tests.
5. Update Compose BOM to `2023.09.00` - required for `NavigationSuiteScaffold` to work in integration tests.
6. Remove custom `notificationDot` modifier and replace is with `BadgedBox`

|Permanent Drawer|Navigation Rail|Bottom Bar|
|--|--|--|
||||


app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
```diff
@@ -133,6 +130,17 @@ class NiaAppState(
                 initialValue = emptySet(),
             )
 
+    @OptIn(ExperimentalMaterial3AdaptiveApi::class)
+    val navigationSuiteType: NavigationSuiteType
+        @Composable get() =
+            if (LocalConfiguration.current.screenWidthDp >= 1240) {
```
alexvanyo, line 136: Can we use the new `windowSizeAsState` here instead of querying the `LocalConfiguration`? That is the backing value that drives the `windowSizeClass` calculation, and it can differ from the `LocalConfiguration` value slightly.
Also for testing, it might be better if it was passed into `NiaAppState` similarly to the `windowSizeClass`.

alexvanyo (changes requested):

app/src/main/java/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
mmoczkowski: Hey Alex, good idea. However

1. `windowSizeAsState` returns screen size in pixels (`IntSize`). However, for the purposes of nav component selection logic you need a `DpSize`. I think it would useful if the `windowSizeAsState` returned `DpSize` directly so there's no need for conversion.
2. I do not see a need for a `windowSizeClass` at all. It's not flexible enough since the `Expanded` case covers both `NavRail` and `PermanentDrawer`. If I have screen size in Dp I don't need `windowSizeClass` anymore.

What do you think?

mmoczkowski: @alexvanyo ping 😄

alexvanyo: 1. Totally agreed
2. The 1240dp breakpoint is the odd one out for not aligning with the window size class, do you think it'd be too verbose to say `windowSizeClass.widthSizeClass == WindowWidthSizeClass.Expanded && windowSize.width >= 1240.dp`?

   Not using it here definitely keeps the code a bit simpler, though could you add a comment to say that the `600.dp` width aligns with the MEDIUM breakpoint?

mmoczkowski: The problem with `WindowSizeClass` visualised:

window_size_class_vs_m3_vs_camal

mmoczkowski: @alexvanyo I added a comment to the navigationSuiteType property explaining the selection logic of the Navigation Component

alexvanyo: After the internal discussion: how do you feel about removing the customization here entirely, and instead just rely on the logic that is provided by default by the library (which may still be subject to change)

mmoczkowski: @alexvanyo To better accommodate large-screen devices, we strongly advocate for the inclusion of a Permanent Drawer.

app/src/testDemo/screenshots/expandedWidth_mediumHeight_showsNavigationRail.png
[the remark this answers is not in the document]
mmoczkowski: @alexvanyo Not with current API. The `NavigationSuiteScaffold` composable is missing `NavigationSuiteColors` arguments, making it impossible to customize the colors of the navigation components' container and content.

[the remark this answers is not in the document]
alexvanyo: Unfortunate, thanks for looking into that! We should see if that is possible to do in some way by changing the API.

alexvanyo (approved):
Current limitations make sense with the current APIs, looks great @mmoczkowski !

app/build.gradle.kts
```diff
@@ -113,7 +113,7 @@ dependencies {
     implementation(libs.androidx.compose.runtime)
     implementation(libs.androidx.lifecycle.runtimeCompose)
     implementation(libs.androidx.compose.runtime.tracing)
-    implementation(libs.androidx.compose.material3.windowSizeClass)
+    implementation(libs.androidx.compose.material3.adaptive)
```
IanGClifton, line 116: Note that nav suite will be material3-navigation-suite

app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -139,9 +142,10 @@ class MainActivity : ComponentActivity() {
                     androidTheme = shouldUseAndroidTheme(uiState),
                     disableDynamicTheming = shouldDisableDynamicTheming(uiState),
                 ) {
+                    val windowSize by windowSizeAsState()
```
IanGClifton, line 145: windowSizeAsState is now collectWindowSizeAsState

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -39,25 +31,19 @@ import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.SnackbarResult.ActionPerformed
 import androidx.compose.material3.Text
 import androidx.compose.material3.TopAppBarDefaults
-import androidx.compose.material3.windowsizeclass.WindowSizeClass
+import androidx.compose.material3.adaptive.ExperimentalMaterial3AdaptiveApi
+import androidx.compose.material3.adaptive.NavigationSuiteScaffold
```
IanGClifton, line 35: import androidx.compose.material3.adaptive.navigation.suite.NavigationSuiteScaffold

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
```diff
@@ -16,12 +16,14 @@
 
 package com.google.samples.apps.nowinandroid.ui
 
-import androidx.compose.material3.windowsizeclass.WindowSizeClass
-import androidx.compose.material3.windowsizeclass.WindowWidthSizeClass
+import androidx.compose.material3.adaptive.ExperimentalMaterial3AdaptiveApi
+import androidx.compose.material3.adaptive.NavigationSuiteType
```
IanGClifton, line 20: import androidx.compose.material3.adaptive.navigation.suite.NavigationSuiteType

app/src/testDemo/screenshots/expandedWidth_mediumHeight_showsNavigationRail.png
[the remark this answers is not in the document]
IanGClifton: I think navigationSuiteColors was added as a param more recently, so this should be possible now. If not, we should definitely file a bug

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
mmoczkowski: Fixed

app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
mmoczkowski: Fixed

app/build.gradle.kts
mmoczkowski: Fixed

app/src/testDemo/screenshots/expandedWidth_mediumHeight_showsNavigationRail.png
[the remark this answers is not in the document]
mmoczkowski: Done, it works 😄

lihenggui:
Hello @mmoczkowski Would you be willing to reconsider reopening this pull request?

IanGClifton:
NavigationSuiteScaffold will get added back in later but the APIs need to stabilize and depend on a stable version of Compose.

PR #952, opened 2023-09-16
Fix wrong configuration name, leaking `work-testing` to runtime classpath

SimonMarquis:
Typo introduced recently when Roborazzi was added to the project:

https://github.com/android/nowinandroid/commit/886158d3cb9c7271bdf46241addc4db694174473#diff-8cff73265af19c059547b76aca8882cbaa3209291406f52df1dafbbc78e80c46


tunjid (approved):
Thanks so much for fixing this!

Issue #360, opened 2022-10-25
Multiple warnings during build


dturner:
Steps to repro: 

```
./gradlew testDemoDebugUnitTest
```

Expected result: 

No warnings

Actual result: 

```
$ ./gradlew testDemoDebugUnitTest
Starting a Gradle Daemon, 1 busy and 1 incompatible Daemons could not be reused, use --status for details
Configuration cache is an incubating feature.
Calculating task graph as no configuration cache is available for tasks: testDemoDebugUnitTest

> Task :core:ui:compileDemoDebugKotlin
w: /Users/donturner/Code/workspace-android/nowinandroid/core/ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt: (255, 5): Parameter 'newsResource' is never used

> Task :app:processDemoDebugMainManifest
/Users/donturner/Code/workspace-android/nowinandroid/app/src/main/AndroidManifest.xml Warning:
        meta-data#com.google.samples.apps.nowinandroid.sync.initializers.SyncInitializer was tagged at AndroidManifest.xml:0 to remove other declarations but no other declaration present

> Task :core:navigation:kaptDemoDebugKotlin
warning: The following options were not recognized by any processor: '[dagger.fastInit, dagger.hilt.android.internal.disableAndroidSuperclassValidation, dagger.hilt.internal.useAggregatingRootProcessor, kapt.kotlin.generated]'

> Task :core:datastore-test:kaptDemoDebugKotlin
warning: unknown enum constant NiaDispatchers.IO
  reason: class file for com.google.samples.apps.nowinandroid.core.network.NiaDispatchers not found
warning: unknown enum constant NiaDispatchers.IO
  reason: class file for com.google.samples.apps.nowinandroid.core.network.NiaDispatchers not found
warning: unknown enum constant NiaDispatchers.IO
  reason: class file for com.google.samples.apps.nowinandroid.core.network.NiaDispatchers not found
warning: unknown enum constant NiaDispatchers.IO
  reason: class file for com.google.samples.apps.nowinandroid.core.network.NiaDispatchers not found

> Task :core:data-test:kaptDemoDebugKotlin
warning: unknown enum constant NiaDispatchers.IO
  reason: class file for com.google.samples.apps.nowinandroid.core.network.NiaDispatchers not found
warning: unknown enum constant NiaDispatchers.IO
warning: unknown enum constant NiaDispatchers.IO
[...]

PR #953, opened 2023-09-16
Declare `SyncNotificationsService` only in `prod` sourceSet

SimonMarquis:
And move Firebase Messaging dependency to the appropriate configuration.

Relates to #360

```
nowinandroid/app/src/main/AndroidManifest.xml Warning:
        meta-data#com.google.samples.apps.nowinandroid.sync.initializers.SyncInitializer was tagged at AndroidManifest.xml:0 to remove other declarations but no other declaration present
```


dturner (approved):
Thanks @SimonMarquis

PR #954, opened 2023-09-20
Remove `nonTransitiveRClass` flag

Goooler:
It's the default from AGP 8.0, see https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes#default-changes


dturner (approved):
Thanks for doing this.

PR #956, opened 2023-09-20
Add Test for Saved Top Level Destination

KDW03:
This Pull Request addresses an existing TODO in the codebase. It adds a test to verify that the ‘Up’ arrow does not show when the user is on the ‘Saved’ top level destination. This ensures consistency with the existing tests for the ‘For you’ and ‘Interests’ top level destinations.

This code change still passes all the relevant tests.

Please let me know if any changes are required. Thank you for considering this contribution!


dturner (approved):
Thanks for doing this.

PR #639, opened 2023-03-29
Enable gradle type-safe project accessors and replace all occurrences

Xsims:
Closes #638

Issue #638, opened 2023-03-29
[FR]: Use Gradle type-safe project accessors


Xsims:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

I discovered in issue #319 that `enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")` has been removed from `build-logic` because it was not being used. 

However, it has never been added to the root project file `settings.gradle.kts`, and as a result, the current code is not robust against naming changes.

```kotlin
dependencies {
    implementation(project(":feature:interests"))
    implementation(project(":feature:foryou"))
    implementation(project(":feature:bookmarks"))
}
```

### Describe the solution

We should introduce `enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")`. This will improve readability and maintainability of the project by using a more consistent way to reference other modules.

With the [gradle type-safe project accessors](https://docs.gradle.org/7.0/userguide/declaring_dependencies.html#sec:type-safe-project-accessors), the above code snippet can be replaced with:
```kotlin
dependencies {
    implementation(projects.feature.interests)
    implementation(projects.feature.foryou)
    implementation(projects.feature.bookmarks)
}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dturner:
Thanks for raising this. Type safe project accessors were enabled in #958.

PR #958, opened 2023-09-22
Use typesafe project accessors

McGalanes:


SimonMarquis:
:muscle: nice! The initial intent was back in March, but it hasn't changed since.
- https://github.com/android/nowinandroid/issues/638
- https://github.com/android/nowinandroid/pull/639

The feature is still experimental, but I did not found any issue that could impact us in [this list](https://github.com/gradle/gradle/issues?q=is%3Aopen+is%3Aissue+label%3Ain%3Aproject-accessors).

dturner (approved):
Thanks for doing this @McGalanes. As well as the benefit of added type safety, Android Studio also provides auto completion so you don't have to remember the exact name/path of a module when adding it as a dependency.

PR #959, opened 2023-09-22
Use `alias` instead of `id` when specifying plugins

McGalanes:


dturner:
What's the benefit of doing this please? It seems to add a lot of noise for no benefit but maybe I'm missing something.

gradle/libs.versions.toml
```diff
@@ -157,3 +157,18 @@ ksp = { id = "com.google.devtools.ksp", version.ref = "ksp" }
 protobuf = { id = "com.google.protobuf", version.ref = "protobufPlugin" }
 roborazzi = { id = "io.github.takahirom.roborazzi", version.ref = "roborazzi" }
 secrets = { id = "com.google.android.libraries.mapsplatform.secrets-gradle-plugin", version.ref = "secrets" }
+nowinandroid-android-application = { id = "nowinandroid.android.application", version = "unspecified" }
```
dturner, line 160:
```suggestion

# Plugins defined by this project
nowinandroid-android-application = { id = "nowinandroid.android.application", version = "unspecified" }
```

dturner (approved):
Thanks for the explanation @mmoczkowski, makes sense.

dawidhyzy:
@McGalanes 
I'm trying to do the same but I get `The request for this plugin could not be satisfied because the plugin is already on the classpath with an unknown version, so compatibility cannot be checked.` Did you face the same issue?

McGalanes:
> @McGalanes I'm trying to do the same but I get `Error resolving plugin [id: 'myapp.android.application', version: 'unspecified']` `The request for this plugin could not be satisfied because the plugin is already on the classpath with an unknown version, so compatibility cannot be checked.` Did you face the same issue?

Yes, try disabling your plugins in your project gradle file:
https://github.com/McGalanes/groomr/blob/0340a04fcd97f50eeccd209f04685c0d0396a88a/build.gradle.kts#L29C1-L38C2

dawidhyzy:
I did that already. It's not a problem here because it is complaining about convention plugins defined by my project.

McGalanes:
if you already tried what I said and it still doesn't work I don't understand. Sorry mate

jsoberg:
@dawidhyzy did you ever happen to find a workaround for your issue? I'm experiencing the same problem while trying to add internal convention plugins in this manner and apply them.

dawidhyzy:
> @dawidhyzy did you ever happen to find a workaround for your issue? I'm experiencing the same problem while trying to add internal convention plugins in this manner and apply them.

Yes, I think I registered them in the main grade with `apply false`. I will double check that on Monday.

jsoberg:
I tried that as well (applying all the local plugins with false) but it was giving the same issue. I have a root project-level plugin that's applied so it might be a race condition with that?

ZhaohuaZeng-at:
I think there might be some confusions on what to add in the root project level plugins in previous conversations. 

I got the same error but was able to resolve after adding my defined plugins to the plugins list in the root project level. 

```
plugins {

    alias(libs.plugins.ksp).apply(false)
    // some other plugins... 

    // Add my defined plugins here
    alias(libs.plugins.my.android.application).apply(false)
    alias(libs.plugins.my.android.hilt).apply(false)
}
```

PR #960, opened 2023-09-23
[KMP]Adapt to Kotlin Multiplatform project

xiaobailong24:


Spirarel:
Do you have this working on iOS?

xiaobailong24:
> Do you have this working on iOS?

Have not. Working in process but a large work.

Spirarel:
No doubt. We'll I'll follow your fork. I'll be very interested to see if Compose can handle an app like this.

dturner:
Thanks for investigating this. Whilst it is exciting to see progress being made with KMP, making this project KMP compatible is a large undertaking which will need to start with a proposal and a design document. 

It seems like this PR is more research and experimentation. I would suggest experimenting on your own fork, then raising an issue which includes your suggested approach so we can discuss further. 

I will close this PR because we're not in a position to consider merging this right now. 

Side note: When submitting a PR please include a description which states what you've done and why you've done it.

PR #966, opened 2023-09-29
Update remember to rememberSaveable

### LINKED ISSUE #965: [Bug]: App-Nia-Catalog does not retain selections on rotating device

### What happened?

A bug happened!

Expected: Rotating device to keep same button selections.

Result: Rotating the device causes app-nia-catalog to lose selection states.

Proposal: Update references from `remember` to `rememberSaveable` in order to retain state through lifecycle changes. 

### Relevant logcat output

_No response_

### Code of Conduct

tjmtic:
Fixes #965 
Allows selection states to survive lifecycle updates, i.e. phone rotations.

Intended as a best practice, would be interested if there are implementation decisions against this as a default.


tunjid (approved):
Thanks for adding this. Looks good to me.

PR #969, opened 2023-10-03
Fit typo in `SystemTrayNotifier.kt`

kim-minseong:


dturner (approved):

PR #970, opened 2023-10-03
ListDetailPaneScaffold

mmoczkowski:


alexvanyo:
One behavior issue I noticed: on a small screen (without list+detail showing together), if you tap on a topic, then navigate back, then tap on the same topic, the second tap doesn't navigate to the topic again:

https://github.com/android/nowinandroid/assets/4217560/922899b4-f4ab-409a-9040-a4c7ee94095c

feature/interests/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/interests/InterestsViewModel.kt
```diff
@@ -16,46 +16,57 @@
...
+    )
 
     fun followTopic(followedTopicId: String, followed: Boolean) {
         viewModelScope.launch {
             userDataRepository.setTopicIdFollowed(followedTopicId, followed)
         }
     }
+
+    fun onTopicClick(topicId: String) {
+        viewModelScope.launch {
```
alexvanyo, line 59: There's no need for a `launch` here if we aren't calling a suspending function.

jdkoren:
> One behavior issue I noticed: on a small screen (without list+detail showing together), if you tap on a topic, then navigate back, then tap on the same topic, the second tap doesn't navigate to the topic again:

Good catch. I found a few more bugs with navigation, both on phone and tablet. Repro steps:

- From **For you**, tap on a topic chip.
  - Observed: App switched to **Interests** but shows the list pane (no detail pane, not even on tablet).
  - Expected: Switch to **Interests** and show the selected topic (detail pane if single pane, list and detail if multi-pane).
- In this state, tap on **For you**.
  - Observed: Nothing happens, **Interests remains selected**.
  - Expected: App switches back to **For You**.
- Tap on **Saved**, app switches to **Saved**. Now tap on **For you**.
  - Observed: App switches back to **Interests**.
  - Expected: App switches to **For you**. 


https://github.com/android/nowinandroid/assets/771156/acfa18ef-4ef4-4bbf-883d-bc32e85513dd

jdkoren:
Superseded by #1234

XiaomiTpro:
On 8 Mar 2024 01:16, Jonathan Koren ***@***.***> wrote:
Superseded by #1234

PR #947, opened 2023-09-15
Use lazy vertical staggered grid in feed to maximize space utilization

tunjid:
Replaces the content grid with a staggered grid bringing us more in line with the M3 [feed](https://m3.material.io/components/cards/guidelines#99e8d17d-5bde-4bb9-8784-0ca403325b10) [guidance](https://developer.android.com/guide/topics/large-screens/large-screen-canonical-layouts#feed).

Screenshots to be updated in a bit.

PR #972, opened 2023-10-04
Update primitive mutable states to mutableState API where applicable

### LINKED ISSUE #888: [FR]: Use primitive mutable states where appropriate.

### Describe the problem

In [Compose runtime 1.5.0](https://developer.android.com/jetpack/androidx/releases/compose-runtime#1.5.0), primitive versions of mutable state APIs were released to prevent primitive autoboxing.

### Describe the solution

Primitive mutable states are used all over the app. Replacing them with their specific primitive mutable state APIs, especially in high rate of update components like scrollbars would be useful.

### Additional context

_No response_

### Code of Conduct

tjmtic:
Fixes #888 

It looks like a few states were based on Int or Float, which are converted here. The rest it seems are either an object value or a boolean, and from the 1.5.0 commit: 

> We are not adding `BooleanState` here because the autoboxing penalty for
`Boolean` types is virtually nonexistent because the JVM manages a pool
of the boxed versions of true and false, meaning that the boxing
operation will never cause an allocation.

https://android-review.googlesource.com/c/platform/frameworks/support/+/2375802

So I think that covers them, but happy to make any other update.


gradle/libs.versions.toml
```diff
@@ -7,6 +7,7 @@ androidxAppCompat = "1.5.1"
 androidxBrowser = "1.4.0"
 androidxComposeBom = "2023.06.01"
 androidxComposeCompiler = "1.5.0"
+androidxComposeRuntime = "1.5.2"
```
SimonMarquis, line 10: Since #947 has been merged, this is no longer necessary and should be removed.
Compose BoM `2023.09.00` already depends on runtime `1.5.1` (this feature was added in 1.5.0).

tjmtic: @SimonMarquis thank you for pointing that out, updated now.

tunjid (approved):

PR #973, opened 2023-10-05
Run Android tests on FTL managed device

AndreSand:
Run Android test on Firebase test lab device. 
Run time: ~19mins.
Gradle tasks: $./gradlew app:myFtlDeviceDemoDebugAndroidTest

Android Studio Hedgehog | 2023.1.1 Canary 14

TODO:
- Add the workflow Ftl.yaml
- Add service account JSON file to Gtihub secrets


JoseAlcerreca:
Let's mark this PR as a draft PR if it's not ready for review.

app/build.gradle.kts
```diff
@@ -71,6 +72,16 @@ android {
             excludes.add("/META-INF/{AL2.0,LGPL2.1}")
         }
     }
+    firebaseTestLab {
```
yschimke, line 75: Is it worth making this block conditional on the file being present? or is it safe to include even when file is missing?

.github/workflows/Ftl.yaml
```diff
@@ -0,0 +1,44 @@
...
+          java-version: 17
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+
+      - name: Get service account JSON file
+        env:
+          PLAY_FILE: ${{ secrets.SERVICE_ACCOUNT }}
+        run: |
+          echo $SERVICE_ACCOUNT  > ./app/service-account.json
```
SimonMarquis, line 41:
```suggestion
          SERVICE_ACCOUNT: ${{ secrets.SERVICE_ACCOUNT }}
        run: |
          echo $SERVICE_ACCOUNT  > ./app/service-account.json
```

build.gradle.kts
```diff
@@ -41,4 +41,8 @@ plugins {
     alias(libs.plugins.ksp) apply false
     alias(libs.plugins.roborazzi) apply false
     alias(libs.plugins.secrets) apply false
+
+    alias(libs.plugins.firebase.testlab) apply false
+    }
```
SimonMarquis, line 46:
```suggestion
```

build.gradle.kts
```diff
@@ -41,4 +41,8 @@ plugins {
...
 }
```
SimonMarquis, line 47:
```suggestion
    alias(libs.plugins.firebase.testlab) apply false
}
```

gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 accompanist = "0.28.0"
 androidDesugarJdkLibs = "2.0.3"
-androidGradlePlugin = "8.1.1"
+androidGradlePlugin = "8.2.0-alpha14"
```
SimonMarquis, line 4: Updating to alpha versions of AGP is pretty risky IMO.

SimonMarquis:
Is there a specific rational behind using a dedicated workflow instead of merging this into `Build.yaml`?

gradle/libs.versions.toml
yschimke: Required for FTL https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:firebase/testlab/testlab-gradle-plugin/src/main/java/com/android/tools/firebase/testlab/gradle/TestLabGradlePlugin.kt;l=52

SimonMarquis: Yes, I understand that this version is a prerequisite for enabling FTL managed devices. I'm simply pointing out that it might have unexpected consequences on the overall build, since AGP is applied on almost all modules of this project. 😃

app/build.gradle.kts
AndreSand: I think it's safe, the file is only needed when we run the Gradle task for FTL ./gradlew app:myFtlDeviceDemoDebugAndroidTest. If the file is not found it will give error ```> Unable to find the application-default credentials to send a request to Firebase TestLab.```. 

If developers want to test locally they can add their service-account json to the app codebase path ./app/service-account.json.

AndreSand:
> Is there a specific rational behind using a dedicated workflow instead of merging this into `Build.yaml`?

No specific reason, just to make the CI faster

SimonMarquis:
This could still be merged into the `Build.yaml` without any dependency, like what the `test` job does:

https://github.com/android/nowinandroid/blob/a77d7b69b58238e2b5e934ab9e492c0932bf2904/.github/workflows/Build.yaml#L52

You said the run time was 19mins. Is it only the test execution time, or did you include the compilation time as well?
Did you experimented with sharding (this could improve the test execution time).

SimonMarquis:
By the way, this PR currently only runs instrumented tests from the `:app` module (26 tests) and not the full test suite.

gradle/libs.versions.toml
```diff
@@ -33,6 +33,7 @@ dependencyGuard = "0.4.3"
 firebaseBom = "32.4.0"
 firebaseCrashlyticsPlugin = "2.9.9"
 firebasePerfPlugin = "1.4.2"
+firebaseTestlab = "0.0.1-alpha05"
```
yschimke, line 36: Just a warning, if you ever need to bump this to alpha06 or above, it requires AGP 8.3. mentioning because of previous hesitancy to use unstable releases.

keyboardsurfer:

PR #978, opened 2023-10-06
Optimize compilation time by adjusting the order of the repository

penkzhou:
ref: [link](https://docs.gradle.org/current/userguide/performance.html#optimize_repository_order)

[Optimize repository order](https://docs.gradle.org/current/userguide/performance.html#optimize_repository_order)
When Gradle resolves dependencies, it searches through each repository in the declared order. To reduce the time spent searching for dependencies, declare the repository hosting the largest number of your dependencies first. This minimizes the number of network requests required to resolve all dependencies.

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -16,8 +16,8 @@
 
 dependencyResolutionManagement {
     repositories {
-        google()
         mavenCentral()
+        google()
     }
     versionCatalogs {
         create("libs") {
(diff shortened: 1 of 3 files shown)


tunjid:
I believe most of our dependencies are androidx dependencies found in the google maven repo, so it would seem the current order is correct. Unless the transient dependencies greatly outnumber the declared ones.

penkzhou:
> I believe most of our dependencies are androidx dependencies found in the google maven repo, so it would seem the current order is correct. Unless the transient dependencies greatly outnumber the declared ones.

You are right. This is my clean build stats:
Screenshot 2023-10-07 at 23 28 44

penkzhou:
So this pr should be ignore.

PR #981, opened 2023-10-07
Refactor/create extensions for implementations

### LINKED ISSUE #979: [Refactor] Why do we use libs.findLibrary(alias).get() in all implementations?
why do we use libs.findLibrary(alias).get() in all implementations

I think we can write extensions for this and similar implementations

```
internal fun Project.findLibrary(alias: String): Provider<MinimalExternalModuleDependency> {
    return libs.findLibrary(alias).get()
}

internal fun Project.implementation(alias: String) {
    dependencies.add("implementation", findLibrary(alias))
}
```

What you think?

boranfrkn:
resolve #979   
created extension functions for dependency management


build-logic/convention/src/main/kotlin/AndroidHiltConventionPlugin.kt
```diff
@@ -30,10 +31,10 @@ class AndroidHiltConventionPlugin : Plugin<Project> {
             }
 
             dependencies {
-                "implementation"(libs.findLibrary("hilt.android").get())
-                "kapt"(libs.findLibrary("hilt.compiler").get())
```
boranfrkn, line 34: I didn't write an extension for `Kapt` and `Ksp` because we use them only once. But I'm not sure, I would like to get your opinion on this. In this case, we have to make the `findLibrary` function external. If we write extension for `Kapt` and `Ksp`, we can make `findLibrary` function private.

build-logic/convention/src/main/kotlin/AndroidHiltConventionPlugin.kt
```diff
@@ -30,10 +31,10 @@ class AndroidHiltConventionPlugin : Plugin<Project> {
...
-                "kaptAndroidTest"(libs.findLibrary("hilt.compiler").get())
-                "kaptTest"(libs.findLibrary("hilt.compiler").get())
+                implementation("hilt.android")
+                "kapt"(findLibrary("hilt.compiler"))
```
boranfrkn, line 35: I didn't write an extension for `Kapt` and `Ksp` because we use them only once. But I'm not sure, I would like to get your opinion on this. In this case, we have to make the `findLibrary` function external. If we write extension for `Kapt` and `Ksp`, we can make `findLibrary` function private.

boranfrkn:
Could you review this PR? @dturner

boranfrkn:
Hi @bentrengrove @javadude, could you review this PR please?

build-logic/convention/src/main/kotlin/AndroidHiltConventionPlugin.kt
boranfrkn: any idea? @javadude

javadude (commented):
I've gotta talk with some folks internally within Google to see if this is an approach we want to recommend.

javadude:
After chatting with some of our Gradle folks, we don't want to merge this change. We get what you're doing, but as a sample, we want to keep it a little more direct for now.

There's some new support for the versionCatalogs extension accessor in precompiled scripts in Gradle 8.5 and we'll take a look at some point. (See https://docs.gradle.org/8.5/release-notes.html)

PR #983, opened 2023-10-08
Apply consistent use of waitAndFindObject extension

### LINKED ISSUE #982: [FR]: Consistent Application of waitAndFindObject Extension Function

### Describe the problem

I've noticed inconsistent use of the waitAndFindObject extension function defined in the benchmark module's Utils.kt.

image

### Describe the solution

While this function is properly utilized in this segment 
image
<br>

It's not applied in the corresponding section of interestActions.kt 
image
<br>

By applying the already-implemented waitAndFindObject extension function as shown below, the behavior remains consistent with the previous implementation
Untitled
<br>

I kindly suggest considering the application of this extension function wherever the same logic is executed. While there may not be any immediate concerns, I believe that this approach could prevent potential challenges in the future. Adopting this method would ensure greater consistency and would also facilitate clearer error messaging via the extension function. 

If you agree with this approach, I'd like to proceed with submitting a pull request. 😊

### Additional context

_No response_

### Code of Conduct

KDW03:
resolve #982
Apply consistent use of waitAndFindObject extension


dturner (approved):
Thanks for doing this.

PR #988, opened 2023-10-14
Create Build-Sonar

democicd:


dturner:
If you want to add something like this I suggest you do it on your own fork. If you think that adding SonarCloud is a good addition to the project please open an issue where we can discuss further.

PR #994, opened 2023-10-21
Topic Navigation Route & String Resource

Askeri-Muhendis:
String Resource instead of static texts
Topic Moving Navigation Route Key to const val

**Do tests pass?**
- [ ] Run local tests on `DemoDebug` variant: `./gradlew testDemoDebug`
- [ ] Check formatting: `./gradlew --init-script gradle/init.gradle.kts spotlessApply`

- [ ] [Sign the CLA](https://cla.developers.google.com/)
- [ ] Run `./tools/setup.sh`


dturner:
> String Resource instead of static texts

This should be changed throughout the project rather than in part of it.

> Topic Moving Navigation Route Key to const val

Please don't combine 2 changes into one PR. Please remove this change from this PR and submit a separate PR for another change.

PR #1001, opened 2023-10-26
Add automatic badging verification

alexvanyo:
This PR automates the manual badging process outlined here (https://developer.android.com/guide/topics/manifest/uses-feature-element#testing). Instead of manually needing to invoke `aapt2`, this PR sets up 3 Gradle tasks for each variant's artifact. Using `prodRelease` as an example, this PR creates three tasks:
- `generateProdReleaseBadging` - generates the badging information from the universal apk
- `updateProdReleaseBadging` - updates the golden, checked-in badging information from the generated version
- `checkProdReleaseBadging` - verifies that the generated badging information matches the golden version

This badging describes important pieces of information about the resulting app package, which impacts how the app is distributed. This includes package name, version code, sdk versions, requsted permissions, launchable activities, and features.

Many of these attributes are resolved dynamically as part of build scripts, meaning that it may not be obvious when something changes - it could be due to a build configuration change, a newly included library, or implicitly required features.

These attributes are also vital to know, as they impact whether an app is available to download on certain devices, 

In the spirit of https://developer.squareup.com/blog/surfacing-hidden-change-to-pull-requests/, committing this badging information into version control, and then verifying it as part of CI will require surfacing changes during review, which also provides an opportunity to allow checking if that change to badging is intentional.


dturner (approved):
Great idea. Thanks for adding this.

PR #1005, opened 2023-10-31
Edit calculate list sizes using the derivedStateOf() function

oguzsout:
If we have a list and we want to calculate the length of this list from another state object, we can use the derivedStateOf() function.

Do tests pass?

[✓] Run local tests on DemoDebug variant: ./gradlew testDemoDebug
[✓] Check formatting: ./gradlew --init-script gradle/init.gradle.kts spotlessApply
Is this your first pull request?

[ ✓] [Sign the CLA](https://cla.developers.google.com/)
[✓ ] Run ./tools/setup.sh
[✓ ] Import the code formatting style as explained in [the setup script](https://github.com/tools/setup.sh#L40).

### DIFF
--- a/feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
+++ b/feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
@@ -52,6 +52,7 @@ import androidx.compose.material3.TextField
 import androidx.compose.material3.TextFieldDefaults
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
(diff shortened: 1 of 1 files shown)


feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
```diff
@@ -298,9 +298,11 @@ private fun SearchResultBody(
     searchQuery: String = "",
 ) {
     val state = rememberLazyStaggeredGridState()
+    val itemsAvailable by remember {
```
tunjid, line 301: This won't do what you may think it does. The `remember` isn't keyed on topics or news resources. Also the variables aren't backed by compose state so this change introduces a bug.

feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
```diff
@@ -298,9 +298,11 @@ private fun SearchResultBody(
...
+        derivedStateOf { topics.size + newsResources.size }
```
hoc081098, line 302: derivedStateOf is unnecessary.
> It is useful when reading other State.value inside derivedStateOf

tunjid:
I'm going to close this PR as the changes introduce a few bugs. Feel free to ping me if you need more clarification.

PR #1006, opened 2023-10-31
upgrade to Firebase BoM 32.5.0 and remove Firebase KTX

thatfiredev:

**What I have done and why**
This PR should:
- upgrade Firebase to v32.5.0
- migrate from Firebase KTX to the main modules, per https://firebase.google.com/docs/android/kotlin-migration

**Do tests pass?**

- [x] Run `./tools/setup.sh`
- [x] Import the code formatting style as explained in [the setup script](/tools/setup.sh#L40).


dturner:
@thatfiredev This change adds the following permissions and properties to the app manifest: 

```
uses-permission: name='android.permission.ACCESS_ADSERVICES_ATTRIBUTION
uses-permission: name='android.permission.ACCESS_ADSERVICES_AD_ID
property: name='android.adservices.AD_SERVICES_CONFIG' resource='res/xml/ga_ad_services_config.xml'
```

These seem related to advertising but the app has no ads. Please could you explain why these entries are necessary? Or how they can be removed if they are not.

thatfiredev:
@dturner that was added to the Firebase Analytics SDK in [v21.5.0](https://firebase.google.com/support/release-notes/android#analytics_v21-5-0) to support the [Android Privacy Sandbox](https://developer.android.com/design-for-safety/privacy-sandbox/setup-api-access).

Since they're unrelated to the original intention of this PR, I went ahead and removed the permissions with `tools:node="remove"`.
I'm just not sure how to remove that `property` - how much of a concern is it to keep it? If it causes trouble, we can file a bug for the SDK team to not include it.

dturner:
Thanks for making the changes. 

> I'm just not sure how to remove that property - how much of a concern is it to keep it?

I guess that's the point, we don't know. Does it have any impact on what is shown in the Play Store? Or on the user's device at install or runtime? As this property is being added by a 3rd party library (Firebase) it makes sense for that library provider to justify the impact of any public facing changes to the app. 

Seems related to the Privacy Sandbox. Docs are here: https://developer.android.com/design-for-safety/privacy-sandbox/setup-api-access#configure-api-specific. Any chance you could investigate and let me know the impact of adding, or whether we should take steps to remove?

thatfiredev:
```
Any chance you could investigate and let me know the impact of adding, or whether we should take steps to remove?
```

Can do :) Will post my findings soon

dturner:
Any update here @thatfiredev?

app/src/main/AndroidManifest.xml
```diff
@@ -20,11 +20,13 @@
     <uses-permission android:name="android.permission.INTERNET" />
 
     <!--
-    Firebase automatically adds the AD_ID permission, even though we don't use it. If you use this
-    permission you must declare how you're using it to Google Play, otherwise the app will be
-    rejected when publishing it. To avoid this we remove the permission entirely.
+    Firebase automatically adds these AD_ID and ADSERVICES permissions, even though we don't use it.
```
dturner, line 23:
```suggestion
    Firebase automatically adds these AD_ID and ADSERVICES permissions, even though we don't use them.
```

app/src/main/AndroidManifest.xml
```diff
@@ -20,11 +20,13 @@
     <uses-permission android:name="android.permission.INTERNET" />
 
     <!--
-    Firebase automatically adds the AD_ID permission, even though we don't use it. If you use this
-    permission you must declare how you're using it to Google Play, otherwise the app will be
-    rejected when publishing it. To avoid this we remove the permission entirely.
+    Firebase automatically adds these AD_ID and ADSERVICES permissions, even though we don't use them.
+    If you use this permission you must declare how you're using it to Google Play, otherwise the
```
dturner, line 24:
```suggestion
    If you use these permissions you must declare how you're using them to Google Play, otherwise the
```

app/src/main/AndroidManifest.xml
```diff
@@ -20,11 +20,13 @@
...
+    app will be rejected when publishing it. To avoid this we remove the permission entirely.
```
dturner, line 25:
```suggestion
    app will be rejected when publishing it. To avoid this we remove the permissions entirely.
```

dturner (approved):

thatfiredev:
@dturner Sorry for the radio silence, I totally forgot to look into this, but it seems like the engineering team is looking at it internally: [b/340307974](http://b/340307974)

dturner:
No worries, thanks for the link. It enabled me to figure out how to remove the property. Done in #1539

PR #1010, opened 2023-11-02
Apply spotless automatically during build and commit to PR

dturner:
Update our CI build workflow to apply spotless formatting and add a commit to the source PR, instead of just checking it and failing the build. 

This avoids situations like this where formatting issues cause the build to fail: https://github.com/android/nowinandroid/actions/runs/6713253053/job/18297469836?pr=1006


tunjid (approved):
LGTM

PR #1017, opened 2023-11-03
Update MainDispatcherRule.kt

### LINKED ISSUE #1015: val testDispatcher should be private in MainDispatcherRule class
val testDispatcher should be private in testing module, specifically in MainDispatcherRule class because it is only accessed from MainDispatcherRule class only.

class MainDispatcherRule(
    val testDispatcher: TestDispatcher = UnconfinedTestDispatcher()
) : TestWatcher() {
    //......
}

mariomanhique:
**What I have done and why**
Updated MainDispatcherRule parameter accessibility to private because its only needs to be accessible inside that class.

Fixes #1015

**Do tests pass?**
N/A

Yes, I have signed the CLA


dturner:
Thanks for doing this.

Issue #1018, opened 2023-11-05
[Bug]: unit tests are executed twice per `Build.yaml` workflow execution


SimonMarquis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

As discovered in https://github.com/android/nowinandroid/issues/1011#issuecomment-1792195181, [`Build.yaml`](https://github.com/android/nowinandroid/blob/f414cd482e67807b558653d3b9ef31f7fc4d1bd4/.github/workflows/Build.yaml) contains two instances of `gradlew testDemoDebug testProdDebug`:

- in the `build` job [~5min 50s](https://github.com/android/nowinandroid/actions/runs/6744267534/job/18333858408#check-step-10): https://github.com/android/nowinandroid/blob/f414cd482e67807b558653d3b9ef31f7fc4d1bd4/.github/workflows/Build.yaml#L49-L50

- in the `test` job [~14min 30s](https://github.com/android/nowinandroid/actions/runs/6744267534/job/18333858232#check-step-11):
https://github.com/android/nowinandroid/blob/f414cd482e67807b558653d3b9ef31f7fc4d1bd4/.github/workflows/Build.yaml#L107-L109

(the duration difference can be explain by the `build` job having already compiled most of the needed source code)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ayaankhan98:
Can I proceed to fix this issue?

SimonMarquis:
I think we should first find out what to do with #1011 and think about the bigger picture of job dependencies.

ayaankhan98:
Okay, makes sense.

JoseAlcerreca:
Fixed in #1019

PR #1019, opened 2023-11-06
[CI] Runs build in macos to generate a cache for androidTest runs

### LINKED ISSUE #1011: In CI the `test` workflow shouldn't run if the `build` workflow fails
The `test` workflow can take ~10 mins to run and it blocks other workflows from running. It's pointless to run the tests if the build fails.

JoseAlcerreca:
Experiments to fix #1011 and #1018

Relying on gradle-build-action for cache sharing. Followed some guidance in:

https://github.com/gradle/gradle-build-action#deduplication-of-gradle-user-home-cache-entries

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -11,44 +11,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-    timeout-minutes: 90
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Validate Gradle Wrapper
-        uses: gradle/wrapper-validation-action@v1
-
-      - name: Copy CI gradle.properties
-        run: mkdir -p ~/.gradle ; cp .github/ci-gradle.properties ~/.gradle/gradle.properties        
-
-      - name: Set up JDK 17
-        uses: actions/setup-java@v3
-        with:
-          distribution: 'zulu'
-          java-version: 17
-
-      - name: Setup Gradle
-        uses: gradle/gradle-build-action@v2
-
-      - name: Check spotless
-        run: ./gradlew spotlessCheck --init-script gradle/init.gradle.kts --no-configuration-cache
-
-      - name: Build all build type and flavor permutations
-        run: ./gradlew assemble
-
-      - name: Upload build outputs (APKs)
-        uses: actions/upload-artifact@v3
-        with:
-          name: APKs
-          path: '**/build/outputs/apk/**/*.apk'
-
-      - name: Run local tests
-        run: ./gradlew testDemoDebug testProdDebug :lint:test
-
   test:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


JoseAlcerreca:
So far it seems like the linux job takes 9-11 minutes (with cache hits), and will fails screenshot tests first, then jvm tests, spotless. Then it builds and uploads APKs. We probably want to fail spotless first.

If the build is broken, it can be caught by either the linux job or the `androidTest-build` job, whichever finds it first because they run in parallel.

JoseAlcerreca:
Looking at the low-hanging fruit for the androidTest-run job:

1. Runners wait for the emulator to boot for 100 seconds.

After that, the connectedCheck task is run and then gradle takes a while to assemble everything.

This seems like it could be done in parallel? Not sure how much CPU the emulator boot takes. 

2. Caching the emulator image 
https://github.com/ReactiveCircus/android-emulator-runner/issues/229

This looks good but wouldn't it take most of the 10Gb cache?

3. Configuration cache is disabled. See #1022 

4. The reactivecircus action seems to be downloading and installing a lot of stuff that should already be present in macos-latest. https://github.com/ReactiveCircus/android-emulator-runner/issues/265

For example, the command `[command]/bin/sh	-c	\sdkmanager	--install	system-images;android-26;default;x86_64'	--channel=0	>	/dev/null`

takes 53 seconds to run.

JoseAlcerreca:
On 2: I'm running a test in https://github.com/JoseAlcerreca/nowinandroid/actions/runs/6788492716/job/18453744948 to see how much cache it takes. After this run, I'll try with an ATD image.

JoseAlcerreca:
> On 2: I'm running a test in https://github.com/JoseAlcerreca/nowinandroid/actions/runs/6788492716/job/18453744948 to see how much cache it takes. After this run, I'll try with an ATD image.

avd-26 Cache Size: ~4551 MB (4771592759 B)
avd-30 Cache Size: ~3835 MB (4021327173 B)

We'd probably run out of cache for other things that might impact CI times more.

yschimke:
> This looks good but wouldn't it take most of the 10Gb cache?

Should we consider a separate job that runs, produces the AVD snapshot, then upload as build artifacts?

It might be worthwhile having another form of cache for this?

JoseAlcerreca:
> Should we consider a separate job that runs, produces the AVD snapshot, then upload as build artifacts instead of cache?

I don't think there's an API to download the build artifacts from another workflow, is there?
From a quick search:
"Downloading files: You can only download artifacts that were uploaded during the same workflow run. When you download a file, you can reference it by name."

JoseAlcerreca:
I think we can start reviewing this PR:

(1) Moved some of the build to a previous step so that gradle has a bit more resources to run. Probably irrelevant.

(2) Can't do this because of the size of the AVD images

(3) Pinged the authors of the plugin and will look into it separately. We'll probably skip the problematic tasks on CI.

(4) Filed https://github.com/android/nowinandroid/issues/1026 for tracking

.github/workflows/Build.yaml
```diff
@@ -106,7 +69,19 @@ jobs:
       # Run local tests after screenshot tests to avoid wrong UP-TO-DATE. TODO: Ignore screenshots.
       - name: Run local tests
         if: always()
-        run: ./gradlew testDemoDebug testProdDebug
+        run: ./gradlew testDemoDebug testProdDebug :lint:test
+
+      - name: Check spotless
+        run: ./gradlew spotlessCheck --init-script gradle/init.gradle.kts --no-configuration-cache
```
SimonMarquis, line 75: nit: could we avoid checking for spotless this late in the job?
Formatting issues happen very frequently, and checking for them after compiling and testing the codebase wastes precious time (for devs and the CI).

This also means, that removing `needs: build` from `androidTest` now will run entirely, even if formatting issues happen.

JoseAlcerreca: Ah, I meant to do that, thanks.

> This also means, that removing needs: build from androidTest now will run entirely, even if formatting issues happen.

Yeah this is fine, as long as one job fails.

JoseAlcerreca:
"Configuration cache entry stored." 

Now the next run should be faster.

JoseAlcerreca:
Took 21 minutes in API 26 and 30 in API 30 without configuration cache. Triggering another build now.

.github/workflows/Build.yaml
SimonMarquis: > Yeah this is fine, as long as one job fails.

I'm really uncomfortable with this. This will waste ~40 minutes of CI every time this happens.

dturner (approved):

PR #788, opened 2023-06-07
[do not merge] Build project with K2 compiler

zsmb13:
This PR demonstrates trying the [K2 compiler](https://kotlinlang.org/docs/whatsnew19.html#new-kotlin-k2-compiler-updates) in an Android project.

It should be continuously rebased onto `main` to ensure that everything still works on the K2 compiler, but only merged once K2 is stable.

PR #1025, opened 2023-11-07
🚧Kotlin K2 compiler

SimonMarquis:


lihenggui:
It seems to be the same as https://github.com/android/nowinandroid/pull/788

SimonMarquis:
@lihenggui yes it is :) but with more recent versions.
Don't get me wrong, I'm not trying to overtake this PR in any way 😛, but only to try with the latest dependencies.

SimonMarquis:
Closing, as K2 is now enabled by default on Kotlin 2.0.0, see https://github.com/android/nowinandroid/pull/1036

PR #1035, opened 2023-11-14
Add dependency guard plugin

jdkoren:
This adds the [Dependency Guard][1] gradle plugin and also commits baseline files for all `*RuntimeClasspath` configurations. Subsequent PRs will incorporate verification against these baselines in CI.

[1]: https://github.com/dropbox/dependency-guard


app/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ plugins {
     id("jacoco")
     alias(libs.plugins.nowinandroid.android.application.firebase)
     id("com.google.android.gms.oss-licenses-plugin")
+    alias(libs.plugins.dependencyGuard)
```
JoseAlcerreca, line 27: Should it be part of the Android Application convention plugin?

jdkoren: I had thought about that briefly. The only other app is the catalog, and that one literally is just UI, so I'm not sure how valuable that would be. What do you think?

app/build.gradle.kts
```diff
@@ -133,3 +134,14 @@ dependencies {
     kspTest(libs.hilt.compiler)
 
 }
+
+dependencyGuard {
+
```
SimonMarquis, line 139: Do we need to track all configuration? I would be in favor of only tracking the production build `prodReleaseRuntimeClasspath`.

app-nia-catalog/dependencies/releaseRuntimeClasspath.txt
SimonMarquis, line 1: This should probably either be removed, or configured correctly in the corresponding build.gradle.kts file of `:app-nia-catalog`.

app/build.gradle.kts
jdkoren: Added to the Android Application convention plugin

app/build.gradle.kts
jdkoren: Done

app-nia-catalog/dependencies/releaseRuntimeClasspath.txt
jdkoren: Done. Not sure how that happened :sweat_smile:

SimonMarquis (commented):
LGTM 👍

tunjid (approved):
LGTM

PR #1024, opened 2023-11-07
Update AGP 8.1.3

SimonMarquis:
No changelog for recent AGP version...
https://developer.android.com/build/releases/gradle-plugin

PR #1038, opened 2023-11-16
Update AGP 8.1.4

SimonMarquis:
No changelog for recent AGP version...
https://developer.android.com/build/releases/gradle-plugin

Follow up on #1024


SimonMarquis:
- #1114

PR #1060, opened 2023-11-29
Configure baseline profiles in the baselineProfile lambda

keyboardsurfer:
This is the correct and easiest way to set up baseline profiles correctly. See b/313428246 for more context.


app/build.gradle.kts
```diff
@@ -80,6 +78,20 @@ android {
         }
     }
     namespace = "com.google.samples.apps.nowinandroid"
+
+    baselineProfile {
```
malbano, line 82: This should be outside the android component.

You can refer to the example in [this integration test](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:benchmark/integration-tests/baselineprofile-flavors-consumer/build.gradle;l=67).

benchmarks/build.gradle.kts
```diff
@@ -46,9 +47,9 @@ android {
     }
 
     testOptions.managedDevices.devices {
-        create<com.android.build.api.dsl.ManagedVirtualDevice>("pixel6Api33") {
+        create<ManagedVirtualDevice>("pixel6Api31") {
             device = "Pixel 6"
-            apiLevel = 33
```
mlykotom, line 51: why downgrading the api level?

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/foryou/ForYouActions.kt
```diff
@@ -41,6 +44,12 @@ fun MacrobenchmarkScope.forYouWaitForContent() {
  */
 fun MacrobenchmarkScope.forYouSelectTopics(recheckTopicsIfChecked: Boolean = false) {
     val topics = device.findObject(By.res("forYou:topicSelection"))
+    val withChildren = topics.childCount != 0
+    if (!withChildren) {
+        // TODO: Ensure ForYou has topics.
+        Log.w(TAG, "no topics found, can't scroll for baseline profile generation.")
```
mlykotom, line 50: I think we should keep failing the generator if there's no topic 🤔 . I think this happens when something is loading longer than this test anticipates. Or am I missing something?

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/foryou/ForYouActions.kt
```diff
@@ -25,6 +26,8 @@ import com.google.samples.apps.nowinandroid.waitAndFindObject
 import com.google.samples.apps.nowinandroid.waitForObjectOnTopAppBar
 import org.junit.Assert.fail
 
+private const val TAG = "ForYouActions"
+
 fun MacrobenchmarkScope.forYouWaitForContent() {
     // Wait until content is loaded by checking if topics are loaded
     device.wait(Until.gone(By.res("loadingWheel")), 5_000)
```
mlykotom, line 33: In the past there was an issue sometimes that workmanager would sleep for ~20 seconds when you kill the app process too often, which was happening for `CompilationMode.Partial()`. Maybe we need to increase the timeout here?

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
```diff
@@ -35,7 +38,12 @@ fun MacrobenchmarkScope.goToInterestsScreen() {
 fun MacrobenchmarkScope.interestsScrollTopicsDownUp() {
     device.wait(Until.hasObject(By.res("interests:topics")), 5_000)
     val topicsList = device.findObject(By.res("interests:topics"))
-    device.flingElementDownUp(topicsList)
+    if (topicsList != null) {
+        // TODO: Ensure topics are availble.
+        device.flingElementDownUp(topicsList)
+    } else {
+        Log.w(TAG, "No topics found, can't scroll during baseline profile generation.")
```
mlykotom, line 45: I would suggest failing the test as well. Given benchmark is a test, it would be hard to figure out if the test is doing the right thing. WDYT?

benchmarks/build.gradle.kts
keyboardsurfer: Reverted

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
keyboardsurfer: Failing now.

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/foryou/ForYouActions.kt
keyboardsurfer: Failing now.

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
```diff
@@ -35,7 +39,13 @@ fun MacrobenchmarkScope.goToInterestsScreen() {
 fun MacrobenchmarkScope.interestsScrollTopicsDownUp() {
     device.wait(Until.hasObject(By.res("interests:topics")), 5_000)
     val topicsList = device.findObject(By.res("interests:topics"))
-    device.flingElementDownUp(topicsList)
+    if (topicsList != null) {
+        // TODO: Ensure topics are availble.
+        device.flingElementDownUp(topicsList)
+    } else {
+        Log.e(TAG, "No topics found, can't scroll during baseline profile generation.")
+        fail()
```
mlykotom, line 47: Would it make more sense to show the error in the `fail` function? You would see it in the test output.
If it's just in the Logcat, you would need to search for it in the logs and wouldn't see it in CI output for example.

```suggestion
        fail("No topics found, can't scroll during baseline profile generation.")
```

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/foryou/ForYouActions.kt
```diff
@@ -41,6 +44,12 @@ fun MacrobenchmarkScope.forYouWaitForContent() {
  */
 fun MacrobenchmarkScope.forYouSelectTopics(recheckTopicsIfChecked: Boolean = false) {
     val topics = device.findObject(By.res("forYou:topicSelection"))
+    val withChildren = topics.childCount != 0
+    if (!withChildren) {
+        // TODO: Ensure ForYou has topics.
+        Log.e(TAG, "no topics found, can't scroll for baseline profile generation.")
+        fail()
```
mlykotom, line 51:
```suggestion
        fail("No topics found, can't scroll for baseline profile generation.")
```

mlykotom (approved):
🎉

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/foryou/ForYouActions.kt
```diff
@@ -41,6 +43,11 @@ fun MacrobenchmarkScope.forYouWaitForContent() {
  */
 fun MacrobenchmarkScope.forYouSelectTopics(recheckTopicsIfChecked: Boolean = false) {
     val topics = device.findObject(By.res("forYou:topicSelection"))
+    val withChildren = topics.childCount != 0
+    if (!withChildren) {
```
Jaehwa-Noh, line 47:
```suggestion
    val noChildren = topics.childCount == 0
    if (noChildren) {
```
Maybe use `noChildren` is more readable, or noTopics.

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/foryou/ForYouActions.kt
```diff
@@ -25,6 +25,8 @@ import com.google.samples.apps.nowinandroid.waitAndFindObject
 import com.google.samples.apps.nowinandroid.waitForObjectOnTopAppBar
 import org.junit.Assert.fail
 
+private const val TAG = "ForYouActions"
```
Jaehwa-Noh, line 28:
```suggestion
```
It seemed to be no where use the TAG.
Also other files TAGs

benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/interests/InterestsActions.kt
```diff
@@ -21,6 +21,9 @@ import androidx.test.uiautomator.By
 import androidx.test.uiautomator.Until
 import com.google.samples.apps.nowinandroid.flingElementDownUp
 import com.google.samples.apps.nowinandroid.waitForObjectOnTopAppBar
+import org.junit.Assert.fail
+
+private const val TAG = "InterestsActions"
```
Jaehwa-Noh, line 26:
```suggestion
```
Ditto.

PR #1063, opened 2023-11-30
Move Scrollbar state reads out of composition

bentrengrove:
The way scrollbar was implemented, it observes all changes in composition. This is bad for scroll performance causing recomposition on every frame of scroll that doesn't need to happen.

This PR moves the work of observing the scroll state into the layout phase.

### Benchmarks Results
Iterations: 10 (so probably liable to a lot of noise)
#### Before
```
ScrollForYouFeedBenchmark_scrollFeedCompilationBaselineProfile
frameDurationCpuMs   P50   7.0,   P90   9.0,   P95  14.0,   P99  15.2
frameOverrunMs       P50  -5.7,   P90  -1.1,   P95   5.0,   P99   5.8
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```
#### After
```
ScrollForYouFeedBenchmark_scrollFeedCompilationBaselineProfile
frameDurationCpuMs   P50   3.7 (+89%),   P90  6.7 (+34%),   P95  10.7  (+30%),   P99  16.0 (-5%)
frameOverrunMs       P50  -8.3,   P90 -5.4,   P95  -1.0,   P99   5.5
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```

### DIFF
--- a/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/Scrollbar.kt
+++ b/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/Scrollbar.kt
@@ -32,9 +32,9 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.State
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableFloatStateOf
+import androidx.compose.runtime.mutableLongStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
@@ -44,10 +44,10 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.input.pointer.PointerInputChange
 import androidx.compose.ui.input.pointer.pointerInput
+import androidx.compose.ui.layout.Layout
 import androidx.compose.ui.layout.layout
 import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.layout.positionInRoot
-import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.unit.Constraints
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.IntOffset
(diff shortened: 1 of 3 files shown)


bentrengrove:
Ran out of time for the day, but keen for your thoughts, TJ.
I probably still need to clean this up sorry

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/Scrollbar.kt
```diff
@@ -320,84 +304,123 @@ fun Scrollbar(
...
+                        interactionThumbTravelPercent.isNaN() -> state.thumbMovedPercent
+                        else -> interactionThumbTravelPercent
+                    }
+
+                    val thumbMovedPx = min(
+                        a = track.size * thumbTravelPercent,
+                        b = track.size - thumbSizePx,
+                    )
+
+                    val scrollbarThumbMovedDp = max(
```
tunjid, line 328: The unit here is still pixels instead of dp, so maybe the val name should be `scrollbarThumbMovedPx`?

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/Scrollbar.kt
```diff
@@ -320,84 +304,123 @@ fun Scrollbar(
...
+                                maxWidth = constraints.maxWidth,
+                                minHeight = thumbSizePx.roundToInt(),
+                                maxHeight = thumbSizePx.roundToInt()
+                            )
+                        }
+                    }
+
+                    val placeable = measurable.measure(constrained)
+                    layout(placeable.width, placeable.height) {
+                        placeable.place(offset)
```
tunjid, line 365: Noice

tunjid:
LGTM, thanks so much for this! I was wondering, `ScrollbarState` itself is a value class for a packed `Float` primitive. For something that changes so frequently on scroll, I was trying to find a way to use `mutableFloatStateOf` for it's backing value instead of just `mutableStateOf`. Is there a way to do this?

bentrengrove:
> LGTM, thanks so much for this! I was wondering, `ScrollbarState` itself is a value class for a packed `Float` primitive. For something that changes so frequently on scroll, I was trying to find a way to use `mutableFloatStateOf` for it's backing value instead of just `mutableStateOf`. Is there a way to do this?

There sure is, I have updated the implementation with this. I'd love to know how much performance the packing you are doing is saving, I think its probably pretty minimal but I'm curious. You could probably save a bit of complexity just by having two 1mutableFloatStateOf` instead of packing into one `mutableLongState`.

I also added a custom Modifier.Node implementation that more efficiently updates it's color by not reading the color state in composition. You could actually go really far with this and move all the interaction observation into the Modifier.Node and then it would fully scroll with no recomposition needed. That was too much for today though.

I'm measuring the benchmark now to update the PR

tunjid (approved):

PR #1064, opened 2023-11-30
Run dependencyGuard task during CI

jdkoren:
Run the `dependencyGuard` task during CI, which will fail if there are detected changes.

Also add github actions to update the baseline files for PRs.


.github/renovate.json
```diff
@@ -12,5 +12,10 @@
       "matchPackageNames": ["com.google.protobuf"],
       "allowedVersions": "<=0.8.19"
     }
-  ]
+  ],
+  "postUpgradeTasks": {
+    "commands": [
+      "./gradlew app:dependencyGuardBaseline app-nia-catalog:dependencyGuardBaseline"
```
SimonMarquis, line 18: nit: you don't need to specify the modules. The tasks will be available on all module that have applied the plugin.

```suggestion
      "./gradlew dependencyGuardBaseline"
```

.github/workflows/Build.yaml
```diff
@@ -42,6 +42,9 @@ jobs:
       - name: Check spotless
         run: ./gradlew spotlessCheck --init-script gradle/init.gradle.kts --no-configuration-cache
 
+      - name: Check Dependency Guard
+        run: ./gradlew app:dependencyGuard app-nia-catalog:dependencyGuard
```
SimonMarquis, line 46: Same
```suggestion
        run: ./gradlew dependencyGuard
```

.github/renovate.json
```diff
@@ -12,5 +12,10 @@
       "matchPackageNames": ["com.google.protobuf"],
       "allowedVersions": "<=0.8.19"
     }
-  ]
+  ],
+  "postUpgradeTasks": {
```
SimonMarquis, line 16: It seems like this will work only on self-hosted instances:

> Post-upgrade tasks can only be used on self-hosted Renovate instances.
> https://docs.renovatebot.com/configuration-options/#postupgradetasks

jdkoren: Bummer :cry:

.github/renovate.json
jdkoren: Thanks, that's much cleaner.

.github/renovate.json
jdkoren: Rebased, this diff is obsolete.

dturner:
@jdkoren Looks like we're missing the dependabot configuration to run `dependencyGuardBaseline` for automated updates. 

Also minor point, please don't force push as it messes up the commit and comment history.

.github/workflows/Build.yaml
```diff
@@ -45,13 +45,41 @@ jobs:
...
+        id: dependencyguard_baseline
+        if: steps.dependencyguard_verify.outcome == 'failure' && github.event_name == 'pull_request'
+        run: |
+          ./gradlew dependencyGuardBaseline
+
+      - name: Push new Dependency Guard baselines if available
+        uses: stefanzweifel/git-auto-commit-action@v5
+        if: steps.dependencyguard_baseline.outcome == 'success'
+        with:
+          file_pattern: '*/*.txt'
```
dturner, line 72: Is it worth specifying the exact path for the dependency list file (i.e. `/app/dependencies/prodReleaseRuntimeClasspath.txt`)? Or are there other text files generated by dependencyGuard?

dturner (approved):
Approved with one minor question/suggestion.

.github/workflows/Build.yaml
jdkoren: There are multiple baseline files, one per module per configuration that we've set the plugin to track, so a single PR could update any or all of them at once. We can list them all out, we'd just have to keep the list up to date if we ever modify those configurations.

AFAICT there aren't any other `.txt` files in the repo outside of directories ignored by git, and there's a similar action to update screenshots which also uses the glob syntax, so this seemed reasonable. WDYT of merging as-is and changing to a static list if it causes an issue later?

SimonMarquis: > There are multiple baseline files, one per module per configuration that we've set the plugin to track

This is not how the project is set up at this moment. DependencyGuard is currently configured only on the two app modules `:app` and `app-nia-catalog`

For the file pattern, a simpler glob like this could be better: `**/dependencies/*.txt`

jdkoren: Done

PR #1087, opened 2023-12-09
Remove unnecessary RequiresOptIn opt-in

SimonMarquis:


alexvanyo (approved):
Looks good as per https://kotlinlang.org/docs/whatsnew17.html#stable-opt-in-requirements

PR #1103, opened 2023-12-19
Add comment explaining why the launcher icon colors differ per build

dturner:
When you open the `app/src/debug|benchmark` folders it's not immediately clear why the `colors.xml` files are different. Added a comment explaining this.

### DIFF
--- a/app/src/benchmark/res/values-night/colors.xml
+++ b/app/src/benchmark/res/values-night/colors.xml
@@ -15,6 +15,8 @@
      limitations under the License.
 -->
 <resources>
+    <!-- Allow users to distinguish between build variants by having a different background color
+        for the launcher icon. See https://github.com/android/nowinandroid/pull/989. -->
     <color name="ic_launcher_background_tint">#FFFFFF</color>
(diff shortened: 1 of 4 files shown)


JoseAlcerreca:
Spotless removed your comments (XD)

dturner:
🤣  and CI failed on a completely empty PR. Good times.

SimonMarquis:
I think that moving the comment inside the `<resources>` tag should work.

dturner:
Yeah, I thought that as well, sadly spotless still removes the comment.

SimonMarquis:
What about comments as xml attributes?

```xml
<resources readme="Allow users to distinguish between build variants by having a different background color for the launcher icon. See https://github.com/android/nowinandroid/pull/989.">
```

dturner:
> Yeah, I thought that as well, sadly spotless still removes the comment.

Turns out I was wrong here (could've sworn I tried this and it didn't work). Moving to _after_ the first XML tag solved the problem because Spotless tries to replace everything before the first tag with the license header.

Thanks @SimonMarquis

dturner:
Related #1118

PR #1119, opened 2023-12-21
Replace task exclusions with `-Pandroidx.baselineprofile.skipgeneration`

SimonMarquis:
Available since [1.2.0-alpha15](https://developer.android.com/jetpack/androidx/releases/benchmark#1.2.0-alpha15), and we are currently using 1.2.2

https://github.com/android/nowinandroid/blob/ad15f0137eb196ed332cf1650dcdf41b58fe4ea0/gradle/libs.versions.toml#L20


SimonMarquis:
🔔 @keyboardsurfer this does not seem to be fixed as advertised in the release notes 😕

PR #1140, opened 2024-01-03
Update AGP to 8.3.0

alexvanyo:
Updates to AGP 8.3.0.

This required a couple of changes due to AGP 8.3.0 being a bit stricter with how dependencies are resolved.

f46e506:

Previously, the Roborazzi `ScreenshotHelper` utilities were in `:core:testing`. However, this module is brought in for adding utilities to both unit tests and instrumentation tests.

This change was needed due to the following:

- `ScreenshotHelper` calls into Robolectric utilities
- Therefore, `:core:testing` had a dependency on Robolectric, which meant that Robolectric was being added into instrumentation tests
- This brought in a `jre` guava was being brought into instrumentation tests, which shouldn't be there

0610407:

`kotlinx-coroutines-guava` depends on `jre` guava, which meant that `:sync:work` depended on `jre` guava in the instrumentation tests. Bumping this to a later `android` guava avoids a `jre` guava in instrumentation tests again. I had to add this as a main implementation, otherwise the error from AGP 8.3.0 returned.

1d91231:

A follow-up from investigating more - `:sync:work` doesn't use the `kotlinx-coroutines-guava` dependency in tests anyway, so it is cleanest to just remove it.


lihenggui:
@alexvanyo AGP 8.3 is stable now.

lihenggui:
It will throw this error:
```
> Task :core:designsystem:checkDemoDebugAndroidTestDuplicateClasses FAILED
ASM Instrumentation process wasn't able to resolve some classes, this means that
the instrumented classes might contain corrupt stack frames. Make sure the
dependencies that contain these classes are on the runtime or the provided
classpath. Otherwise, the jvm might fail to load the corrupt classes at runtime
when running in a jvm environment like unit tests.

Classes that weren't resolved:
> androidx.compose.animation.tooling.ComposeAnimatedProperty
> com.google.common.util.concurrent.ListenableFuture
> androidx.window.extensions.embedding.SplitPlaceholderRule
> androidx.window.extensions.embedding.ActivityRule
> androidx.window.extensions.embedding.SplitPairRule


* What went wrong:
Execution failed for task ':core:designsystem:checkDemoDebugAndroidTestDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class com.google.common.util.concurrent.ListenableFuture found in modules guava-31.1-jre (com.google.guava:guava:31.1-jre) and listenablefuture-1.0 (com.google.guava:listenablefuture:1.0)
     
     Go to the documentation to learn how to <a href="d.android.com/r/tools/classpath-sync-errors">Fix dependency resolution errors</a>.
```

alexvanyo:
Thanks for the reminder @lihenggui , updating PR with `8.3.0` to see what failures CI reports.

keyboardsurfer (approved):

JoseAlcerreca (approved):

dturner:
Thanks for doing this @alexvanyo. Do you have any more info you can share on "AGP 8.3.0 being a bit stricter with how dependencies are resolved"? I'm guessing that it's stricter about detecting duplicate classes (e.g. `ListenableFuture`), but are there other changes?

dturner (approved):

alexvanyo:
The behavior change for library modules is reverting to `8.2` behavior in `8.4.0-alpha12`, and this revert will also be cherry-picked to `8.3.x`.

PR #1143, opened 2024-01-06
Remove unnecessary dispatcher switch in DefaultSearchContentsRepository

realOxy:
check out coroutine dispatcher to io is unnecessary.
room has done these job for suspend methods and flow methods.

### DIFF
--- a/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/DefaultSearchContentsRepository.kt
+++ b/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/DefaultSearchContentsRepository.kt
@@ -24,16 +24,12 @@ import com.google.samples.apps.nowinandroid.core.database.model.PopulatedNewsRes
 import com.google.samples.apps.nowinandroid.core.database.model.asExternalModel
 import com.google.samples.apps.nowinandroid.core.database.model.asFtsEntity
 import com.google.samples.apps.nowinandroid.core.model.data.SearchResult
-import com.google.samples.apps.nowinandroid.core.network.Dispatcher
-import com.google.samples.apps.nowinandroid.core.network.NiaDispatchers.IO
-import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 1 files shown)


SimonMarquis:
Hi @realOxy, while this is true that room already handles switching to the proper dispatcher, it will only do it for the room calls.
In this block of code there are other calls that could block the main thread, and create multiple unnecessary context switches.

Here is my understanding of it, ordered by execution:

1. https://github.com/android/nowinandroid/blob/f5b3ae56dcf0022456d061d0c4c121be5a144984/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/DefaultSearchContentsRepository.kt#L50-L55
  `first()` will be executed the `getNewsResources()` on the current context (main), and the following `map` as well.
2. https://github.com/android/nowinandroid/blob/f5b3ae56dcf0022456d061d0c4c121be5a144984/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/DefaultSearchContentsRepository.kt#L49
   Room will then switch to io context to insert, and restore the previous context (main).
3. https://github.com/android/nowinandroid/blob/f5b3ae56dcf0022456d061d0c4c121be5a144984/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/DefaultSearchContentsRepository.kt#L57
  - Room will switch to io context to do `topicDao.getOneOffTopicEntities()`, and restore the previous context (main).
  - `map { it.asFtsEntity() }` will be executed on the current context (main).
  - Room will switch to io context to do `topicFtsDao.insertAll(...)`, and restore the previous context (main)

There is a total of 3 context switches, and 3 operations that will have a negative impact on the current context.

realOxy:
Hi @SimonMarquis , What do you think about the below function "searchContents", the room handled it with io-dispatcher, but the topics instance is created in main-dispatcher.
https://github.com/android/nowinandroid/blob/f5b3ae56dcf0022456d061d0c4c121be5a144984/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/DefaultSearchContentsRepository.kt#L74
https://github.com/android/nowinandroid/blob/f5b3ae56dcf0022456d061d0c4c121be5a144984/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/DefaultSearchContentsRepository.kt#L79-L80
Why we need not flowOn operator to change its dispatcher.

SimonMarquis:
To be honest, I'm not sure how this behaves, especially the combine lambda.

Issue #875, opened 2023-08-01
[Bug]: Move Fakes to a test module


JoseAlcerreca:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

`core/data/src/main/java/com/google/samples/apps/nowinandroid/core/data/repository/fake` has a bunch of fakes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SimonMarquis:
> So, basically we have to move these bunch of repositories to core/data/src/test/java/com/google/samples/apps/nowinandroid/core/data/repository? Isn't?

No, because we probably need to use them in other modules, and you can't depend on test sourceSets of other modules.

FWIW, I think that until we have proper AGP support for testFixtures in Kotlin for Android modules, migrating fakes to a centralized module is a bad practice in terms of modularisation. It encourages devs to merge unrelated code together and/or to create "test" modules for every other module.

To start this migration, one step forward would be to make this `:core:data` a pure JVM module. Only `ConnectivityManagerNetworkMonitor` brings the "Android" dependency.

For reference:
- https://github.com/android/nowinandroid/issues/451

JoseAlcerreca:
> to create "test" modules for every other module

Why is this a bad thing? We could create a `:test` category if it becomes annoying.

SimonMarquis:
This might not have any meaningful impact on small-scale projects, but if you have tens (or hundreds) of modules, and each of them has an extra companion module, this creates unnecessary load on the build process, while `testFixtures` has been designed specifically to solve this issue.
This will also break many annotations that rely on detecting the test classpath like `@VisibleForTesting`, `@TestOnly`, etc., and other similar features like IDE scopes, Lint, etc.
So, I'm not sure we should make this "companion module" approach a general solution to this problem.

SimonMarquis:
Ping @JoseAlcerreca the fakes have been moved to the `:core:data-test` in this PR (now that they are no longer used outside):
- #1144

Issue #709, opened 2023-05-04
[Bug]: Unused binds in TestDataModule


pitoszud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

There are some repository binds in TestDataModule that are not provided or injected anywhere in the app, so you've got some FakeRepository classes sitting idle e.g FakeTopicsRepository

Instead, you are using test doubles data sources e.g  TestNiaNetworkDataSource where you create a fake data source e.g FakeNiaNetworkDataSource from the network module.
I'm not saying this is a bad approach, but it's confusing because your fake repository e.g FakeTopicsRepository injects the same fake data source e.g FakeNiaNetworkDataSource

Why not go with the Hilt @TestInstallIn approach, similar to what you're doing in the NavigationUiTest ?

Thanks
Patryk

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SimonMarquis:
@pitoszud This `TestDataModule` is used in `@HiltAndroidTest` unit-tests and instrumented-tests in the main `:app` module. Without it, the tests would use the real implementations and this is not what we want for tests. Let's close this issue unless you need more info 😉

JackEblan:
> @pitoszud This `TestDataModule` is used in `@HiltAndroidTest` unit-tests and instrumented-tests in the main `:app` module. Without it, the tests would use the real implementations and this is not what we want for tests. Let's close this issue unless you need more info 😉

Can you explain where Fakes where used?

`package com.google.samples.apps.nowinandroid.core.data.repository.fake` I can't find any usage for the classes inside it. Only the TestdataModule is using them and yet I also can't find where did you inject those fakes

SimonMarquis:
The fakes are implicitly used because of this:

https://github.com/android/nowinandroid/blob/f5b3ae56dcf0022456d061d0c4c121be5a144984/core/data-test/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/test/TestDataModule.kt#L36-L41

Hilt will replace the production binding for the test, you can read more on this in https://dagger.dev/hilt/testing.html

JackEblan:
> The fakes are implicitly used because of this:
> 
[...]

PR #1144, opened 2024-01-06
Move `:core:data` fake repositories to `:core:data-test`

SimonMarquis:
Now that these fake implementations are no longer used in the production source code (only in tests), we can safely extract them out of this module. Hopefully, we'll later be able to use Kotlin testFixtures for that.

References:
- #709
- #875


JoseAlcerreca:
Thanks for this @SimonMarquis !

JoseAlcerreca (approved):

SimonMarquis:
@JoseAlcerreca @dturner this PR is ready to be merged 🙇

dturner (approved):

PR #1167, opened 2024-01-21
Disable animations of instrumented tests

SimonMarquis:
https://developer.android.com/reference/tools/gradle-api/com/android/build/api/dsl/TestOptions#animationsDisabled()


SimonMarquis:
Wow 🤯
```
Build / Local tests and APKs (pull_request) Successful in 7m
Build / androidTest (26) (pull_request) Successful in 9m
Build / androidTest (30) (pull_request) Successful in 9m
```

JoseAlcerreca:
This is already done by the `disable-animations` parameter in the action, which is true by default. Any reason to do it here as well?

SimonMarquis:
@JoseAlcerreca does my answer makes sense for this project or should I discard this PR?

JoseAlcerreca (approved):

PR #1173, opened 2024-01-22
Nightly builds with baseline profiles

keyboardsurfer:
Change-Id: I162cd9eca724c30d839add912359482cbce7c68e

**What I have done and why**
Enabled baseline profile generation for new nightly build type.
With KVM, GMD run smoothly and BP generation is not negatively affected any longer.
Enabling nightly builds for this keeps PR builds fast while allowing for baseline profiles to be generated regularly as well.

Enabled Startup Profiles for smaller classes.dex and faster app startup

**Do tests pass?**
- [X] Run local tests on `DemoDebug` variant: `./gradlew testDemoDebug`
- [X] Check formatting: `./gradlew --init-script gradle/init.gradle.kts spotlessApply`


keyboardsurfer:
RFR. @JoseAlcerreca & @dturner PTAL.

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+          commit_message: "🤖 Updates screenshots"
+
+      # Run local tests after screenshot tests to avoid wrong UP-TO-DATE. TODO: Ignore screenshots.
+      - name: Run local tests
+        if: always()
+        run: ./gradlew testDemoDebug :lint:test
+
+      # Replace task exclusions with `-Pandroidx.baselineprofile.skipgeneration` when
+      # https://android-review.googlesource.com/c/platform/frameworks/support/+/2602790 landed in a
+      # release build
```
mlykotom, line 80: maybe this has already landed?

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: 17
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+
+      - name: Build projects before running emulator
+        run: ./gradlew packageDemoDebug packageDemoDebugAndroidTest
```
mlykotom, line 144: All of these steps run unnecessarily twice in the pipeline and might take a long time.
What do you think you would split the pipeline into multiple jobs (which would also increase the duration):

1) build APKs 
2a) tests + screenshot tests
2b) generate baseline profile

You'd need to probably cache the APKs between 1) and 2)

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+      - name: Run instrumentation tests
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: ${{ matrix.api-level }}
+          arch: x86_64
+          disable-animations: true
+          disk-size: 6000M
+          heap-size: 600M
+          script: ./gradlew connectedDemoDebugAndroidTest --daemon
```
mlykotom, line 154: Is there a reason for starting the `--daemon` on this pipeline? IIRC daemons are not worth on CI because every workflow runs on a different runner.

app/src/main/baseline-prof.txt
mlykotom, line 1: Is the reason to remove this because we now get BPs nightly in production? What if I'd like to investigate performance locally? I wouldn't have BPs in my project and thus saw a suboptimal performance.

Could we commit and push the generated BP instead, so that everyone locally can have the benefits?

mlykotom (commented):
I'm generally ok with this, but it would be nice if Don/Jose took a look as they maintain the project.

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
+name: Nightly
```
JoseAlcerreca, line 1: Nightly_BaselineProfiles

JoseAlcerreca: and rename file name

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+on:
+  schedule:
+    - cron:  '0 5 * * *'
```
JoseAlcerreca, line 5: How about something like '29 5 * * *' to balance the load in github's servers?

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+concurrency:
```
JoseAlcerreca, line 7: After reading about `concurrency` I don't understand why you're using it here, so remove it or add a comment explaining

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test_and_apk:
```
JoseAlcerreca, line 12: You don't need this job to generate baseline profiles

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: lint-reports
+          path: '**/build/reports/lint-results-*.html'
+
+      - name: Check badging
+        run: ./gradlew :app:checkProdReleaseBadging
+
+  androidTest:
```
JoseAlcerreca, line 113: baselineProfilesGeneration or similar

.github/workflows/Nightly.yaml
```diff
@@ -0,0 +1,161 @@
...
+      # Run local tests after screenshot tests to avoid wrong UP-TO-DATE. TODO: Ignore screenshots.
+      - name: Run local tests
+        if: always()
+        run: ./gradlew testDemoDebug :lint:test
+
+      # Replace task exclusions with `-Pandroidx.baselineprofile.skipgeneration` when
+      # https://android-review.googlesource.com/c/platform/frameworks/support/+/2602790 landed in a
+      # release build
+      - name: Build all build type and flavor permutations including baseline profiles
+        run: ./gradlew :app:assemble
```
JoseAlcerreca, line 82: Remove most steps and just execute this one

README.md
```diff
@@ -17,6 +17,8 @@ of Android development by providing regular news updates.
 
 The app is currently in development. The `prodRelease` variant is [available on the Play Store](https://play.google.com/store/apps/details?id=com.google.samples.apps.nowinandroid).
 
+Build status: ![example workflow](https://github.com/android/nowinandroid/actions/workflows/Nightly.yaml/badge.svg)
```
JoseAlcerreca, line 20: Remove

JoseAlcerreca (dismissed):
I started the review assuming this nightly was only for BP generation and only realized you were creating a full nightly. I think the `main` build should be our SOT for status, so just make this one about BP.

.github/workflows/NightlyBaselineProfiles.yaml
```diff
@@ -0,0 +1,40 @@
...
+      contents: write
+
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
```
Jaehwa-Noh, line 22:
```suggestion
        uses: gradle/actions/wrapper-validation@v3
```
The latest version is `uses: gradle/actions/wrapper-validation@v3` check this [documentation](https://github.com/gradle/actions/blob/main/docs/wrapper-validation.md#usage)
Is there any reason to use old version validation action?

keyboardsurfer:
@JoseAlcerreca could this be a flaky test?
```
com.google.samples.apps.nowinandroid.ui.SnackbarScreenshotTests > snackbarShown_tablet FAILED
    java.lang.AssertionError at SnackbarScreenshotTests.kt:236
```

If so, could you show me how to fix it or approve the PR so it can be merged?

README.md
JoseAlcerreca: ?

JoseAlcerreca:
Let's merge #1395 first and make sure this one passes

keyboardsurfer:
SGTM

.github/workflows/NightlyBaselineProfiles.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Copy CI gradle.properties
+        run: mkdir -p ~/.gradle ; cp .github/ci-gradle.properties ~/.gradle/gradle.properties
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: 17
+
+      # Replace task exclusions with `-Pandroidx.baselineprofile.skipgeneration` when
```
JoseAlcerreca, line 33: How come you're using it in the other yaml file and not here then?

Also: which task exclusions?

README.md
JoseAlcerreca: I really don't see the point of these badges

app/app.google-services.json.debug
```diff
@@ -0,0 +1,125 @@
+{
```
JoseAlcerreca, line 1: What happens if you remove this file?

.github/workflows/NightlyBaselineProfiles.yaml
keyboardsurfer: Copy-paste issue. This needs to be removed

app/app.google-services.json.debug
keyboardsurfer: IIRC there were issues with the build failing on CI. Let's find out.

Jaehwa-Noh:
I just wonder this, the baseline profile is recommended run on real device, not emulator. Is this CI based Android baseline profile has a value?

keyboardsurfer:
> I just wonder this, the baseline profile is recommended run on real device, not emulator. Is this CI based Android baseline profile has a value?

Measuring performance requires a physical device to create accurate results.
Generating a baseline profile does not require a physical device as it's not about timing but exercising the app and collecting methods and classes in the process.

benchmarks/build.gradle.kts
```diff
@@ -59,6 +59,7 @@ android {
 
 baselineProfile {
     // This specifies the managed devices to use that you run the tests on.
+    managedDevices.clear()
```
yuuki3655, line 62: Why do you need to call clear() here?

.github/workflows/Release.yml
```diff
@@ -37,11 +37,14 @@ jobs:
...
+      - name: Setup Android SDK
+        uses: android-actions/setup-android@v3
 
-      - name: Accept Android licenses
-        run: yes | "$ANDROID_HOME"/cmdline-tools/latest/bin/sdkmanager --licenses || true
+      - name: Accept licenses
+        run: yes | sdkmanager --licenses || true
+
+      - name: Setup AVD for GMD use
+        run: yes | sdkmanager "system-images;android-33;aosp_atd;x86_64" || true
```
yuuki3655, line 47: GMD will download system images for you, so you don't have to install it manually here.

benchmarks/build.gradle.kts
keyboardsurfer: There are several devices added in a plugin and we only want to run BP generation only a single device

.github/workflows/Release.yml
keyboardsurfer: Good point. Removed

.github/workflows/Build.yaml
```diff
@@ -101,19 +114,22 @@ jobs:
...
       # Run local tests after screenshot tests to avoid wrong UP-TO-DATE. TODO: Ignore screenshots.
-      - name: Run local tests and create report
-        if: always()
+      - name: Run local tests
         run: ./gradlew testDemoDebug :lint:test
-      # Replace task exclusions with `-Pandroidx.baselineprofile.skipgeneration` when
-      # https://android-review.googlesource.com/c/platform/frameworks/support/+/2602790 landed in a
-      # release build
+
+      - name: Setup GMD
```
dturner, line 120: Is setting up GMD necessary in the `Build` workflow?

keyboardsurfer: Yes. That's the part where the [GH Action](https://github.com/android/nowinandroid/actions?query=branch%3Abw%2FbpGenOnKvm) failed consistently.

yuuki3655: This is a workaround for GHA. Gradle runs tasks in parallel up to the number specified by the `--max-workers` flag. The GHA runner we're using doesn't have enough processing power to run an emulator while other build tasks, like the Kotlin compiler and aapt, are running. This causes the emulator process to crash. By running the setup task in a separate step, we can ensure that it runs while no other tasks are running in parallel. Alternatively, we could upgrade the GHA runner to one with more cores.

dturner (approved):

lihenggui:
It seems that not enabling the KVM before setting up GMD causes this problem.

https://github.com/android/nowinandroid/actions/runs/10953338005/job/30413441493

```
* What went wrong:
Execution failed for task ':benchmarks:pixel6Api33Setup'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.ManagedDeviceInstrumentationTestSetupTask$ManagedDeviceSetupRunnable
   > com.android.build.gradle.internal.AvdSnapshotHandler$EmulatorSnapshotCannotCreatedException: Gradle was not able to complete device setup for: dev33_default_x86_64_Pixel_6
     The emulator failed to open the managed device to generate the snapshot.
     This is because the emulator closed unexpectedly (1),
     try updating the emulator and ensure a device can be run from Android Studio.
```

keyboardsurfer:
@lihenggui the issue was caused by GMD requiring a lot more processing power than what the GitHub runners offer in the free tier. So the relevant step is this one

```
- name: Setup GMD
        run: ./gradlew :benchmarks:pixel6Api33Setup
          --info
          -Pandroid.experimental.testOptions.managedDevices.emulator.showKernelLogging=true
          -Pandroid.testoptions.manageddevices.emulator.gpu="swiftshader_indirect"
```

Which has to be executed before the entire build is kicked off.

PR #1187, opened 2024-01-30
Remove TimeZoneBroadcastReceiver from each item

mlykotom:

**What I have done and why**
We use TimeZone broadcast receiver registered for each item composed on screen.
This is not necessary and takes precious time when composing an item as it takes **~1ms - 3ms** contributing to jank.

Trace from before: 
https://ui.perfetto.dev/#!/?s=2613b122d01fb9a070616ad3adf9157f1d034f1266d0bf0a2fe224efc5a50a4f

image

Trace after:
https://ui.perfetto.dev/#!/?s=ff59121daa4469dda32b9c94db3e1223bf89e35a4f15538dd374d254b8406d18

The change I propose is to have `TimeZoneMonitor` similarly to `NetworkMonitor` and subscribe to this only once.

**Do tests pass?**


core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/LocalTimeZone.kt
```diff
@@ -0,0 +1,26 @@
...
+package com.google.samples.apps.nowinandroid.core.ui
+
+import androidx.compose.runtime.compositionLocalOf
+import java.time.ZoneId
+
+/**
+ * TimeZone that can be provided with the TimeZoneMonitor.
+ * This way, it's not needed to pass every single composable the time zone to show in UI.
+ */
+val LocalTimeZone = compositionLocalOf { ZoneId.systemDefault() }
```
alexvanyo, line 26: Nit: maybe this should be `LocalZoneId`? to match the type of the composition local?

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
```diff
@@ -0,0 +1,94 @@
...
+                override fun onReceive(context: Context, intent: Intent) {
+                    if (intent.action != Intent.ACTION_TIMEZONE_CHANGED) return
+
+                    val zoneIdFromIntent = if (VERSION.SDK_INT < VERSION_CODES.R) {
+                        null
+                    } else {
+                        // Starting Android R we also get the new TimeZone.
+                        intent.getStringExtra(Intent.EXTRA_TIMEZONE)?.let { zoneId ->
+                            // We need to convert it from java.util.Timezone to java.time.ZoneId
+                            ZoneId.of(zoneId, ZoneId.SHORT_IDS)
```
alexvanyo, line 74: Nit: The direct conversion method might be clearer here?

```suggestion
                            java.util.TimeZone.getTimeZone(zoneId).toZoneId()
```

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
```diff
@@ -0,0 +1,94 @@
...
+                        }
+                    }
+
+                    // If there isn't a zoneId in the intent, fallback to the systemDefault, which should also reflect the change
+                    trySend(zoneIdFromIntent ?: ZoneId.systemDefault())
+                }
+            }
+
+            trace("TimeZoneBroadcastReceiver.register") {
```
alexvanyo, line 83: In Android Studio, I'm seeing this `trace` method show up as deprecated

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
```diff
@@ -0,0 +1,94 @@
...
+                context.registerReceiver(receiver, IntentFilter(Intent.ACTION_TIMEZONE_CHANGED))
+            }
+
+            awaitClose {
+                context.unregisterReceiver(receiver)
+            }
+        }
+            .flowOn(ioDispatcher)
```
alexvanyo, line 91: A `callbackFlow` has a fixed buffer size by default, but since we only care about the most recent value I think we can call `.conflate()` here to ensure that the `trySend`s never fail.

```suggestion
            .conflate()
            .flowOn(ioDispatcher)
```

app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -126,17 +133,25 @@ class MainActivity : ComponentActivity() {
...
 
-            CompositionLocalProvider(LocalAnalyticsHelper provides analyticsHelper) {
+            val appState = rememberNiaAppState(
+                windowSizeClass = calculateWindowSizeClass(this),
+                networkMonitor = networkMonitor,
+                userNewsResourceRepository = userNewsResourceRepository,
+                timeZoneMonitor = timeZoneMonitor,
+            )
+
+            val currentTimeZone by appState.currentTimeZone.collectAsState()
```
alexvanyo, line 143: Should this be `collectAsStateWithLifecycle`, to unregister the broadcast receiver if the app is backgrounded?

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/LocalTimeZone.kt
mlykotom: Yeah, that makes sense 👍🏼

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
mlykotom: Yeah, I thought about it, but I thought it's pointless to create `TimeZone` object when we immediatelly convert it to `ZoneId`. The implementation of `toZoneId()` is basically `ZoneId.of(getID(), ZoneId.SHORT_IDS)`.

But I can change it, maybe for readability it's better.

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
mlykotom: It shows deprecation in favor of `androidx.tracing`, but it's not added as a dependency to various modules.
WDYT of adding `implementation(libs.androidx.tracing.ktx)` to the convention plugins as a dependency, so we have it consistent everywhere?

app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
mlykotom: Yeah, sounds good.

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.flow.shareIn
+import java.time.ZoneId
```
alexvanyo, line 39: Ah just realized this - elsewhere in the project we use `kotlinx.datetime` - should we prefer that here instead of using `java.time` directly?

mlykotom: Oh yes! I forgot it could exist. This simplifies things. 👍🏼

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
mlykotom: Ok, added the dependencies to android library + feature and refactored to be used everywhere. We need to use `alpha` version of the library to be able to use it with composables, but there aren't any changes in the library with the alpha (actually filed a bug to stabilize it)

app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -126,17 +133,25 @@ class MainActivity : ComponentActivity() {
...
 
-            CompositionLocalProvider(LocalAnalyticsHelper provides analyticsHelper) {
+            val appState = rememberNiaAppState(
+                windowSizeClass = calculateWindowSizeClass(this),
+                networkMonitor = networkMonitor,
+                userNewsResourceRepository = userNewsResourceRepository,
+                timeZoneMonitor = timeZoneMonitor,
+            )
+
+            val currentZoneId by appState.currentTimeZone.collectAsStateWithLifecycle()
```
alexvanyo, line 143: Nit: this variable name should likely be `currentTimeZone` with the API update

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
```diff
@@ -0,0 +1,98 @@
...
+@Singleton
+internal class TimeZoneBroadcastMonitor @Inject constructor(
+    @ApplicationContext private val context: Context,
+    @ApplicationScope appScope: CoroutineScope,
+    @Dispatcher(IO) private val ioDispatcher: CoroutineDispatcher,
+) : TimeZoneMonitor {
+
+    override val currentTimeZone: SharedFlow<TimeZone> =
+        callbackFlow {
+            // Send the default time zone first.
```
alexvanyo, line 62: Nit: this is probably extremely unlikely, but since we are registering the receiver asynchronously, and reporting the initial time zone first, I think there's technically a race condition if the timezone happens to change after the initial `trySend`, but before the receiver gets registered.

I think a fix would be to call this `trySend` after we register the receiver, but that might result in a longer length of time before the `callbackFlow` emits anything. It should be safe if we call `trySend` twice, once here and once after registering the receiver? And if we wanted to avoid duplicate emissions, we could add a `distinctUntilChanged`

alexvanyo (approved):
Looks great, just a last couple nitpicks

core/data-test/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/test/DefaultZoneIdTimeZoneMonitor.kt
```diff
@@ -0,0 +1,27 @@
...
+package com.google.samples.apps.nowinandroid.core.data.test
+
+import com.google.samples.apps.nowinandroid.core.data.util.TimeZoneMonitor
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.datetime.TimeZone
+import javax.inject.Inject
+
+class DefaultZoneIdTimeZoneMonitor @Inject constructor() : TimeZoneMonitor {
+    override val currentTimeZone: Flow<TimeZone> = flowOf(TimeZone.of("Europe/Warsaw"))
```
SimonMarquis, line 26: I'm wondering, shouldn't we use the system default instead of an hardcoded value?
```kotlin
TimeZone.currentSystemDefault()
```

Same question for `TestTimeZoneMonitor.kt`.

alexvanyo: Using a fixed time-zone makes sense to me for reproducability - we previously ran into an issue where screenshot tests could fail due to using the local time zone instead of a fixed time zone: https://github.com/android/nowinandroid/pull/876#pullrequestreview-1587229309

SimonMarquis: That makes sense, thanks for the link :)

jdkoren:
I can't seem to reproduce the error given in CI. @mlykotom it may be that you need to merge `main` and run `./gradlew dependencyGuardBaseline`.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
mlykotom: Fixed

core/data-test/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/test/DefaultZoneIdTimeZoneMonitor.kt
mlykotom: Thanks for verifying!

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
mlykotom: I added it there, but I think we're overcomplicating it a bit here 😀 .

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/TimeZoneMonitor.kt
```diff
@@ -0,0 +1,98 @@
...
+                    if (intent.action != Intent.ACTION_TIMEZONE_CHANGED) return
+
+                    val zoneIdFromIntent = if (VERSION.SDK_INT < VERSION_CODES.R) {
+                        null
+                    } else {
+                        // Starting Android R we also get the new TimeZone.
+                        intent.getStringExtra(Intent.EXTRA_TIMEZONE)?.let { timeZoneId ->
+                            // We need to convert it from java.util.Timezone to java.time.ZoneId
+                            val zoneId = ZoneId.of(timeZoneId, ZoneId.SHORT_IDS)
+                            TimeZone.of(zoneId.id)
```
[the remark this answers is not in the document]
mlykotom: Works fine, found a `toKotlinTimeZone()` extension function that's a bit more efficient, because doesn't have to parse `ZoneId` twice

alexvanyo (approved):

PR #1190, opened 2024-01-31
OkHttp lazy initialization

mlykotom:
**What I have done and why**
Move Coil's initialization with OkHttp to a background thread.
Without this change, Coil and OkHttp is initialized with the first composed image, which causes skipped frames, because the initialization can take up to 100ms.

image

**Do tests pass?**


SimonMarquis:
100ms 🤯 can you measure this on release builds as well? Debug builds are notoriously slow because they are not optimized and contain debugging features that slows down everything (especially Compose related code)

app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
```diff
@@ -41,7 +40,7 @@ class NiaApplication : Application(), ImageLoaderFactory {
         // Initialize Sync; the system responsible for keeping data in the app up to date.
         Sync.initialize(context = this)
         profileVerifierLogger()
+        // We set immediately Coil's image loader factory to prevent initialization with the first image.
+        Coil.setImageLoader(imageLoaderAsyncFactory)
```
yschimke, line 44: I don't actually understand why you need this call, and to remove ImageLoaderFactory.

The `imageLoaderAsyncFactory` will cause your early init of the loader. The call to newImageLoader() could call your AsyncFactory.

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
```diff
@@ -0,0 +1,59 @@
...
+ */
+class ImageLoaderAsyncFactory @Inject constructor(
+    @ApplicationScope
+    appScope: CoroutineScope,
+    private val imageLoader: dagger.Lazy<ImageLoader>,
+) : ImageLoaderFactory {
+    private lateinit var asyncNewImageLoader: Deferred<ImageLoader>
+
+    init {
+        appScope.launch {
```
yschimke, line 47: Should this be `appScope.launch(Dispatchers.Default)`? or is that implicit?

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
```diff
@@ -0,0 +1,59 @@
...
+            // Initializing asynchronously, but start immediately
```
yschimke, line 48: TIOLO: this didn't really capture why you want an async here.  Feel free to ignore.

```suggestion
            // Initialize immediately, but need a Deferred for callers
```

app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
mlykotom: If we leave it as it was, Coil will request the `newImageLoader` with the first loading image on the main thread, which I think is way too late. There's a huge time window we can initialize it before that happens (and on a background thread).

yschimke (approved):

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
mlykotom: This is implicit by the `@ApplicationScope`, but I do agree it's a bit weird to rely on that.

mlykotom:
> 100ms 🤯 can you measure this on release builds as well? Debug builds are notoriously slow because they are not optimized and contain debugging features that slows down everything (especially Compose related code)

I'm always measuring on release (benchmark) build 😁 . I think on the new versions of Android might not be as problematic, but on Pixel 6 it takes ~15ms.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
yschimke: It doesn't seem to change timing. Injection will cause it to kick that off at same point in time.

mlykotom: I guess I misunderstood what you meant. Yes, we could keep this class implementing the `ImageLoagerFactory` and inject the `imageLoaderAsyncFactory`. I thought to change it because someone might ask why we have 2 classes implementing `ImageLoaderFactory`. 

This way I thought it's more clear that we set the factory immediatelly, but it is doing async init internally.

I also thought about the fact that when dagger injects the `imageLoaderAsyncFactory` it comes with a side effect of starting the initialization (in the `init` of the class). I was thinking this is not great and had a `initialize()` function as well, which was called in `Application.onCreate` .. but then I realised I just overcomplicate it.

I can change it back :D

yschimke: Your call.

yschimke: Why not upstream this to Coil cleanly? Update the sample to use a provider, avoid early init.

Have a simple method to kick off init in background? Perhaps conditionally, say based on process lifecycle in case a service is run?

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
```diff
@@ -0,0 +1,59 @@
...
+    appScope: CoroutineScope,
+    private val imageLoader: dagger.Lazy<ImageLoader>,
+) : ImageLoaderFactory {
+    private lateinit var asyncNewImageLoader: Deferred<ImageLoader>
+
+    init {
+        appScope.launch {
+            // Initialize immediately, but need a Deferred for callers
+            asyncNewImageLoader = async { imageLoader.get() }
+        }
```
hoc081098, line 50: asyncNewImageLoader can be a val.

```suggestion
        asyncNewImageLoader = appScope.async {
            // Initialize immediately, but need a Deferred for callers
            imageLoader.get()
        }
```

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * This runBlocking here is on purpose to prevent any unfinished Coil initialization.
+     * Most likely this will be already initialized by the time we want to show an image on the screen.
+     */
+    override fun newImageLoader() =
+        trace("NiaImageLoader.runBlocking") { runBlocking { asyncNewImageLoader.await() } }
```
hoc081098, line 58: Checking `asyncNewImageLoader.isCompleted` before calling `runBlocking` to avoid blocking the main thread in case asyncNewImageLoader has been completed

```suggestion
    override fun newImageLoader() =
        trace("NiaImageLoader.runBlocking") {
            if (asyncNewImageLoader.isCompleted) asyncNewImageLoader.getCompleted() 
            else runBlocking { asyncNewImageLoader.await() }
        }
```

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
mlykotom: This is great! ❤️  I was wondering if this works exactly the same, but wasn't sure and wondered if that'd be called out in the review.

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
mlykotom: Great! Thanks for the suggestion!

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/di/NetworkModule.kt
```diff
@@ -51,16 +52,18 @@ internal object NetworkModule {
...
-        .addInterceptor(
-            HttpLoggingInterceptor()
-                .apply {
-                    if (BuildConfig.DEBUG) {
-                        setLevel(HttpLoggingInterceptor.Level.BODY)
-                    }
-                },
-        )
-        .build()
+    fun okHttpCallFactory(): Call.Factory = trace("NiaOkHttpClient") {
```
yschimke, line 55: Do you see this in traces via NiaNetworkDataSource? I wonder whether Retrofit and OkHttp can be deferred in the case you have cached content if so?

mlykotom:
@hoc081098 could you pls sign the CLA so that we can merge this PR? Thanks you :blush

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * This runBlocking here is on purpose to prevent any unfinished Coil initialization.
+     * Most likely this will be already initialized by the time we want to show an image on the screen.
+     */
+    override fun newImageLoader() =
+        trace("NiaImageLoader.runBlocking") {
+            if (asyncNewImageLoader.isCompleted) {
+                asyncNewImageLoader.getCompleted()
+            } else {
+                runBlocking { asyncNewImageLoader.await() }
```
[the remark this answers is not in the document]
yschimke: It might be worth putting it behind an async ImageLoader decorator. I mean you can do exactly this.

hoc081098:
> @hoc081098 could you pls sign the CLA so that we can merge this PR? Thank you 😊

I have signed the CLA for "hoc081098@gmail.com". Don't know why the check failed 😂?

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/di/NetworkModule.kt
```diff
@@ -72,20 +75,21 @@ internal object NetworkModule {
     @Provides
     @Singleton
     fun imageLoader(
-        okHttpCallFactory: Call.Factory,
+        // We specifically request dagger.Lazy here, so that it's not instantiated from Dagger.
+        okHttpCallFactory: dagger.Lazy<Call.Factory>,
         @ApplicationContext application: Context,
-    ): ImageLoader = ImageLoader.Builder(application)
-        .callFactory(okHttpCallFactory)
```
colinrtwhite, line 78: It looks like this is the culprit of the slow initialization, as the `OkHttpClient` is created eagerly instead of using `ImageLoader.Builder.callFactory(() -> Call.Factory)`, which accepts a callback that's invoked on a background thread. It looks like this PR also handles that fix using a `dagger.Lazy`.

I'd recommend migrating to `ImageLoader.Builder.callFactory { }` instead of creating the `ImageLoader` asynchronously, as that should fix the performance issue. I included a quick benchmark [here](https://github.com/coil-kt/coil/issues/2097#issuecomment-1925892114). Also added more context [here](https://github.com/coil-kt/coil/issues/2097#issuecomment-1922061816)!

app/src/main/java/com/google/samples/apps/nowinandroid/util/ImageLoaderAsyncFactory.kt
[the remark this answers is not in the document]
colinrtwhite: I'd recommend against eagerly creating your `ImageLoader` on a background thread as the main slow parts of an `ImageLoader` (`DiskCache` and `OkHttpClient`) are already initialized lazily on a background thread by default. If those are passed lazily then an `ImageLoader` takes about 0.1 milliseconds to initialize in [my tests](https://github.com/coil-kt/coil/issues/2097#issuecomment-1925892114).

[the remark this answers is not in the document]
mlykotom: As per https://github.com/coil-kt/coil/issues/2097 we might not need to have the async pre-loader, but Coil needs to register the system callbacks on a `Default` thread instead. Then, the init will be fast and won't jank loading first image.

mlykotom:
> > @hoc081098 could you pls sign the CLA so that we can merge this PR? Thank you 😊
> 
> I have signed the CLA for "[hoc081098@gmail.com](mailto:hoc081098@gmail.com)". Don't know why the check failed 😂?

Pls check the CLA failure for more info. Maybe you used a different email? Not sure what went wrong then.

mlykotom:
Rebased on `main`, so it's not dependent on the TimeZone changes PR.
Also, I removed the `AsyncImageLoaderFactory`, because it will be fixed directly in Coil (https://github.com/coil-kt/coil/issues/2097), so once 2.6.0 is shipped, we'll get the improvement automatically.

alexvanyo (approved):
Awesome! If I'm understanding it right - we are now lazily creating the `okHttpCallFactory`, both for our own normal network calls, and once the Coil changes are released, for Coil as well.

Code looks good, maybe the title of the PR should be "Lazily initialize okHttpCallFactory" or something similar if we aren't affecting Coil too directly in this PR?

mlykotom:
We're initializing OkHttp lazily async both for our Retrofit and Coil implementation. So with this PR we have half of the improvement. The remaining work for Coil library is to lazily initialize system callbacks on a worker thread.

Also, renamed the PR.

Thanks for all reviews! ❤️

PR #1088, opened 2023-12-09
Reduce overall build complexity

SimonMarquis:
- Reduce the visibility of multiple Hilt `Module`s and implementations of public interfaces
- Correctly configure the visibility of dependencies:
  - `api` when it's part of it's public `api`
  - `implementation` when it's part of it's internal `implementation`
- Remove unnecessary dependencies in build.gradle.kts files
- Remove unnecessary dependencies provided by plugins
- Remove unnecessary applied plugins (i.e. roborazzi)
- Sort dependencies in `build.gradle.kts`
- Delete unused entries in `libs.versions.toml`

This could still be improved by splitting the "tests" modules (like `:core:testing`) that currently exposes way too many dependencies, and by merging `testFixtures` sourceSets as explained in #451.

---

Here is the before/after for the build graph:

```diff
- GraphStatistics(modulesCount=25, edgesCount=102, height=7, longestPath=':app -> :feature:search -> :feature:bookmarks -> :core:ui -> :core:domain -> :core:data -> :core:database -> :core:model')
+ GraphStatistics(modulesCount=25, edgesCount=71, height=6, longestPath=':app -> :feature:search -> :feature:foryou -> :core:domain -> :core:data -> :core:database -> :core:model')
```

| Before | After |
|---|---|
| graph | graph (1) |

And here are some more useful stats, thanks to https://github.com/cdsap/ProjectGraphMetrics:

```diff
 ┌───────────────────────┬──────────┬──────────┬───────────────────────┬────────┐
 │ Module                │ Indegree │ Outdegre │ BetweennessCentrality │ Height │
 ├───────────────────────┼──────────┼──────────┼───────────────────────┼────────┤
-│ :app                  │        0 │       13 │                   0.0 │      7 │
+│ :app                  │        0 │       13 │                   0.0 │      6 │
 ├───────────────────────┼──────────┼──────────┼───────────────────────┼────────┤
-│ :app-nia-catalog      │        0 │        2 │                   0.0 │      5 │
+│ :app-nia-catalog      │        0 │        2 │                   0.0 │      2 │
 ├───────────────────────┼──────────┼──────────┼───────────────────────┼────────┤
-│ :core:analytics       │       11 │        0 │                   0.0 │      0 │
+│ :core:analytics       │        5 │        0 │                   0.0 │      0 │
 ├───────────────────────┼──────────┼──────────┼───────────────────────┼────────┤
-│ :core:common          │       15 │        0 │                   0.0 │      0 │
[...]

PR #1199, opened 2024-02-02
Restores screenshot tests in foryou

JoseAlcerreca:
Fixes regression from #1088


JoseAlcerreca:
Looks like the png files needed a refresh, but I don't see the difference

dturner (approved):

PR #1207, opened 2024-02-06
Remove legacy benchmark build type

keyboardsurfer:
Now that we have the baseline profile gradle plugin, there is no need for a benchmark build type.
The plugin does the setup and configuration and runs against the correct app flavor.

Change-Id: I7df90c982e7a27c2fd2051ada441038ac1c76484

**What I have done and why**
Include a summary of what your pull request contains, and why you have made these changes.

**Do tests pass?**

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -109,13 +109,9 @@ jobs:
       - name: Build all build type and flavor permutations
         run: ./gradlew :app:assemble :benchmarks:assemble
           -x pixel6Api33ProdNonMinifiedReleaseAndroidTest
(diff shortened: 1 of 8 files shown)


mlykotom:
I'm not sure we can/want to just remove the `applicationIdSuffix` when we have infrastructure set up around this?

keyboardsurfer:
I mean, it just builds and doesn't cause any realistic issues other than not having a specific benchmark build type any longer.
Also this way the configuration is easier and relying on the Gradle plugin rather than our legacy solution.

mlykotom (approved):
I don't like that I won't be able to have multiple build types installed on my device, but I guess it's less confusing than duplicated build types.

PR #1210, opened 2024-02-08
Update Lifecycle to 2.7.0. Use new LifecycleEventEffect.

dturner:
**What I have done and why**

- Updated AndroidX Lifecycle to version 2.7.0 ([release notes](https://developer.android.com/jetpack/androidx/releases/lifecycle#2.7.0)). 
- Replaced a `DisposableEffect` which was observing lifecycle events with the newly introduced `LifecycleEventEffect` ([docs](https://developer.android.com/topic/libraries/architecture/compose)) which does the same thing with less code. 

Before:
```
val lifecycleOwner = LocalLifecycleOwner.current
DisposableEffect(lifecycleOwner) {
    val observer = LifecycleEventObserver { _, event ->
       if (event == Lifecycle.Event.ON_STOP) {
            clearUndoState()
        }
    }
    lifecycleOwner.lifecycle.addObserver(observer)
    onDispose { lifecycleOwner.lifecycle.removeObserver(observer) }
}
```

After:

```
LifecycleEventEffect(Lifecycle.Event.ON_STOP) {
   clearUndoState()
}
```

Also added an instrumented test to verify the behavior.


feature/bookmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
```diff
@@ -128,15 +126,8 @@ internal fun BookmarksScreen(
...
-    DisposableEffect(lifecycleOwner) {
-        val observer = LifecycleEventObserver { _, event ->
-            if (event == Lifecycle.Event.ON_STOP) {
-                clearUndoState()
-            }
-        }
-        lifecycleOwner.lifecycle.addObserver(observer)
-        onDispose { lifecycleOwner.lifecycle.removeObserver(observer) }
+    LifecycleEventEffect(Lifecycle.Event.ON_STOP) {
+        clearUndoState()
```
tunjid, line 130: niceeee

tunjid (approved):

hoc081098 (commented):
Nice

feature/bookmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt
ianhanniballake: Looks like a great use case for this effect!

ianhanniballake (approved):

ianhanniballake (approved):
Looks like you will need to run `./gradlew :feature:bookmarks:spotlessApply` to fix spotless errors on `BookmarksScreenTest`

PR #1228, opened 2024-02-22
Update BookmarksViewModel.kt

lso033c:

**What I have done and why**
Include a summary of what your pull request contains, and why you have made these changes.

**Do tests pass?**
- [ ] Run local tests on `DemoDebug` variant: `./gradlew testDemoDebug`
- [ ] Check formatting: `./gradlew --init-script gradle/init.gradle.kts spotlessApply`

- [ ] [Sign the CLA](https://cla.developers.google.com/)
- [ ] Run `./tools/setup.sh`

### DIFF
--- a/feature/bookmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksViewModel.kt
+++ b/feature/bookmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksViewModel.kt
@@ -55,7 +55,7 @@ class BookmarksViewModel @Inject constructor(
             )
 
     fun removeFromSavedResources(newsResourceId: String) {
-        viewModelScope.launch {
+        viewModelScope.launch(Dispatchers.Default) {
             shouldDisplayUndoBookmark = true
             lastRemovedBookmarkId = newsResourceId
             userDataRepository.updateNewsResourceBookmark(newsResourceId, false)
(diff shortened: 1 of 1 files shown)


feature/bookmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksViewModel.kt
```diff
@@ -55,21 +55,21 @@ class BookmarksViewModel @Inject constructor(
             )
 
     fun removeFromSavedResources(newsResourceId: String) {
-        viewModelScope.launch {
+        viewModelScope.launch(Dispatchers.Default) {
             shouldDisplayUndoBookmark = true
             lastRemovedBookmarkId = newsResourceId
             userDataRepository.updateNewsResourceBookmark(newsResourceId, false)
```
hoc081098, line 61: This PR does not follow the Coroutines Best Practice doc :)).
Repository should do switch context instead ViewModel

hoc081098: https://developer.android.com/kotlin/coroutines/coroutines-best-practices#main-safe

Suspend functions should be main-safe, meaning they're safe to call from the main thread. If a class is doing long-running blocking operations in a coroutine, it's in charge of moving the execution off the main thread using withContext. This applies to all classes in your app, regardless of the part of the architecture the class is in.

```kotlin
class NewsRepository(private val ioDispatcher: CoroutineDispatcher) {

    // As this operation is manually retrieving the news from the server
    // using a blocking HttpURLConnection, it needs to move the execution
    // to an IO dispatcher to make it main-safe
    suspend fun fetchLatestNews(): List<Article> {
        withContext(ioDispatcher) { /* ... implementation ... */ }
    }
}

// This use case fetches the latest news and the associated author.
class GetLatestNewsWithAuthorsUseCase(
    private val newsRepository: NewsRepository,
    private val authorsRepository: AuthorsRepository
) {
    // This method doesn't need to worry about moving the execution of the
    // coroutine to a different thread as newsRepository is main-safe.
    // The work done in the coroutine is lightweight as it only creates
    // a list and add elements to it
    suspend operator fun invoke(): List<ArticleWithAuthor> {
        val news = newsRepository.fetchLatestNews()

        val response: List<ArticleWithAuthor> = mutableEmptyList()
        for (article in news) {
            val author = authorsRepository.getAuthor(article.author)
            response.add(ArticleWithAuthor(article, author))
        }
        return Result.Success(response)
    }
}
```

This pattern makes your app more scalable, as classes calling suspend functions don't have to worry about what Dispatcher to use for what type of work. This responsibility lies in the class that does the work.

Issue #1045, opened 2023-11-18
[Bug]: CI allows merging PRs with failing checks


SimonMarquis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

#### Initial push, all checks are run:

![Screenshot 2023-11-18 093425](https://github.com/android/nowinandroid/assets/1921278/f040c588-fd64-4ba0-804d-046df477ec75)

#### Because commit contains UI changes, an automated commit `🤖 Updates screenshots` is added, resetting all checks but `cla/google`

![image](https://github.com/android/nowinandroid/assets/1921278/33f46339-da77-41fd-aa7e-96f38d84826c)

#### Resulting in `✅ All checks have passed`

----

For reference, this is what happened in #942 [breaking the main branch after merge](https://github.com/android/nowinandroid/issues/1041).
And #1044 for a simpler example.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

JoseAlcerreca:
I've been looking for a way to work around this, but to avoid recursion GHA doesn't let workflows start other worflows. [SO thread](https://stackoverflow.com/questions/73631383/triggering-a-workflow-run-after-an-action-commits).

SimonMarquis:
You can either do that with PAT or a GitHub app (Google probably already has such bot available on other Google repos).

PR #1230, opened 2024-02-26
Replace Accompanist TestHarness with DeviceConfigurationOverride

### LINKED ISSUE #1201: Replace testharness with `DeviceConfigurationOverride`

### Describe the problem

The replacement for accompanist/testharness is `DeviceConfigurationOverride`, which is available in `ui-test` `1.7.0-alpha01`.

### Describe the solution

We should migrate our usages.

### Additional context

_No response_

### Code of Conduct

dturner:
[Compose UI 1.7.0-alpha02](https://developer.android.com/jetpack/androidx/releases/compose-ui#1.7.0-alpha02) added the DeviceConfigurationOverride API which replaces the Accompanist TestHarness API. 

This mostly involves changing

```
TestHarness(size = DpSize(600.dp, 400.dp))
```
to
```
DeviceConfigurationOverride(
    DeviceConfigurationOverride.ForcedSize(DpSize(600.dp, 400.dp)),
)
```
[Full migration guide here](https://google.github.io/accompanist/testharness/#migration).

I have: 
- Replaced all instances of `TestHarness` with `DeviceConfigurationOverride`
- Removed the testharness library

Fixes #1201


app/src/testDemo/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppScreenSizesScreenshotTests.kt
```diff
@@ -139,7 +140,9 @@ class NiaAppScreenSizesScreenshotTests {
             CompositionLocalProvider(
                 LocalInspectionMode provides true,
             ) {
-                TestHarness(size = DpSize(width, height)) {
+                DeviceConfigurationOverride(
```
dturner, line 143: There seems to be behaviour change here. Previously, the screenshot would be the same size as the `BoxWithConstraints` but now it's fixed at 1500x1404px, top left aligned.

alexvanyo: Interesting, we are generating the screenshots from the root, and it looks like the root in `DeviceConfigurationOverride` is filling up the whole screen, instead of behaving like the one in testharness and more closely wrapping the content.

alexvanyo: Can you report a bug for `DeviceConfigurationOverride.ForcedSize`? Definitely agreed that `ForcedSize` shouldn't fill up the max space available if the requested size is smaller than that.

In the interim, we should be able to preserve behavior if we take a screenshot of the node specifically corresponding to the app, instead of the root.

dturner: Bug filed here: https://b.corp.google.com/issues/327149166 and screenshot updated to use the app node instead of the root.

app/src/testDemo/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppScreenSizesScreenshotTests.kt
```diff
@@ -135,27 +137,31 @@ class NiaAppScreenSizesScreenshotTests {
     }
 
     private fun testNiaAppScreenshotWithSize(width: Dp, height: Dp, screenshotName: String) {
+        // TODO: Remove this when https://b.corp.google.com/issues/327149166 is fixed
```
alexvanyo, line 140: Avoiding the corp link is nicer for everyone externally when it is a public issue.

```suggestion
        // TODO: Remove this when https://issuetracker.google.com/issues/327149166 is fixed
```

alexvanyo (approved):

app/src/testDemo/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppScreenSizesScreenshotTests.kt
```diff
@@ -139,29 +139,32 @@ class NiaAppScreenSizesScreenshotTests {
     }
 
     private fun testNiaAppScreenshotWithSize(width: Dp, height: Dp, screenshotName: String) {
+        // TODO: Remove this when https://issuetracker.google.com/issues/327149166 is fixed
```
alexvanyo, line 142: This issue is fixed in `1.7.0-alpha04`, so we should be able to remove this workaround now.

dturner: Updated to `1.7.0-alpha05` and workaround removed. Just need to verify that the screenshots haven't changed.

alexvanyo (approved):

jdkoren:
Instrumented tests fail when I check out the code and run them locally. Is there a way to kick Github Actions? "Show all checks" is only showing the `cla/google` check, I don't see the one which actually builds & tests.

SimonMarquis:
@jdkoren this is because of this issue:
- https://github.com/android/nowinandroid/issues/1045
If you want to check previous builds, you can click the ❌ next to the commit id in this page.
For example, the latest is:
- https://github.com/android/nowinandroid/pull/1230/commits/739dc0098727cf12cd480997e6cadd267ed027ea
- https://github.com/android/nowinandroid/actions/runs/8466150965?pr=1230

PR #1234, opened 2024-02-28
2 pane support in Interests screen

jdkoren:
Integrates `ListDetailPaneScaffold` to show the topics list and the selected topic on the Interests screen when wide enough.


feature/topic/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicScreen.kt
```diff
@@ -280,13 +286,18 @@ private fun TopicToolbar(
...
             .padding(bottom = 32.dp),
     ) {
-        IconButton(onClick = { onBackClick() }) {
-            Icon(
-                imageVector = NiaIcons.ArrowBack,
-                contentDescription = stringResource(
-                    id = com.google.samples.apps.nowinandroid.core.ui.R.string.core_ui_back,
-                ),
-            )
+        if (showBackButton) {
```
alexvanyo, line 289: Nit: Could be fun to animate the visibility here.

feature/interests/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/interests/InterestsItem.kt
```diff
@@ -83,7 +84,11 @@ fun InterestsItem(
             )
         },
         colors = ListItemDefaults.colors(
-            containerColor = Color.Transparent,
+            containerColor = if (isSelected) {
```
alexvanyo, line 87: Instead of just changing the container color, we should also update the semantics to reflect being selected for accessibility.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
```diff
@@ -0,0 +1,140 @@
...
+        nestedNavController.navigateToTopic(topicId) {
+            popUpTo(DETAIL_PANE_NAVHOST_ROUTE)
+        }
+        listDetailNavigator.navigateTo(ListDetailPaneScaffoldRole.Detail)
+    }
+
+    ListDetailPaneScaffold(
+        scaffoldState = listDetailNavigator.scaffoldState,
+        listPane = {
+            InterestsRoute(
```
alexvanyo, line 98: Can we wrap both the `listPane` content and the `detailPane` content in `AnimatedPane(Modifier)`?

That should give us animations for swapping between list and detail.

jdkoren: We cannot. Or rather, we can only use it for the list pane, which doesn't improve things much.

The problem is that `AnimatedPane` skips its content when the pane is hidden, so if you have a device which shows only one pane, the call to `navigate` on the nested `NavController` happens before the `NavHost` has been composed, causing a crash: [animated_pane_crash.txt](https://github.com/android/nowinandroid/files/14450438/animated_pane_crash.txt)

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
```diff
@@ -0,0 +1,140 @@
...
+    LaunchedEffect(Unit) {
+        if (selectedTopicId != null) {
+            // Initial topic ID was provided when navigating to Interests, so show its details.
+            onTopicClickShowDetailPane(selectedTopicId)
+        }
+    }
+}
+
+@OptIn(ExperimentalMaterial3AdaptiveApi::class)
+private fun <T> ThreePaneScaffoldNavigator<T>.isListPaneHidden(): Boolean {
```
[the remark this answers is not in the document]
jdkoren: Done

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
alexvanyo: Tracing this issue internally - I vote we merge this now to iterate upon, and update to new versions of the library.

alexvanyo (approved):

feature/topic/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicScreen.kt
jdkoren: We can explore that. My initial impression is that it wouldn't be particularly noticeable though, because the animation would only be relevant if there's a config change that changes how many panes are visible.

PR #1256, opened 2024-03-06
Fix typo in TopicViewModel.kt

lnbs97:
Fixes typo in TopicScreen, TopicViewModel and TopicViewModelTest.

Wrong: newUiState
Correct: newsUiState


dturner (approved):
Good spot. Thank you!

PR #1257, opened 2024-03-07
Update module dependency graphs and add script to create them

dturner:
**What I have done and why**
- Added new module dependency graphs for each module (displayed in README.md for each module) so it's easy to visualise module dependencies
- Created a script to regenerate module dependency graphs so it's easier to keep them up to date

To run the script, execute the following command in the project's root folder:

`./generateModuleGraphs.sh --exclude-module :benchmarks --exclude-module :lint --exclude-module :ui-test-hilt-manifest`

The `--exclude-module` parameter is used to exclude modules which don't appear in the root dependency graph (orphaned modules, only used by external programs, like CI or Android Studio)

Related issues/PRs:
https://github.com/android/nowinandroid/pull/1088
https://github.com/android/nowinandroid/issues/707

## Future work
- Decide whether this should be integrated into CI, similar to screenshot tests and badging 

Nice to haves:
- Add URLs to the images either to navigate between module graphs. Requires adding URLs for each node in the graph, which would be better done by [modifying the plugin here](https://github.com/jraska/modules-graph-assert/blob/master/plugin/src/main/kotlin/com/jraska/module/graph/GraphvizWriter.kt), rather than hacking around with the .gv file it creates.  
- Default size of SVGs seems quite small. Tried using `dot -s` for scaling but didn't work.


generateModuleGraphs.sh
```diff
@@ -0,0 +1,92 @@
...
+    if [[ ! " ${excluded_modules[@]} " =~ " ${module_path} " ]]; then
+        # Derive the filename from the module path
+        file_name="dep_graph${module_path//:/_}" # Replace colons with underscores
+        file_name="${file_name//-/_}" # Replace dashes with underscores
+
+        check_and_create_readme "$module_path" "$file_name"
+
+        # Generate the .gv file in a temporary location
+        # </dev/null is used to stop ./gradlew from consuming input which prematurely ends the while loop
+        ./gradlew generateModulesGraphvizText -Pmodules.graph.output.gv="/tmp/${file_name}.gv" -Pmodules.graph.of.module="${module_path}" </dev/null
```
jdkoren, line 85: nit: split into multiple lines

jdkoren (approved):
LGTM overall

generateModuleGraphs.sh
```diff
@@ -0,0 +1,92 @@
...
+# Get the module paths
+module_paths=$(./gradlew -q printModulePaths --no-configuration-cache)
+
+# Function to check and create a README.md for modules which don't have one.
+check_and_create_readme() {
+    local module_path="$1"
+    local file_name="$2"
+
+    local readme_path="${module_path:1}" # Remove leading colon
+    readme_path=$(echo "$readme_path" | sed 's/:/\//g') # Replace colons with slashes using sed
```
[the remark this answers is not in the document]
dturner: Good suggestion. Amended.

core/common/README.md
```diff
@@ -1,3 +1,3 @@
 # :core:common module
 
-![Dependency graph](../../docs/images/graphs/dep_graph_core_common.png)
+![Dependency graph](../../docs/images/graphs/dep_graph_core_common.svg)
```
JoseAlcerreca, line 3: Does it make sense to show these graphs where there's only one module? There's no context so opening this page just shows an ellipse with the name of the module inside.

generateModuleGraphs.sh
```diff
@@ -0,0 +1,94 @@
...
+check_and_create_readme() {
+    local module_path="$1"
+    local file_name="$2"
+
+    local readme_path="${module_path:1}" # Remove leading colon
+    readme_path=${readme_path//:/\/} # Replace colons with slashes using sed
+    readme_path="${readme_path}/README.md" #Append the filename
+
+    # Check if README.md exists and create it if not
+    if [[ ! -f "$readme_path" ]]; then
```
JoseAlcerreca, line 63: Remove this check. If the README template changes they wouldn't be replaced and it's a minor optimization.

generateModuleGraphs.sh
```diff
@@ -0,0 +1,94 @@
...
+        echo "Creating README.md for ${module_path}"
+        # Calculate the correct relative path to the image
+        local relative_image_path="../../docs/images/graphs/${file_name}.svg"
+        echo "# ${module_path} module" > "$readme_path"
+        echo "" >> "$readme_path"
+        echo "![Dependency graph](${relative_image_path})" >> "$readme_path"
```
JoseAlcerreca, line 69: add `echo "## Dependency graph"` 
to add context to the image.

generateModuleGraphs.sh
dturner: That's exactly the issue. We might add extra info to a README and shouldn't overwrite that info when regenerating the module graphs. This is just to generate a new README when a new module is added.

core/common/README.md
dturner: I think it does still make sense to show modules with no dependencies, as it's explicitly showing that.

PR #1259, opened 2024-03-08
Fix search flickering

mlykotom:
**What I have done and why**
The search screen was flickering whenever user typed something into the search box.
This was caused by the `.asResult()` mapping, which maps to `Loading` whenever the query was started with a different search query.

https://github.com/android/nowinandroid/assets/3973717/6d8ca9bf-abd3-4f90-8079-028ba254ee80

https://github.com/android/nowinandroid/assets/3973717/dd1c2307-6fc0-4c6f-a04e-35412ea628d7

**Do tests pass?**


feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchViewModel.kt
```diff
@@ -61,18 +61,15 @@ class SearchViewModel @Inject constructor(
                             flowOf(SearchResultUiState.EmptyQuery)
                         } else {
                             getSearchContentsUseCase(query)
-                                .asResult()
```
jdkoren, line 64: Is it safe to remove this? (Don't we lose error handling, for instance?)

mlykotom: The error handling is done the same way as in the asResult() function via the catch operator.

So basically the functionality is the same except of the onStart { Loading}

jdkoren (approved):

PR #1282, opened 2024-03-13
Update usage of material3 adaptive to be edge-to-edge

alexvanyo:
Before:

https://github.com/android/nowinandroid/assets/4217560/592f8fe6-5dc5-4bf1-add4-0b9571860f02

After:

https://github.com/android/nowinandroid/assets/4217560/2223217a-47e9-45ec-a6fa-10f28d58fd2d


app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -164,6 +164,7 @@ fun NiaApp(appState: NiaAppState) {
                     Column(Modifier.fillMaxSize()) {
                         // Show the top app bar on top level destinations.
                         val destination = appState.currentTopLevelDestination
+                        val shouldShowTopAppBar = destination != null
```
yongsuk44, line 167: How about managing this code like `shouldShowBottomBar` in the `NiaAppState` class

alexvanyo: I thought about that, but it makes the logic a little bit tricky here since we would still need to check that `destination` is not null below.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
```diff
@@ -132,3 +153,59 @@ private fun <T> ThreePaneScaffoldNavigator<T>.isListPaneVisible(): Boolean =
...
+    if (windowAdaptiveInfo.windowPosture.isTabletop) {
+        maxVerticalPartitions = 2
+        horizontalSpacerSize = 24.dp
+    } else {
+        maxVerticalPartitions = 1
+        horizontalSpacerSize = 0.dp
+    }
+
+    return PaneScaffoldDirective(
+        PaddingValues(0.dp),
```
jdkoren, line 194: I think this was removed in alpha10, should we bump the version?

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
```diff
@@ -132,3 +153,59 @@ private fun <T> ThreePaneScaffoldNavigator<T>.isListPaneVisible(): Boolean =
 @OptIn(ExperimentalMaterial3AdaptiveApi::class)
 private fun <T> ThreePaneScaffoldNavigator<T>.isDetailPaneVisible(): Boolean =
     scaffoldValue[ListDetailPaneScaffoldRole.Detail] == PaneAdaptedValue.Expanded
+
+/**
+ * This is a direct clone of [calculateStandardPaneScaffoldDirective] with the only change of
```
jdkoren, line 158: It's unfortunate that there isn't a simpler way to configure a scaffold directive.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
alexvanyo: Updated, with the latest changes this also simplified the implementation greatly.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
alexvanyo: This is no longer needed with the update to alpha10

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -162,6 +162,7 @@ fun NiaApp(appState: NiaAppState) {
...
                         if (destination != null) {
```
Jaehwa-Noh, line 166:
```suggestion
                        if (shouldShowTopAppBar) {
```
This is same logic from ```shouldShowTopAppBar```

Sgzeus23: Null=$Sgzeus23

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
Sgzeus23: Its Null=$sgsike2914 my apologies its early

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
alexvanyo: This is the same logic, but using `shouldShowTopAppBar` directly results in not being able to smart-cast `destination` to be non-nullable below.

Jaehwa-Noh: Good, I missed the point of ```smart cast```.

jdkoren (approved):

PR #1289, opened 2024-03-15
Uses jacoco coverage reporting in AGP

JoseAlcerreca:


JoseAlcerreca:
```
% ./gradlew :app:tasks --all | grep -i CoverageReport                                                                                                                                                                                                                                                                                                                                                                           ✭
createDemoDebugCoverageReport - Creates instrumentation test coverage reports for the demoDebug variant.
createProdDebugCoverageReport - Creates instrumentation test coverage reports for the prodDebug variant.
createDemoDebugAndroidTestCoverageReport - Creates JaCoCo test coverage report from data gathered on the device.
createDemoDebugUnitTestCoverageReport - Generates a Jacoco code coverage report from unit tests.
createDemoReleaseUnitTestCoverageReport - Generates a Jacoco code coverage report from unit tests.
createManagedDeviceDemoDebugAndroidTestCoverageReport - Creates JaCoCo test coverage report from data gathered on the Gradle managed device.
createManagedDeviceProdDebugAndroidTestCoverageReport - Creates JaCoCo test coverage report from data gathered on the Gradle managed device.
createProdDebugAndroidTestCoverageReport - Creates JaCoCo test coverage report from data gathered on the device.
createProdDebugUnitTestCoverageReport - Generates a Jacoco code coverage report from unit tests.
createProdReleaseUnitTestCoverageReport - Generates a Jacoco code coverage report from unit tests.
```

JoseAlcerreca:
However, running `core:database:createProdDebugUnitTestCoverageReport` fails because it has no jvm tests:


* What went wrong:
A problem was found with the configuration of task ':core:database:createProdDebugUnitTestCoverageReport' (type 'JacocoReportTask').
  - Type 'com.android.build.gradle.internal.coverage.JacocoReportTask' property 'jacocoUnitTestCoverageFile' specifies file '/usr/local/google/home/jalc/src/nowinandroid/core/database/build/outputs/unit_test_code_coverage/prodDebugUnitTest/testProdDebugUnitTest.exec' which doesn't exist.


Filed https://issuetracker.google.com/329869184

We need to find a way to conditionally enable the flag (or run the test). I'll continue next week.

build-logic/convention/src/main/kotlin/AndroidApplicationJacocoConventionPlugin.kt
```diff
@@ -14,22 +14,21 @@
...
     override fun apply(target: Project) {
         with(target) {
-            with(pluginManager) {
-                apply("org.gradle.jacoco")
-                apply("com.android.application")
+            val androidExtension = extensions.getByType<BaseAppModuleExtension>()
+
+            androidExtension.buildTypes.forEach { buildType ->
+                buildType.enableAndroidTestCoverage = true
+                buildType.enableUnitTestCoverage = true
```
SimonMarquis, line 29: Let's use `configureEach` instead to be applied on all `buildTypes`. `forEach` will only use the current buildTypes, not the ones being created "after".

```suggestion
            androidExtension.buildTypes.configureEach {
                enableAndroidTestCoverage = true
                enableUnitTestCoverage = true
```

build-logic/convention/src/main/kotlin/AndroidLibraryJacocoConventionPlugin.kt
```diff
@@ -14,22 +14,26 @@
...
         with(target) {
-            with(pluginManager) {
-                apply("org.gradle.jacoco")
-                apply("com.android.library")
+            val androidExtension: LibraryExtension = extensions.getByType<LibraryExtension>()
+
+            androidExtension.buildTypes.forEach { buildType ->
+                // The jacoco plugin is applied automatically when any of these are set
+                buildType.enableAndroidTestCoverage = true
+                buildType.enableUnitTestCoverage = true
```
SimonMarquis, line 31:
```suggestion
            androidExtension.buildTypes.configureEach {
                // The jacoco plugin is applied automatically when any of these are set
                enableAndroidTestCoverage = true
                enableUnitTestCoverage = true
```

PR #1291, opened 2024-03-16
Rename Fake to Demo

Jaehwa-Noh:
**What I have done and why**
Rename Fake to Demo.
it is make analysis hard, the problem is naming.
Actually it is not fake because it is for demo to build on not product code.
The only ```NiaNetworkDataSource``` reference is ```FakeNiaNetworkDataSource``` in non-production build.
So, I think it is better to rename Fake to Demo.


dturner:
You make an excellent point. I agree, "Demo" is more accurate than "Fake". Thanks.

dturner (approved):

PR #1293, opened 2024-03-17
Fix `printModulePaths` executing during configuration phase

SimonMarquis:


SimonMarquis:
Also, to make it configuration-cache compatible, and avoid the need to specify `--no-configuration-cache` to avoid build errors, we could use this instead:

```kotlin
allprojects {
    if (subprojects.size != 0) return@allprojects
    val projectPath = path
    tasks.register("printModulePaths") {
        doLast {
            println(projectPath)
        }
    }
}
```

WDYT?

dturner:
I've asked someone from the gradle team to take a look at this as I don't have the gradle experience to say whether this is a good practice or not.

timothyfroehlich:
Random suggestion (I saw your review request): Check out the [new Gradle lifecycle callbacks](https://docs.gradle.org/release-nightly/release-notes.html#build-authoring-improvements) in Gradle 8.8 (which is still in development but had an RC a few weeks ago). It looks like it'll add an even cleaner way of doing what you need!

dturner:
Thanks for attempting to fix this @SimonMarquis, after discussion with the gradle team it seemed better to remove the gradle task altogether and keep everything in the bash script .

PR #1303, opened 2024-03-18
Sets up jacoco coverage with a combined report and a GHA

JoseAlcerreca:


.github/workflows/Build.yaml
```diff
@@ -193,9 +192,45 @@ jobs:
           heap-size: 600M
           script: ./gradlew connectedDemoDebugAndroidTest --daemon
 
+      - name: Run local tests (including Roborazzi) for the combined coverage report (only API 26)
+        if: matrix.api-level == 26
+        run: ./gradlew testDemoDebugUnitTest -Proborazzi.test.verify=true # Add Prod if we ever add JVM tests for prod
```
alexvanyo, line 197: Is `-Proborazzi.test.verify=true` still needed with the property in `gradle.properties`?

JoseAlcerreca: No, and I want to remove one of them. Thoughts? 

I decided to add it to `gradle.properties` because without them I noticed that the screenshot tests are still run on Robolectric, but no screenshots are taken or verified, so it's a waste of time. We might as well verify,

alexvanyo: I think `gradle.properties` is better, to always be verifying the screenshot tests, it feels weird to run a test that doesn't actually check anything. With that in `gradle.properties` we can also remove the explicit `verifyRoborazziDemoDebug` that we do above with the `TODO` and just use `testDemoDebugUnitTest`?

JoseAlcerreca: I don't think we can, because steps `checkfork_screenshots` and `screenshotsrecord` depend on whether `screenshotsverify` passes. Also, that step has `continue-on-error: true` so a bad unit test would be ignored.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt
```diff
@@ -16,14 +16,23 @@
...
+import com.android.build.api.artifact.ScopedArtifact
 import com.android.build.api.variant.AndroidComponentsExtension
+import com.android.build.api.variant.ScopedArtifacts
+import com.android.build.gradle.internal.coverage.JacocoReportTask
 import org.gradle.api.Project
+import org.gradle.api.file.Directory
+import org.gradle.api.file.RegularFile
+import org.gradle.api.provider.ListProperty
 import org.gradle.api.tasks.testing.Test
+import org.gradle.internal.impldep.bsh.commands.dir
```
alexvanyo, line 28: Is this import intended? Importing `org.gradle.internal` seems a bit suspicious

.github/workflows/Build.yaml
```diff
@@ -193,9 +192,41 @@ jobs:
...
+      - name: Display local test coverage (only API 30)
+        if: matrix.api-level == 30
+        id: jacoco
+        uses: madrapps/jacoco-report@v1.6.1
+        with:
+          title: Combined test coverage report
+          min-coverage-overall: 40
+          min-coverage-changed-files: 60
+          paths: |
+            ${{ github.workspace }}/**/build/reports/jacoco/**/*Report.xml
```
alexvanyo, line 221: Interesting, so this does the combining at the GHA level, not at the Gradle level? In other words, there isn't a local way to generate a combined coverage report from all modules.

build-logic/convention/src/main/kotlin/AndroidLibraryJacocoConventionPlugin.kt
```diff
@@ -23,13 +25,16 @@ import org.gradle.kotlin.dsl.getByType
...
     override fun apply(target: Project) {
         with(target) {
-            with(pluginManager) {
-                apply("org.gradle.jacoco")
-                apply("com.android.library")
+            pluginManager.apply("jacoco")
+            val androidExtension = extensions.getByType<LibraryExtension>()
+
+            androidExtension.buildTypes.configureEach {
+                // The jacoco plugin is applied automatically when any of these are set
```
alexvanyo, line 32: Is this true? I see that `pluginManager` is applying `jacoco` above directly still.

alexvanyo (commented):
Looks good overall, just a couple minor comments.

.github/workflows/Build.yaml
JoseAlcerreca: There are plugins and snippets that get you [there](https://github.com/SurpSG/jacoco-gradle-plugin-merge-coverage/blob/master/build.gradle#L28C1-L41C2), but a combined (jvm+device), combined (across modules) report is out of the scope of this PR 😁 I'll file a FR.

build-logic/convention/src/main/kotlin/AndroidLibraryJacocoConventionPlugin.kt
JoseAlcerreca: It's complicated: you don't need to apply it if you only use the flags, but we're configuring `JacocoPluginExtension` in Jacoco.kt and that needs the plugin to be applied manually. I'll remove the comment, it's more confusing that helpful.

SimonMarquis:
Hi @JoseAlcerreca & @alexvanyo, it would be great to officially document was is expected in terms of code coverage for this project.

I have strong feelings about code coverage. In my personal experience, this metrics is generally a bad proxy to measure the "quality" of a codebase. People tend to add "stupid" tests simply to make the number not go down in a PR. This is bad for so many resons (and now these tests will even be written by AIs 🙈). It also creates even more legacy code that becomes harder to refactor. And I've seen many teams fall into this false sense of security trap.

Although, as an additionaly metric, it can still be useful. For example, to detect spots in the codebase that are not tested.

PS: It seems like Android Studio no longer show the `Run with coverage` option as described in https://developer.android.com/studio/test/test-in-android-studio#view_test_coverage

lihenggui:
> Hi @JoseAlcerreca & @alexvanyo, it would be great to officially document was is expected in terms of code coverage for this project.
> 
> I have strong feelings about code coverage. In my personal experience, this metrics is generally a bad proxy to measure the "quality" of a codebase. People tend to add "stupid" tests simply to make the number not go down in a PR. This is bad for so many resons (and now these tests will even be written by AIs 🙈). It also creates even more legacy code that becomes harder to refactor. And I've seen many teams fall into this false sense of security trap.
> 
> Although, as an additionaly metric, it can still be useful. For example, to detect spots in the codebase that are not tested.
> 
> PS: It seems like Android Studio no longer show the `Run with coverage` option as described in https://developer.android.com/studio/test/test-in-android-studio#view_test_coverage

I have the same concern that people may introduce unnecessary tests to inflate the numbers and make the code harder to refactor, as I have observed in other projects. Finally, it will make everyone conservative. 
Another question arises: apart from test coverage, what other metrics can enhance project quality and motivate team members to improve their code?

JoseAlcerreca:
@lihenggui and @SimonMarquis we're working on a "NiA Testing Story" which has this section about coverage:

## Code coverage
We use code coverage figures to detect gaps in testing, but a minimum coverage is not enforced, as it depends a lot on the type of feature. 

Our CI system shows a coverage report for information only. The author and reviewers must decide if the changed files are conveniently tested.

PR #1318, opened 2024-03-22
Bump spotless(6.23.3 to 6.25.0) and ktlint(1.0.1 to 1.2.1)

Jaehwa-Noh:
**What I have done and why**
Bump spotless(6.23.3 to 6.25.0) and ktlint(1.0.1 to 1.2.1) up

### DIFF
--- a/gradle/init.gradle.kts
+++ b/gradle/init.gradle.kts
@@ -14,10 +14,10 @@
  *   limitations under the License.
  */
 
-val ktlintVersion = "1.0.1"
+val ktlintVersion = "1.2.1"
 
 initscript {
-    val spotlessVersion = "6.23.3"
+    val spotlessVersion = "6.25.0"
 
     repositories {
         mavenCentral()


SimonMarquis:
Dependabot only checks for specific locations, and this is not using the standard Gradle version catalog.

Jaehwa-Noh:
I got it. Thank you!

gradle/init.gradle.kts
```diff
@@ -14,10 +14,10 @@
  *   limitations under the License.
  */
 
-val ktlintVersion = "1.0.1"
+val ktlintVersion = "1.2.1"
```
blue928sky, line 17: Looks like a new version has been released.
https://github.com/pinterest/ktlint/releases/tag/1.4.0

Vladimir-cod: Интеграл дифирентен

dturner, line 17:
```suggestion
val ktlintVersion = "1.4.0"
```

dturner (approved):

dturner:
The updated version of ktlint changes the formatting rules which causes the build to fail. 

@Jaehwa-Noh Would you be able to take a look at how to disable the new rules so that our existing codebase isn't affected? This post might be useful: https://github.com/android/compose-samples/pull/1443#issuecomment-2284606621

Jaehwa-Noh:
@dturner Okay, I'll get it as soon as possible.
I need a spare time.

Jaehwa-Noh:
@dturner 
I'd removed new features all which is added in version ktlint 1.3.0(https://github.com/pinterest/ktlint/releases/tag/1.3.0), but still need to change some code styles.
It was as though some inner code's behaviors had changed.
So that, I've disabled `Function signature` too, https://pinterest.github.io/ktlint/latest/rules/standard/#function-signature
Would you see the `Function signature` rule and select which option one? `enable`, `disable`, `default`, `multiline`, `always`.

keyboardsurfer (approved):

PR #1319, opened 2024-03-23
Improve the Performance of Gradle Builds

Jaehwa-Noh:
**What I have done and why**
Optimize the compiler.
Run the compiler as a separate process.
[documentation](https://docs.gradle.org/current/userguide/performance.html#optimize_the_compiler)

Total Build Time improved 2%

• Before
before

• After
after

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,3 +59,12 @@ tasks.register("printModulePaths") {
         }
     }
 }
+
+/**
+ * Optimize the compiler
+ * Run the compiler as a separate process
+ * [documentation](https://docs.gradle.org/current/userguide/performance.html#optimize_the_compiler)
+ */
+tasks.withType<JavaCompile>().configureEach {
+    options.isFork = true
+}


SimonMarquis:
Hi @Jaehwa-Noh, thanks for this.
I personally feel like a 2% improvement in build time (only for the `:app:assembleDemoDebug` task) for the scale of this project is kind of insignificant, especially when almost everything is now cacheable.

Is it actually reproducible? I mean, if you run it 10 time on CI, will it lead to the same numbers?
You can use https://github.com/gradle/gradle-profiler to help you achieve this on a larger number of builds and output graphs.
Could you also show the overall memory usage, because forking processes is not free :)

Jaehwa-Noh:
Hello, @SimonMarquis 

Thank you for giving me new insight.

I would try what you suggest.
But before I do, I tell you my thought.
As you also known about the caching, my work is only run on 'doesn't have a cache', because when gradle have cache data, gradle use that cache, not rebuild.
In a nutshell, It would have work when rebuild without cache.

In [Google I/O 2017](https://youtu.be/7ll-rkLCtyk?feature=shared&t=636) they shows 2 seconds improved build time in charts, I don't think 2% is ignorable time.

Jaehwa-Noh:
I tested it by 4 tests: fork_clean, fork_no_clean, no_fork_clean, no_fork_no_clean.
My test scenario.txt
```
fork_no_clean {
	tasks = [":app:assembledemoDebug"]
	jvm-args = ["-Xmx6g", "-XX:+HeapDumpOnOutOfMemoryError", "-Dfile.encoding=UTF-8", "-XX:+UseParallelGC", "-XX:MaxMetaspaceSize=1g"]
	gradle-args = ["--parallel", "--configuration-cache", "--configuration-cache-problems=warn", "--offline"]
}

fork_clean {
	tasks = [":app:assembledemoDebug"]
	jvm-args = ["-Xmx6g", "-XX:+HeapDumpOnOutOfMemoryError", "-Dfile.encoding=UTF-8", "-XX:+UseParallelGC", "-XX:MaxMetaspaceSize=1g"]
	cleanup-tasks = ["clean"]
	gradle-args = ["--parallel", "--configuration-cache", "--configuration-cache-problems=warn", "--offline"]
}
```

image

image
-

I check out Median value. The unexpected result came out.
I expect both will be improved but just on ```no clean``` did.
(no_fork -> fork)
Cleaned and build: 7,284.06 ms -> 7,310.87 ms (hinder 0.3%)
No clean and build: 988.89 ms -> 934.10 ms (improve 5.5%)

When I'd cleaned before built, build time hindered 0.3%, however I'd not cleaned and built that build time improved 5.4%.

PR #1320, opened 2024-03-23
Update Missing Test Cases

yongsuk44:
**What I have done and why**
- Add given for 'DynamicColorPreference' in `SettingsViewModel`
- Add successful case for deleting RecentSearchQuery
- Add successful case for `setNewsResourceViewed`

**Do tests pass?**

### DIFF
--- a/core/testing/src/main/kotlin/com/google/samples/apps/nowinandroid/core/testing/repository/TestUserDataRepository.kt
+++ b/core/testing/src/main/kotlin/com/google/samples/apps/nowinandroid/core/testing/repository/TestUserDataRepository.kt
@@ -76,16 +76,13 @@ class TestUserDataRepository : UserDataRepository {
 
     override suspend fun setNewsResourceViewed(newsResourceId: String, viewed: Boolean) {
         currentUserData.let { current ->
-            _userData.tryEmit(
-                current.copy(
-                    viewedNewsResources =
-                    if (viewed) {
-                        current.viewedNewsResources + newsResourceId
-                    } else {
-                        current.viewedNewsResources - newsResourceId
-                    },
-                ),
-            )
+            val viewedNews = if (viewed) {
+                current.viewedNewsResources + newsResourceId
+            } else {
+                current.viewedNewsResources - newsResourceId
+            }
+
+            _userData.tryEmit(current.copy(viewedNewsResources = viewedNews))
         }
     }
 
--- a/core/testing/src/main/kotlin/com/google/samples/apps/nowinandroid/core/testing/repository/TestUserDataRepository.kt
+++ b/core/testing/src/main/kotlin/com/google/samples/apps/nowinandroid/core/testing/repository/TestUserDataRepository.kt
@@ -24,6 +24,7 @@ import kotlinx.coroutines.channels.BufferOverflow.DROP_OLDEST
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.filterNotNull
+import org.jetbrains.annotations.TestOnly
 
 val emptyUserData = UserData(
     bookmarkedNewsResources = emptySet(),
@@ -112,9 +113,7 @@ class TestUserDataRepository : UserDataRepository {
         }
     }
 
-    /**
-     * A test-only API to allow setting of user data directly.
-     */
+    @TestOnly
     fun setUserData(userData: UserData) {
         _userData.tryEmit(userData)
     }
--- a/feature/foryou/src/test/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModelTest.kt
+++ b/feature/foryou/src/test/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModelTest.kt
@@ -524,6 +524,17 @@ class ForYouViewModelTest {
             actual = userDataRepository.userData.first().bookmarkedNewsResources,
         )
     }
+
+    @Test
+    fun whenSetNewsResourceViewedIsCalled_viewedStateIsUpdated() = runTest {
+        val newsResourceId = "123"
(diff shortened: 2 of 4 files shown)


core/testing/src/main/kotlin/com/google/samples/apps/nowinandroid/core/testing/repository/TestUserDataRepository.kt
```diff
@@ -112,9 +110,7 @@ class TestUserDataRepository : UserDataRepository {
         }
     }
 
-    /**
-     * A test-only API to allow setting of user data directly.
-     */
+    @TestOnly
```
Jaehwa-Noh, line 113: Testing module consist of bundle of test source.
I'm not sure ```@TestOnly``` needs in here.
Just my opinion.

yongsuk44: @Jaehwa-Noh Thanks for the feedback 😀

One could think that providing duplicate information might be unnecessary.
The change to `@TestOnly` was made to simplify and explicitly state what had previously been indicated through comments.

hoc081098: `@VisibleForTesting` can be used

_Sent from my 2201117TG using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

yongsuk44: My understanding of the differences between each annotation is as follows:

- `@VisibleForTesting` is used to explicitly state that the visibility of an element has been increased for testing purposes. In other words, while a lower visibility (e.g., private) would suffice for the actual app operation, the visibility has been temporarily raised to allow test code to access that element.
- `@TestOnly` indicates that specific code should be used in the test environment only. Thus, by using this annotation, it is explicitly stated that the code should not be used in actual production code.

@hoc081098  Am I misunderstanding anything? 🤔

JoseAlcerreca: Can we refactor this so that the repository can be tested without using this API?

yongsuk44: @JoseAlcerreca Okay, I'll try 😀

yongsuk44:
@JoseAlcerreca RFR :)

yongsuk44:
Fixed Issue - #1337

keyboardsurfer:
@yongsuk44 please take a look at the conflicts and resolve them so we can proceed integrating this PR.

core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/CompositeUserNewsResourceRepositoryTest.kt
```diff
@@ -42,70 +40,52 @@ class CompositeUserNewsResourceRepositoryTest {
...
         // Obtain a stream of user news resources for the given topic id.
-        val userNewsResources =
-            userNewsResourceRepository.observeAll(
-                NewsResourceQuery(
-                    filterTopicIds = setOf(
-                        sampleTopic1.id,
-                    ),
-                ),
-            )
+        val userNewsResources = userNewsResourceRepository.observeAllForFollowedTopics()
```
dturner, line 60: Changing the test in this way tests a different method: observeAllForFollowedTopics, which is tested in an identical way in the test below. 

This test should be left as-is. It tests the scenario where a user is not following any of those topics but is still viewing them. For example, when viewing the TopicDetail screen for a topic that they're not following.

core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/CompositeUserNewsResourceRepositoryTest.kt
```diff
@@ -119,16 +99,14 @@ class CompositeUserNewsResourceRepositoryTest {
...
         // Construct the test user data with bookmarks and followed topics.
-        val userData = emptyUserData.copy(
-            bookmarkedNewsResources = setOf(sampleNewsResources[0].id, sampleNewsResources[2].id),
-            followedTopics = setOf(sampleTopic1.id),
-        )
-
-        userDataRepository.setUserData(userData)
+        userDataRepository.setNewsResourceBookmarked(sampleNewsResources[0].id, true)
+        userDataRepository.setNewsResourceBookmarked(sampleNewsResources[2].id, true)
+        userDataRepository.setFollowedTopicIds(setOf(sampleTopic1.id))
```
dturner, line 104: Please use named arguments e.g. `bookmarked=true`

dturner (changes requested):
Thanks for doing this, and apologies for the delay in reviewing. Please see comments.

.run/Generate Demo Baseline Profile.run.xml
dturner, line 1: This deletion doesn't look relevant to this PR so should be removed from the PR.

.run/spotlessApply.run.xml
dturner, line 1: This deletion doesn't look relevant to this PR so should be removed from the PR.

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
```diff
@@ -25,6 +26,7 @@ import androidx.navigation.navArgument
 import androidx.navigation.navDeepLink
 import com.google.samples.apps.nowinandroid.feature.foryou.ForYouRoute
 
+@VisibleForTesting
```
dturner, line 29: Why is this required?

Issue #1295, opened 2024-03-18
[Bug]: Can't see a snackbar in medium and expanded screen


Jaehwa-Noh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

In airplane mode, and medium and expanded size screen, I cannot see the no connected snackbar.
It hide in Bottom Navigation Bar.

![Cannot see a snackbar](https://github.com/android/nowinandroid/assets/48680511/964e3be2-5688-455a-9be7-f2d4a6fa6b96)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1343, opened 2024-03-26
Adds screenshot tests for snackbar

JoseAlcerreca:


Jaehwa-Noh:
I'm afraid the screenshot tests can't cover the real snackbar screen. Because there's no system UI so that can't check this issue #1295.

JoseAlcerreca:
Yes, good reminder but that's a bug on the UI. When it's fixed, the author should add screenshot tests for it.

alexvanyo (approved):

PR #1356, opened 2024-04-02
Optimize recomposition in NavHost using remember

### LINKED ISSUE #1353: [FR]: Minimize Unnecessary Recompositions

### Describe the problem

In current implementation of NiaApp, we've observed an issue where unrelated state changes, such as modifications to showSettingsDialog, trigger unnecessary recompositions across the entire app, including within NiaNavHost and its child screens

Upon investigation, we identified that lambda expressions, such as onTopicClick, are being recreated with each recomposition cycle. Compose treats these newly instantiated lambda expressions as different from their predecessors, thus triggering recompositions even when there are no pertinent state changes.

image

this behavior will persist until the Jetpack Compose Compiler is updated to version 1.7, which is expected to introduce an enhanced 'Strong skipping' mode for more efficient recomposition skipping. In the meantime, developers are advised to manage lambda instances manually using remember to mitigate unnecessary recompositions.

This is particularly evident in screens like ForYouScreen, where the onTopicClick lambda is passed down and utilized. Despite showSettingsDialog having no direct relation to these screens, they undergo recomposition, leading to performance inefficiencies.

### Describe the solution

I propose to mitigate this issue by employing the `remember` construct to cache navigation-related lambda expressions within `NiaNavHost`. This approach will ensure that lambdas are only recreated when genuinely necessary, such as when the `NavController` instance changes

### change
image

### before
image

### after
image

### Additional context

_No response_

### Code of Conduct

KDW03:
_Thanks for submitting a pull request. Please include the following information._

**What I have done and why**
Implemented remember for lambda expressions in NiaNavHost to reduce unnecessary recompositions, addressing issue #1353. This optimization minimizes performance overhead caused by state changes unrelated to navigation.

Fixes #1353 

**How I'm testing it**

_Choose at least one:_
- Unit tests
- UI tests

**Do tests pass?**

- [x] Run `./tools/setup.sh`
- [x] Import the code formatting style as explained in [the setup script](/tools/setup.sh#L40).

### DIFF
--- a/feature/interests/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/interests/navigation/InterestsNavigation.kt
+++ b/feature/interests/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/interests/navigation/InterestsNavigation.kt
@@ -37,6 +37,15 @@ fun NavController.navigateToInterests(topicId: String? = null, navOptions: NavOp
     navigate(route, navOptions)
 }
 
+fun NavController.navigateToInterests(topicId: String? = null) {
+    val route = if (topicId != null) {
+        "${INTERESTS_ROUTE_BASE}?${TOPIC_ID_ARG}=$topicId"
+    } else {
+        INTERESTS_ROUTE_BASE
+    }
+    navigate(route, null)
+}
+
 fun NavGraphBuilder.interestsScreen(
     onTopicClick: (String) -> Unit,
 ) {
--- a/app/src/main/kotlin/com/google/samples/apps/nowinandroid/navigation/NiaNavHost.kt
+++ b/app/src/main/kotlin/com/google/samples/apps/nowinandroid/navigation/NiaNavHost.kt
@@ -44,7 +44,7 @@ fun NiaNavHost(
     startDestination: String = FOR_YOU_ROUTE,
 ) {
     val navController = appState.navController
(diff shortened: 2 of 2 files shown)


feature/interests/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/interests/navigation/InterestsNavigation.kt
```diff
@@ -37,6 +37,15 @@ fun NavController.navigateToInterests(topicId: String? = null, navOptions: NavOp
     navigate(route, navOptions)
 }
 
+fun NavController.navigateToInterests(topicId: String? = null) {
```
Jaehwa-Noh, line 40: Above line 31 ```navigateToInterests``` function already get ```navOptions``` as null, why do you create this function again?

KDW03: Originally, onTopicClick was used in the form (String) -> Unit, and there was a desire to maintain this structure as much as possible. 

In this situation, minimizing changes to the existing code led to the decision to maintain two versions of the navigateToInterests function while keeping onTopicClick in its original form.

This approach allowed for the integration of new feature requirements without significant changes to the existing code structure

PR #1368, opened 2024-04-10
Offload connectivity monitor to a background thread

mlykotom:
**What I have done and why**

Moved the `ConnectivityManagerNetworkMonitor` to a background thread to prevent blocking main thread by ~5ms during startup.

image

**How I'm testing it**

**Do tests pass?**

### DIFF
--- a/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
+++ b/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
@@ -26,7 +26,6 @@ import android.net.NetworkRequest.Builder
 import android.os.Build.VERSION
 import android.os.Build.VERSION_CODES
 import androidx.core.content.getSystemService
-import androidx.tracing.Trace
 import androidx.tracing.trace
 import com.google.samples.apps.nowinandroid.core.network.Dispatcher
 import com.google.samples.apps.nowinandroid.core.network.NiaDispatchers.IO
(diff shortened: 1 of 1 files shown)


core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -26,57 +26,71 @@ import android.net.NetworkRequest.Builder
...
-            private val networks = mutableSetOf<Network>()
+                override fun onAvailable(network: Network) {
+                    networks += network
+                    channel.trySend(true)
+                }
 
-            override fun onAvailable(network: Network) {
-                networks += network
-                channel.trySend(true)
+                override fun onLost(network: Network) {
```
yschimke, line 68: Does onLost get called when the network remains, but it loses the NET_CAPABILITY_INTERNET capability?

Or do you need to listen to onCapabilitiesChanged() also?

mlykotom: hm, I wasn't changing the implemntation here a tall, just wrapped it with the `trace` 😀

core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -26,57 +26,71 @@ import android.net.NetworkRequest.Builder
...
+            /**
+             * Sends the latest connectivity status to the underlying channel.
+             */
+            channel.trySend(connectivityManager.isCurrentlyConnected())
 
-        /**
-         * Sends the latest connectivity status to the underlying channel.
-         */
-        channel.trySend(connectivityManager.isCurrentlyConnected())
+            Trace.endSection()
```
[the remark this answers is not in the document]
mlykotom: Oh,good catch, I had this for dev, then I wrapped it with `trace`. 👍🏼

alexvanyo (approved):

PR #1369, opened 2024-04-10
Optimize startup by preventing whole screen recomposing twice

mlykotom:
**What I have done and why**

I reworked the way how we set up the MainActivity wrt to theming and user settings.

This includes:
- reacting to `uiMode` config change (before, the whole activity would be recreated when theme changed)
- encapsulate the state logic into its class
- prevent calling `enableEdgeToEdge` twice with different settings

This way I can get rid of one whole-screen recomposition, which is ~12ms on a Pixel 6.
image

**Do tests pass?**


app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -81,58 +74,64 @@ class MainActivity : ComponentActivity() {
...
     val viewModel: MainActivityViewModel by viewModels()
-
     override fun onCreate(savedInstanceState: Bundle?) {
         val splashScreen = installSplashScreen()
         super.onCreate(savedInstanceState)
 
-        var uiState: MainActivityUiState by mutableStateOf(Loading)
+        // We keep this as a mutable state, so that we can track changes inside the composition.
+        // This allows us to react to dark/light mode changes.
+        var themeSettings by mutableStateOf(
```
yschimke, line 83: In practice aren't a bunch of these `staticCompositionLocalOf` which would cause a full recomposition anyway?

https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#staticCompositionLocalOf(kotlin.Function0)

Would this be simpler if you setKeepOnScreenCondition and maybe a conditional to wait for the themeSetting before composing any content?

mlykotom: Sorry, I'm not sure I understand fully what you're asking 🤔 .
I think this is a bit different than `staticCompositionLocalOf`, because the nested composables can be skipped if params haven't changed. Though given we change the theme settings, maybe this would change the whole tree anyway (or most of it).

We're actually using the `setKeepOnScreenCondition` and I was trying to prevent composing anything if the state is not `Success`, but that didn't show better results.

This way we can improve the happy-path (default settings) to only compose once.

yschimke: Yeah, `staticCompositionLocalOf` is basically a global reset.  Your results and benchmarks will vary depending on whether the initial value

```
            ThemeSettings(
                darkTheme = resources.configuration.isSystemInDarkTheme,
                androidTheme = false,
                disableDynamicTheming = true,
            ),
```

is different in a way from the final

```
                    ThemeSettings(
                        darkTheme = uiState.shouldUseDarkTheme(systemDark),
                        androidTheme = uiState.shouldUseAndroidTheme,
                        disableDynamicTheming = uiState.shouldDisableDynamicTheming,
                    )
```

So it seems like the effect of the optimisation is dependent on these values.

So yeah, I was wondering if the correct solution is to wait until you have uiState before removing the splashscreen?

mlykotom: You're right that it depends on the `ThemeSettings` state.
The thing is, we actually wait and keep the splash screen on https://github.com/android/nowinandroid/pull/1369/files#diff-705d6b4cbfd6b5724912a84fedddac4c41309ddfb4c5f0950e6e7bf1d6d4371aR132

yschimke: Can you make the default value null then?

And just skip drawing if null, which it shouldn't be.

mlykotom: I tried preventing to compose anything if the `uiState != Success`, but then I realised there's a bunch of stuff that would be postponed and done the first time (like `rememberNiaAppState`, the composition locals and even stuff within the `NiaApp`).

If this recomposes second time, some composables can be skipped, so it should be shorter.
If we recompose everything the first time, we're just delaying it in time.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/MainActivity.kt
```diff
@@ -81,58 +77,64 @@ class MainActivity : ComponentActivity() {
...
+                ) { systemDark, uiState ->
+                    ThemeSettings(
+                        darkTheme = uiState.shouldUseDarkTheme(systemDark),
+                        androidTheme = uiState.shouldUseAndroidTheme,
+                        disableDynamicTheming = uiState.shouldDisableDynamicTheming,
+                    )
+                }
+                    .onEach { themeSettings = it }
+                    .map { it.darkTheme }
+                    .distinctUntilChanged()
```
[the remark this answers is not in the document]
mlykotom: My thinking was that I don't want the `enableEdgeToEdge` to be called when unrelated things change (`androidTheme` or `disableDynamicTheming`). Althought `isSystemInDarkTheme` and `uiState` deduplicate emissions, the `combine` of them don't. We don't care when setting the `themeSettings`, because `mutableStateOf` automatically uses `structuralEqualityPolicy`, but then the param doesn't. 

Maybe it's over-complicated though 😅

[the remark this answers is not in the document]
yongsuk44: My question comes from the fact that each `Flow` only emits the latest values after doing deduplication upstream. Considering this, is there still a point in doing deduplication downstream in the `combine` function? 
Since each Flow already ensures only the latest values are emitted, deduplicating again downstream seems redundant to me.

```kotlin
val ms = MutableStateFlow(1)

launch {
    delay(5000)
    ms.update { it + 1 }
}

val fs = flow {
    emit(true)
    delay(2000)
    emit(true)
    delay(2000)
    emit(false)
    delay(2000)
    emit(false)
    delay(2000)
    emit(false)
}
    .onStart { println("Start") }
    .onCompletion { println("End") }
    .distinctUntilChanged()

combine(fs, ms) { f, m -> "$f, $m" }
    .onEach { println(it) }
    .collect()

// Start
// true, 1
// false, 1
// false, 2
// End
```

Could I be misunderstanding something here? 🥲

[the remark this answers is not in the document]
mlykotom: Hm, I think your example is not complete.

Let's consider this:
```
combine(fs, ms) { f, m -> f to m }
    .onEach { println("set theme $it") }
    .onEach { println("enable edge to edge ${it.second}") }
    .collect()
```

This will produce
```
Start
set theme (true, 1)
enable edge to edge 1
set theme (false, 1)
enable edge to edge 1 <---- unnecessary call
set theme (false, 2)
enable edge to edge 2
End
```

While if you do the `map` + `distinct`

```
combine(fs, ms) { f, m -> f to m }
    .onEach { println("set theme $it") }
    .map { it.second }
    .distinctUntilChanged()
    .onEach { println("enable edge to edge $it") }
    .collect()
```

```
Start
set theme (true, 1)
enable edge to edge 1
set theme (false, 1)
set theme (false, 2)
enable edge to edge 2 <-- called only when "m" changed
End
```

So the reason is to skip calling `enableEdgeToEdge` if it didn't change.
The reason is that this call doesn't have internal check if it has the same parameters, but the resulting change is short, so maybe it's just overcomplicated.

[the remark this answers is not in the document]
yongsuk44: Ah.. I completely understand, thank you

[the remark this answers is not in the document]
mlykotom: Thanks for the question!

keyboardsurfer:

app/src/main/kotlin/com/google/samples/apps/nowinandroid/util/UiExtensions.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Convenience wrapper for dark mode checking
+ */
+val Configuration.isSystemInDarkTheme
+    get() = (uiMode and Configuration.UI_MODE_NIGHT_MASK) == Configuration.UI_MODE_NIGHT_YES
+
+/**
+ * Registers listener for configuration changes to retrieve whether system is in dark theme or not.
+ * Immediately upon subscribing, it sends the current value and then registers listener for changes.
+ */
+fun ComponentActivity.isSystemInDarkTheme() = callbackFlow {
```
Jaehwa-Noh, line 37: Does callbackFlow need in here?
Did Configuration change already have a way to catch it, didn't?
https://developer.android.com/guide/topics/resources/runtime-changes

mlykotom:
> Thanks for your PR. Currently there are conflicts that need to be resolved before we can move on to possibly accepting your contribution. Please resolve the conflicts.

Fixed. PTAL

keyboardsurfer (approved):

app/src/main/kotlin/com/google/samples/apps/nowinandroid/util/UiExtensions.kt
mlykotom: Configuration changes by default restart the activity, which is what we don't want to have, because Compose can gracefuly handle the theme change recomposing just what needs to be recomposed. 
This is why we listen to it here, so that we can react to it in the Compose code.

PR #1373, opened 2024-04-12
Add NavigationSuiteScaffold

alexvanyo:
Adds `NavigationSuiteScaffold` to replace the current manual logic to switch between a navigation rail and navigation bar.


app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/NavigationUiTest.kt
alexvanyo, line 1: Since we aren't doing this logic in the project directly, I figured we can probably remove this test? The screenshot tests should be verifying that we are indeed switching between the UI components.

app/src/testDemo/screenshots/compactWidth_expandedHeight_showsNavigationBar.png
alexvanyo, line 1: This is changing the tonal elevation we are using for the navigation bar, as the `NavigationSuiteScaffold` doesn't allow that much customization.

If we strongly feel like we want to keep the current look of the navigation bar (which is overriding the tonal elevation to `0dp`), then the intended usage would be to just set up the navigation UI manually, like we are doing right now, but then we wouldn't be using `NavigationSuiteScaffold`.

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/NavigationUiTest.kt
IanGClifton: That seems reasonable to me since this is basically duplicative of NavigationSuiteScaffoldTest.kt

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/Navigation.kt
```diff
@@ -165,6 +182,116 @@ fun NiaNavigationRail(
...
+        containerColor = Color.Transparent,
+        navigationSuiteColors = NavigationSuiteDefaults.colors(
+            navigationBarContentColor = NiaNavigationDefaults.navigationContentColor(),
+            navigationRailContainerColor = Color.Transparent,
+        ),
+        modifier = modifier,
+    ) {
+        Box(
+            // TODO: It'd be nice if navigation suite scaffold did this automatically:
+            //       https://issuetracker.google.com/334178393
```
IanGClifton, line 247: This is fixed in the next release, so maybe create an issue in the nowinandroid repo to update to that version and delete this custom code.

app/src/testDemo/screenshots/compactWidth_expandedHeight_showsNavigationBar.png
IanGClifton: Another option would be overriding the navigation bar colors specifically to what it would be with that elevation, but that seems a bit like fighting the APIs. I think the new colors look fine, but I don't know who should approve them.

IanGClifton (commented):
LGTM but I think someone else needs to approve for the color changes. It seems more inline with what M3 generally looks like, but I don't know if there was any history in the decision.

app/src/testDemo/screenshots/compactWidth_expandedHeight_showsNavigationBar.png
alexvanyo: That could work, but eww... especially since we have to handle dynamic colors and the Android colors.

@dturner do you think the visual update here is fine to merge with?

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/Navigation.kt
alexvanyo: Removed this workaround with the latest version of the library.

jdkoren (approved):

app/src/testDemo/screenshots/compactWidth_expandedHeight_showsNavigationBar.png
alexvanyo: Going to merge as-is, with the color change update.

PR #1375, opened 2024-04-13
Migrate legacy Gradle actions to the new official `gradle/actions`

SimonMarquis:
- Migrated legacy Gradle actions to the new official https://github.com/gradle/actions/ repostory.
  You can read more on that here: https://github.com/gradle/gradle-build-action/releases/tag/v3.3.0
- Configure `validate-wrappers: true` with `gradle/actions/setup-gradle` instead of using `gradle/actions/wrapper-validation` ([docs](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#gradle-wrapper-validation)).
- Use `gradle-home-cache-cleanup: true` parameter as it is not yet the default (but it will in the future).
- Add missing `gradle/actions/setup-gradle` in `Release.yml` (this workflow was never run, see https://github.com/android/nowinandroid/discussions/1374)


SimonMarquis:
@dturner This is ready to be reviewed/merged 😀

dturner (approved):

PR #1380, opened 2024-04-16
Recreate nested nav to work with AnimatedPane

alexvanyo:
Adds `AnimatedPane` to `ListDetailPaneScaffold` for interests by introducing a workaround for the nested navigation host setup.

Previously, adding `AnimatedPane` would crash when just the list was showing, since we'd try to call `navigate` on the nested nav controller before the graph for the nested nav controller was setup.

The workaround in this PR is to only call `navigate` on the nested nav controller in the situation where both the list and detail are showing.

If just the list is showing, then instead of calling navigate on the nested nav controller, we completely tear-down and recreate the nested nav controller and the nested `NavHost` via a `key` on a unique id. Then, the start destination of the nested nav host becomes where we were trying to navigate to.

Since we were popping the nested nav controller anyway, we don't mind that we completely wipe the state of the nested nav host in this situation.


app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
```diff
@@ -76,17 +83,38 @@ internal fun InterestsListDetailScreen(
...
     BackHandler(listDetailNavigator.canNavigateBack()) {
         listDetailNavigator.navigateBack()
     }
 
-    val nestedNavController = rememberNavController()
+    var nestedNavHostStartDestination by remember {
+        mutableStateOf(selectedTopicId?.let(::createTopicRoute) ?: TOPIC_ROUTE)
+    }
+    var nestedNavKey by remember { mutableStateOf(UUID.randomUUID()) }
+    val nestedNavController = key(nestedNavKey) {
```
jdkoren, line 109: I think you can use `nestedNavHostStartDestination` instead and remove `nestedNavKey`? (They are both only changed at the same time.)

alexvanyo: There's an edge case where that doesn't work if `nestedNavHostStartDestination` doesn't change:

On a compact screen, where just one pane is showing, if we tap on "Compose", the `nestedNavHostStartDestination` will be the "Compose" starting route. If we then tap on the "Architecture" chip from an article on the "Compose" details pane, we'll navigate with the `nestedNavController` to "Architecture". If we then navigate back, we'll end up back on the list screen. If we then click on "Compose" again, the `nestedNavHostStartDestination` won't change, so we won't recreate the `nestedNavController` and nested `NavHost`, so we'll end up at the old state of "Architecture".

jdkoren (approved):

PR #1381, opened 2024-04-16
Enable strong skipping

### LINKED ISSUE #1353: [FR]: Minimize Unnecessary Recompositions

bentrengrove:
**What I have done and why**

Enabled [strong skipping mode](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/compiler/design/strong-skipping.md) in the Compose compiler. This mode is no longer experimental (it is renamed in the next release with no changes) and so the timing seemed right to enable it here.

I have also marked the entire data module as stable. An alternative to this would be to run the Compose compiler on the data module. Both options are fine, I decided to go with the stability config because it was already set up in the project. The effect of this is minimal, but it did improve the recomposition behaviour of bookmarking at item on the ForYou screen. Previously this would recompose every item, now only the row that is bookmarked is recomposed.

Fixes #1353

**How I'm testing it**
Unit tests
Performance Benchmarks

_Benchmark 1 - Interest Screen - Toggle Topic_
| No Strong Skipping | P50 | P90 | P95 | P99 |
| ---------------- | ---- | ---- | ---- | ---- |
| frameDurationCpuMs | 3.84 | 10.2 | 11.13 | 13.23 |

| With Strong Skipping | P50 | P90 | P95 | P99 |
| ---------------- | ---- | ---- | ---- | ---- |
| frameDurationCpuMs | 3.8 (+1%) | 7.72 (-24.3%) | 8.89 (-20.1%) | 11.91 (-10%) |

### DIFF
--- a/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
+++ b/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
@@ -57,6 +57,7 @@ internal fun Project.configureAndroidCompose(
         kotlinOptions {
             freeCompilerArgs += buildComposeMetricsParameters() 
             freeCompilerArgs += stabilityConfiguration()
+            freeCompilerArgs += strongSkippingConfiguration()
         }
     }
 }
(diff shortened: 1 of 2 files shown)


mlykotom (approved):
I think in addition to frame timing metrics, it would be interesting to have a `TraceSectionMetric` in the benchmark with the news composable or `SingleTopicButton` to show it doesn't recompose at all anymore.

But I think that's just to showcase :)

PR #1383, opened 2024-04-17
Update to navigation 2.7.7

alexvanyo:
Updates navigation to 2.7.7.

This required a strange workaround to how we applied a `Modifier` to `NiaNavHost`. Without this workaround, we would loose saved instance state in all of the screens.

I'm not sure why the workaround works, but reported https://issuetracker.google.com/338478720


app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -199,23 +200,26 @@ internal fun NiaApp(
...
-                    appState = appState,
-                    onShowSnackbar = { message, action ->
-                        snackbarHostState.showSnackbar(
-                            message = message,
-                            actionLabel = action,
-                            duration = Short,
-                        ) == ActionPerformed
-                    },
+                Box(
                     modifier = if (shouldShowTopAppBar) {
```
alexvanyo, line 204: This should be equivalent: applying a `Modifier` directly or wrapping it in a `Box` with the `Modifier` applied. For some reason though, the old way would cause the `currentCompositeKeyHash` to change in the inner code, and therefore wipe away saved instance state.

dturner (approved):

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
dturner: Very strange. Thanks for investigating and implementing this workaround.

Issue #1394, opened 2024-04-23
SnackbarScreenshotTests are flaky


JoseAlcerreca:
https://github.com/android/nowinandroid/pull/1173 and https://github.com/android/nowinandroid/pull/1060 required multiple tries to pass. This is a tracking bug.

JoseAlcerreca:
All tests pass locally 10/10 times and in CI 1/4 times. It's strange because it always fails on `./gradlew testDemoDebug :lint:test` but `./gradlew verifyRoborazziDemoDebug` was always run beforehand.

JoseAlcerreca:
![snackbar_expanded_expanded_compare](https://github.com/android/nowinandroid/assets/609125/d6819269-3aff-4e00-b522-643f77080a8b)

JoseAlcerreca:
(The image on the left is transparent)

Jaehwa-Noh:
It seems this problem has been resolved.

PR #1395, opened 2024-04-23
Fixes flakiness in app's screenshot tests

JoseAlcerreca:
We were disabling WorkManager inside the SyncManager instead of disabling SyncManager altogether.  This was throwing some SQLite errors (that are printed out but don't fail the test)

This is the first change to address #1394.

We might want to move that TestSyncModule to another module in the future.

This also enables LocalInspectionMode to prevent the flakiness and pins the timezone.


keyboardsurfer (approved):

app/src/testDemo/kotlin/com/google/samples/apps/nowinandroid/ui/TestSyncModule.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Test Module to avoid WorkManager initialization in tests.
+ */
+@Module
+@TestInstallIn(
+    components = [SingletonComponent::class],
+    replaces = [SyncModule::class],
+)
+abstract class TestSyncModule {
```
Jaehwa-Noh, line 38: Maybe you can use this [one](https://github.com/android/nowinandroid/blob/7980c1d0dfde13d0f4c6ece581589b1ee45a501c/sync/sync-test/src/main/kotlin/com/google/samples/apps/nowinandroid/core/sync/test/TestSyncModule.kt#L31) instead creates new one.

```kotlin
@Module
@TestInstallIn(
    components = [SingletonComponent::class],
    replaces = [SyncModule::class],
)
internal interface TestSyncModule {
    @Binds
    fun bindsSyncStatusMonitor(
        syncStatusMonitor: NeverSyncingSyncManager,
    ): SyncManager
}
```

JoseAlcerreca: Good catch, I didn't know it existed. Also, I don't think it was used anywhere.

keyboardsurfer (approved):

keyboardsurfer (approved):

PR #1398, opened 2024-04-24
Revert "Configure baseline profiles in the baselineProfile lambda"

JoseAlcerreca:
Reverts android/nowinandroid#1060


JoseAlcerreca:
Reverted because it was adding baseline profile generation to unrelated tasks, making them slow and flaky.

dturner (approved):

PR #1403, opened 2024-04-25
Update `SyncWorker` to show notifications for foreground service

yongsuk44:
**What I have done and why**

Fix to ensure notifications are visible for foreground synchronization of Topic and News Data.

### DIFF
--- a/sync/work/src/main/kotlin/com/google/samples/apps/nowinandroid/sync/workers/SyncWorker.kt
+++ b/sync/work/src/main/kotlin/com/google/samples/apps/nowinandroid/sync/workers/SyncWorker.kt
@@ -67,6 +67,7 @@ internal class SyncWorker @AssistedInject constructor(
         traceAsync("Sync", 0) {
             analyticsHelper.logSyncStarted()
 
+            setForeground(getForegroundInfo())
             syncSubscriber.subscribe()
 
             // First sync the repositories in parallel


sync/work/src/main/kotlin/com/google/samples/apps/nowinandroid/sync/workers/SyncWorker.kt
```diff
@@ -67,6 +67,7 @@ internal class SyncWorker @AssistedInject constructor(
         traceAsync("Sync", 0) {
             analyticsHelper.logSyncStarted()
 
+            setForeground(getForegroundInfo())
```
SimonMarquis, line 70: [CoroutineWorker#setForeground(ForegroundInfo)](https://developer.android.com/reference/kotlin/androidx/work/CoroutineWorker#setforeground)

> Makes the `CoroutineWorker` run in the context of a foreground `android.app.Service`. This is a suspending function unlike the `setForegroundAsync` API which returns a `ListenableFuture`.
> Calling `setForeground` will throw an `IllegalStateException` if the process is subject to foreground service restrictions. Consider using `WorkRequest.Builder.setExpedited` and `getForegroundInfo` instead.

Are we sure we want this since we are already using `setExpedited`?

yongsuk44: @SimonMarquis 

In versions of Android prior to Android 12, overrid the `getForegroundInfo()` method would automatically display a notification when a `CoroutineWorker` was running in the foreground. However, from Android 12 onwards, this notification is no longer shown. This change could make it difficult for users to be aware that a foreground work is actively running. - [link](https://developer.android.com/develop/background-work/background-tasks/persistent/getting-started/define-work#backwards-compat)

Therefore, `setForeground()` was add to explicitly notify users of ongoing foreground works.

SimonMarquis: I feel like this becomes redundant with the main loading indicator on the home page. Let's see what the maintainers expect 👍

PR #1407, opened 2024-04-27
Regenerate SVG graphs and cleanup/reformat files

SimonMarquis:
Graphviz generates unnecessary comments when targeting SVG output:
- graph random title
- duplicated node titles
- graphviz version (we only care about SVG result)

This PR updates all SVG files with the latest Graphviz 10.0.1 and reformat these files.


dturner (approved):
Thanks for doing this.

PR #1413, opened 2024-05-03
Add type-safe navigation

dturner:
**Overview**
[Navigation 2.8.0](https://developer.android.com/jetpack/androidx/releases/navigation#2.8.0) introduces type-safe APIs for the Navigation DSL. Instead of using strings for routes and argument names, you can define types which are verified at compile-time. [This blog post](https://medium.com/androiddevelopers/type-safe-navigation-for-compose-105325a97657) explains more. 

**What have I done and why**
Migrated the navigation code from strings to types. For example, here's the old navigation code for the topic feature: 

```
const val TOPIC_ID_ARG = "topicId"
const val TOPIC_ROUTE = "topic_route"

fun NavController.navigateToTopic(topicId: String, navOptions: NavOptionsBuilder.() -> Unit = {}) {
    val encodedId = URLEncoder.encode(topicId, URL_CHARACTER_ENCODING)
    val newRoute = "$TOPIC_ROUTE/$encodedId"
    navigate(newRoute) {
        navOptions()
    }
}

fun NavGraphBuilder.topicScreen(
    showBackButton: Boolean,
    onBackClick: () -> Unit,
) {
    composable(
        route = "topic_route/{$TOPIC_ID_ARG}",
        arguments = listOf(
            navArgument(TOPIC_ID_ARG) { type = NavType.StringType },
        ),
    ) {
        TopicScreen(...)
    }
}
```

And here's the new type-safe code: 

```
@Serializable data class Topic(val id: String)

fun NavController.navigateToTopic(topicId: String, navOptions: NavOptionsBuilder.() -> Unit = {}) {
    navigate(route = Topic(topicId)) {
        navOptions()
    }
}

fun NavGraphBuilder.topicScreen(
    showBackButton: Boolean,
    onBackClick: () -> Unit,
) {
    composable<Topic> {
        TopicScreen(...)
    }
}
```

Note that there's no need to URL encode the arguments because there's no danger of them being interpreted as placeholders (e.g. "{topicId}") in the route string (because there is no route string). 

I have also updated the app manifest to allow deeplinks to be tested from the terminal. Example: `adb shell am start -a android.intent.action.VIEW -d "https://www.nowinandroid.apps.samples.google.com/foryou/2" com.google.samples.apps.nowinandroid.demo.debug`


app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
```diff
@@ -26,47 +26,36 @@ import androidx.compose.material3.adaptive.navigation.rememberListDetailPaneScaf
...
     }
 }
 
 @Composable
-internal fun InterestsListDetailScreen(
-    viewModel: Interests2PaneViewModel = hiltViewModel(),
-) {
-    val selectedTopicId by viewModel.selectedTopicId.collectAsStateWithLifecycle()
+internal fun InterestsListDetailScreen() {
+    var selectedTopicId: String? by rememberSaveable { mutableStateOf(null) }
```
alexvanyo, line 55: This always initializes the `selectedTopicId` to `null` to begin with. If we navigated to the interests with a non-null topic id already selected, I think we'd want to use that here instead?

dturner: Good spot. I've fixed this, and refactored this stateless Composable into the destination declaration above to reduce noise.

feature/interests/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/interests/InterestsViewModel.kt
```diff
@@ -19,47 +19,43 @@ package com.google.samples.apps.nowinandroid.feature.interests
...
+            initialValue = InterestsUiState.Loading,
+        )
 
     fun followTopic(followedTopicId: String, followed: Boolean) {
         viewModelScope.launch {
             userDataRepository.setTopicIdFollowed(followedTopicId, followed)
         }
     }
-
-    fun onTopicClick(topicId: String?) {
```
alexvanyo, line 60: I don't think this is safe to remove?

This function was updating the selected topic id, so that we know which one to highlight in the list.

Now, the value that is passed through will always be the one from the initial `InterestsDestination`, so I'd expect that when showing both the list and detail together, the highlighted topic won't change properly.

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
```diff
@@ -19,28 +19,24 @@ package com.google.samples.apps.nowinandroid.feature.foryou.navigation
...
+@Serializable data class ForYouDestination(val linkedNewsResourceId: String? = null)
+
+fun NavController.navigateToForYou(navOptions: NavOptions) = navigate(route = ForYouDestination(), navOptions)
 
 fun NavGraphBuilder.forYouScreen(onTopicClick: (String) -> Unit) {
-    composable(
-        route = FOR_YOU_ROUTE,
+    composable<ForYouDestination>(
         deepLinks = listOf(
             navDeepLink { uriPattern = DEEP_LINK_URI_PATTERN },
```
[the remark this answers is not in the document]
dturner: Yes, another good spot. The name of the argument placeholder in the deeplink currently has to be identical to the property in the `ForYouDestination`. That duplication needs to be removed. 

The new API offers the following: 

```
navDeepLink<ForYouDestination>(basePath = DEEP_LINK_URI_PATTERN),
```

However, I'm not sure how the mapping works between the serializable data class property `ForYouDestination.linkedNewsResourceId` and the deeplink URI format (currently `https://www.nowinandroid.apps.samples.google.com/foryou/{linkedNewsResourceId}`). 

In our specific case it doesn't actually matter, as long as it works, because the only place it's used is when creating a PendingIntent for a notification (in `SystemTrayNotifier`) so there's no external dependency on the deeplink URI pattern. However, I can imagine that in other apps it'd be difficult to change the deeplink URI pattern so we should figure out how to do this migration.  

I will ask the Jetpack team to clarify guidance here. Incidentally, the AndroidX sample has the same issue: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-compose/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/demos/NavByDeepLinkDemo.kt;l=54 `userId` is both a string in the URI, and a property of `Dashboard`.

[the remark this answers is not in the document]
alexvanyo: Kotlin Serialization [supports](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#serial-field-names) having a different serialized value from the property name with `@SerialName`, which I would assume is the mechanism we could make use of here if they didn't match.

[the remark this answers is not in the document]
claraf3: The [kdoc here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-common/src/main/java/androidx/navigation/NavDeepLink.kt;l=469) explains how the final route is generated for deeplinks. 

In this case, since `linkedNewsResourceId` has a default value (null), it would be appended as a query parameter. 

You'd probably want `DEEP_LINK_URI_PATTERN` to be `https://www.nowinandroid.apps.samples.google.com/foryou/`, and you would end up with this deeplink
`https://www.nowinandroid.apps.samples.google.com/foryou?linkedNewsResourceId={linkedNewsResourceId}`

On a side note, please feel free to let me know how we might improve the kdocs for this. I want to make sure people understand what the final deeplink looks like.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt
```diff
@@ -103,15 +86,15 @@ internal fun InterestsListDetailScreen(
         detailPane = {
             NavHost(
                 navController = nestedNavController,
-                startDestination = TOPIC_ROUTE,
-                route = DETAIL_PANE_NAVHOST_ROUTE,
+                startDestination = TopicPlaceholderDestination::class,
```
claraf3, line 89: StartDestination can be a `KClass<T>` or an `object instance of T`, so this can be simplified to just `TopicPlaceholderDestination`.

app/build.gradle.kts
```diff
@@ -25,6 +25,10 @@ plugins {
     id("com.google.android.gms.oss-licenses-plugin")
     alias(libs.plugins.baselineprofile)
     alias(libs.plugins.roborazzi)
+    // Serialization is used for type-safe navigation.
```
SimonMarquis, line 28: I don't think this is true, it works fine since you are in a regular Gradle module:

```kotlin
plugins {
 alias(libs.plugins.kotlin.serialization)
}
```

```toml
[plugins]
kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
```

app/src/main/AndroidManifest.xml
```diff
@@ -44,10 +44,14 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
-            <intent-filter>
-                <data
-                    android:scheme="https"
-                    android:host="www.nowinandroid.apps.samples.google.com" />
+            <intent-filter android:autoVerify="true">
```
SimonMarquis, line 47: I would advise against adding this parameter unless the Digital Asset Link is correctly setup.
This endpoint currently does not exists: https://www.nowinandroid.apps.samples.google.com./.well-known/assetlinks.json
The side effect of this would be to disable all currently associated deeplinks of this app (althrough there are none at the moment).
This also means that the proper flow -where the disambiguation dialog is not shown- is not allowed.

build-logic/convention/src/main/kotlin/AndroidFeatureConventionPlugin.kt
```diff
@@ -28,6 +28,10 @@ class AndroidFeatureConventionPlugin : Plugin<Project> {
             pluginManager.apply {
                 apply("nowinandroid.android.library")
                 apply("nowinandroid.android.hilt")
+                // Serialization is used for type-safe navigation.
+                // TODO: Use the plugin ID from the version catalog when
+                //  https://github.com/gradle/gradle/issues/15383# is resolved
+                apply("org.jetbrains.kotlin.plugin.serialization")
```
SimonMarquis, line 34: I'm not sure this comment is needed here, otherwise we should also comment all these

`add("implementation", libs.findLibrary("...").get())`

since they are not technically "type-safe".
Also, we already do exactly that in other build-logic plugins:

https://github.com/android/nowinandroid/blob/afad1b9ac7e3d5efc798e786bd6548229a55735c/build-logic/convention/src/main/kotlin/AndroidApplicationComposeConventionPlugin.kt#L26

https://github.com/android/nowinandroid/blob/afad1b9ac7e3d5efc798e786bd6548229a55735c/build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt#L33-L36

feature/topic/src/test/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicViewModelTest.kt
```diff
@@ -60,7 +59,8 @@ class TopicViewModelTest {
     @Before
     fun setup() {
         viewModel = TopicViewModel(
-            savedStateHandle = SavedStateHandle(mapOf(TOPIC_ID_ARG to testInputTopics[0].topic.id)),
+            // TODO: Figure out how to supply the correct dependency TopicDestination(id = testInputTopics[0].topic.id)
```
Jaehwa-Noh, line 62: This is ViewModel Unit test, so that I don't think here is the place to consider Navigation Args.
Manual insert `SavedStateHandle` looks fine to me.

feature/interests/src/test/kotlin/com/google/samples/apps/nowinandroid/interests/InterestsViewModelTest.kt
```diff
@@ -55,7 +54,8 @@ class InterestsViewModelTest {
     @Before
     fun setup() {
         viewModel = InterestsViewModel(
-            savedStateHandle = SavedStateHandle(mapOf(TOPIC_ID_ARG to testInputTopics[0].topic.id)),
+            // TODO: Figure out how to supply the correct dependency: InterestsDestination(topicId = testInputTopics[0].topic.id)
```
Jaehwa-Noh, line 57: This is ViewModel Unit test, so that I don't think here is the place to consider Navigation Args.
Manual insert `SavedStateHandle` looks fine to me.

feature/topic/src/test/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicViewModelTest.kt
SimonMarquis: I think what he meant was to be able to not rely on the `TOPIC_ID_ARG` hardcoded parameter, and instead being able to generate a `SavedStateHandle` from a `TopicDestination` instance.
There is this extension that takes a `SavedStateHandle` and returns your Destination data class:

```kotlin
public inline fun <reified T> SavedStateHandle.toRoute(): T = serializer<T>().decodeArguments(this)
```

There should probably be a test fixture to do the opposite. The closest I found is in `RouteSerializerKt#generateRouteWithArgs()` but that does not solves the issue entirely.

I suppose the key is based on the Serializable, and therefore could be accessed in tests with `TopicDestination::topicId.name`

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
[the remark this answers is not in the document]
claraf3: Nulls get encoded to a string literal null, so it will be
`https://www.nowinandroid.apps.samples.google.com/foryou?linkedNewsResourceId=null`

feature/interests/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/interests/InterestsViewModel.kt
dturner: You are correct, thanks for spotting this. I had mistakenly assumed that the list did not need to know the currently selected topic id. 

I've reintroduced the `Interests2PanelViewModel` to manage the state for the `InterestsScreen`. The initial topic id is obtained from `InterestsDestination` via `toRoute` and subsequent changes to the selected topic are stored using the previous mechanism of writing to `savedStateHandle`. Not ideal, but the best we could come up with for now.

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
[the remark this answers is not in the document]
dturner: Thanks @claraf3. According to the docs, in order to keep our existing format, which uses _path parameters_, we have to have a destination argument _without_ a default value. 

Changing the destination's constructor to change the deeplink pattern results in some duplicated code at the call site. For example, we now have 2 calls to `ForYouDestination(linkedNewsResourceId=null)` which is an obvious case for a default value, but if I add a default value, the argument is represented as a query parameter which breaks our original format. 

Any advice here?

app/build.gradle.kts
dturner: Oh nice, thanks for spotting that. Fixed.

app/src/main/AndroidManifest.xml
dturner: Noted and removed 👍

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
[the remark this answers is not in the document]
claraf3: What are the consequences of breaking the original format? If it is absolutely crucial to keep it, a workaround is to add a deeplink to the NavDestination with a new class where the `linkedNewsResourceId` does not have a default value.

For example
```
@Serializable
data class ForYouDestinationAlternative(val linkedNewsResourceId: String?)
```

This way you can navigate to the destination with either deeplink:
`https://www.nowinandroid.apps.samples.google.com/foryou?linkedNewsResourceId=null`
`https://www.nowinandroid.apps.samples.google.com/foryou/null`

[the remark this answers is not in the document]
dturner: > What are the consequences of breaking the original format?

None for this app, but could be severe for another app wanting to migrate to use these APIs.

I've updated the PR to include this workaround just to demonstrate that it's possible.

feature/interests/src/test/kotlin/com/google/samples/apps/nowinandroid/interests/InterestsViewModelTest.kt
dturner: As mentioned on the comment for `TopicViewModelTest` this should be addressed in a new API.

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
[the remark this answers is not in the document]
alexvanyo: After online discussion, since the workaround was removed, I think we need to update the corresponding logic in `SystemTrayNotifier` to use the query parameter instead of the path parameter.

feature/topic/src/test/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicViewModelTest.kt
JackEblan: Is there any update to the new API? 
```kotlin
public operator fun SavedStateHandle.Companion.invoke(
    route: Any,
    typeMap: Map<KType, @JvmSuppressWildcards NavType<*>> = emptyMap()
): SavedStateHandle
```

Hopefully, they'll include a documentation for its usage.

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
```diff
@@ -19,29 +19,27 @@ package com.google.samples.apps.nowinandroid.feature.foryou.navigation
...
         deepLinks = listOf(
-            navDeepLink { uriPattern = DEEP_LINK_URI_PATTERN },
-        ),
-        arguments = listOf(
-            navArgument(LINKED_NEWS_RESOURCE_ID) { type = NavType.StringType },
+            navDeepLink<ForYouRoute>(basePath = DEEP_LINK_BASE_PATH),
         ),
     ) {
-        ForYouRoute(onTopicClick)
+        ForYouScreen(onTopicClick)
```
Jaehwa-Noh, line 43: Isn't it need to pass the `linkedNewsResourceId` from `backStackEntry.toRoute<ForYouScreen>` to ForYouScreen() for conducting to the news?

dturner: Good question. Yes, previously `linkedNewsResourceId` was a property of `ForYouRoute`, however, I've removed that and am continuing to use `SavedStateHandle` (which is populated from the hardcoded URI pattern) to store the deep linked news resource ID. This is because we clear the news resource ID after it's been opened so that when the user navigates back from the browser into the app the route stays the same.

feature/topic/src/test/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicViewModelTest.kt
dturner: The issue of `SavedStateHandle.toRoute` causing ViewModel tests to be instrumented (i.e. needing an emulator or Robolectric) is tracked here: https://issuetracker.google.com/349807172. Feel free to upvote if this affects you.

feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt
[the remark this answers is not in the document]
dturner: I've refactored this so that the `linkedNewsResourceId` isn't part of the route and is supplied as a path parameter as it was previously. More details here: https://github.com/android/nowinandroid/blob/dt/nav-safe-args/feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/navigation/ForYouNavigation.kt#L36

jdkoren (approved):

feature/interests/src/test/kotlin/com/google/samples/apps/nowinandroid/interests/InterestsViewModelTest.kt
```diff
@@ -25,20 +26,29 @@ import com.google.samples.apps.nowinandroid.core.testing.repository.TestUserData
...
 
 /**
  * To learn more about how this test handles Flows created with stateIn, see
  * https://developer.android.com/kotlin/flow/test#statein
+ *
+ * These tests use Robolectric because the subject under test (the ViewModel) uses
+ * `SavedStateHandle.toRoute` which has a dependency on `android.os.Bundle`.
+ *
+ * TODO: Remove Robolectric if/when AndroidX Navigation API is updated to remove Android dependency.
+ *  See b/340966212.
```
alexvanyo, line 49: Nit: The full URL is nicer for external reference

```suggestion
 *  See https://issuetracker.google.com/340966212.
```

alexvanyo (approved):
Looks great!

Jaehwa-Noh:
Could you check this PR release mode onto the device or emulator? May the `@Serializable` Route be obfuscated, and makes runtime error. In this, you need to update a proguard rules.

PR #1419, opened 2024-05-06
Fix state loss with workaround

alexvanyo:
This is a continuation of #1383 .

After merging #1380 , the state loss issue appeared again that is due to https://issuetracker.google.com/338478720

This additional workaround doesn't change behavior the direct behavior of consuming insets (and shouldn't have any difference overall) but avoids the state loss issue again.

### DIFF
--- a/app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
+++ b/app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
@@ -201,13 +201,13 @@ internal fun NiaApp(
                 }
 
                 Box(
-                    modifier = if (shouldShowTopAppBar) {
-                        Modifier.consumeWindowInsets(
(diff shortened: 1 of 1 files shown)


dturner (approved):

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -201,13 +201,13 @@ internal fun NiaApp(
...
 
                 Box(
-                    modifier = if (shouldShowTopAppBar) {
-                        Modifier.consumeWindowInsets(
-                            WindowInsets.safeDrawing.only(WindowInsetsSides.Top),
-                        )
-                    } else {
-                        Modifier
-                    },
+                    modifier = Modifier.consumeWindowInsets(
```
dturner, line 204:
```suggestion
                    // Workaround for https://issuetracker.google.com/338478720
                    modifier = Modifier.consumeWindowInsets(
```

PR #1424, opened 2024-05-10
Fix reversed follow/unfollow states in InterestsItem-ToggleButton

yongsuk44:
During the UI test review, I found that the follow and unfollow states in the `contentDescription` of the `InterestsItem-ToggleButton` were reversed, so I corrected them.


yongsuk44:
I misunderstood the section written based on [accessibility](https://developer.android.com/develop/ui/compose/accessibility).
I will close this.

PR #1431, opened 2024-05-12
Loading Wheel flexible number of lines

JackEblan:
_Thanks for submitting a pull request. Please include the following information._

**What I have done and why**

The number of lines in Loading Wheel is set to 12 and there is no way to change this because the rotate degrees for each line drawn is hard-coded to `rotate(degrees = index * 30f)`

**How I'm testing it**

_Choose at least one:_
- Unit tests
- UI tests
- Screenshot tests
- N/A _(provide justification)_

**Do tests pass?**

- [ ] Run `./tools/setup.sh`

### DIFF
--- a/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/LoadingWheel.kt
+++ b/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/LoadingWheel.kt
@@ -54,14 +54,16 @@ import kotlinx.coroutines.launch
 fun NiaLoadingWheel(
     contentDesc: String,
     modifier: Modifier = Modifier,
+    rotationTimeInMillis: Int = 12000,
+    numberOfLines: Int = 12,
 ) {
     val infiniteTransition = rememberInfiniteTransition(label = "wheel transition")
 
     // Specifies the float animation for slowly drawing out the lines on entering
     val startValue = if (LocalInspectionMode.current) 0F else 1F
-    val floatAnimValues = (0 until NUM_OF_LINES).map { remember { Animatable(startValue) } }
+    val floatAnimValues = (0 until numberOfLines).map { remember { Animatable(startValue) } }
     LaunchedEffect(floatAnimValues) {
-        (0 until NUM_OF_LINES).map { index ->
+        (0 until numberOfLines).map { index ->
             launch {
                 floatAnimValues[index].animateTo(
                     targetValue = 0F,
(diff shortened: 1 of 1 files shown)


core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/LoadingWheel.kt
```diff
@@ -54,14 +54,16 @@ import kotlinx.coroutines.launch
 fun NiaLoadingWheel(
     contentDesc: String,
     modifier: Modifier = Modifier,
+    rotationTimeInMillis: Int = 12000,
+    numberOfLines: Int = 12,
```
lihenggui, line 58: It is better to check the validity of these numbers if you want to customize the value. If you pass 0 or negative numbers it would be nonsense.

JackEblan: Maybe let's set a default value if the client pass invalid arguments instead of throwing exceptions.

Jaehwa-Noh:
What situation we need to customize the `rotationTimeInMillis` and `numberOfLines`?

JackEblan:
By making this loading composable customizable like:

- trackThickness
- indicatorColor
- trackColor
- indicatorInset
- indicatorDirectionCircular
_This is from Jetpack Compose progress indicator_

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/LoadingWheel.kt
```diff
@@ -115,8 +117,8 @@ fun NiaLoadingWheel(
             .semantics { contentDescription = contentDesc }
             .testTag("loadingWheel"),
     ) {
-        repeat(NUM_OF_LINES) { index ->
-            rotate(degrees = index * 30f) {
+        repeat(numberOfLines) { index ->
+            rotate(degrees = (360 / numberOfLines * index).toFloat()) {
```
anhtuannd, line 121: It can be ```360f / numberOfLines * index```

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/LoadingWheel.kt
anhtuannd: If numberOfLines is too big, it could be performance problem. @IntRange shows warning only and it doesn't enforce the value between the range.

SimonMarquis:
This is a great contribution, but I don't think this will ever be merged.
Providing a customizable design system is definitely not the primary goal of this project.

PR #1440, opened 2024-05-15
Replace github action `always` into `!cancelled`.

### LINKED ISSUE #1439: [Bug]: Github action some jobs can't cancel

### What happened?

Github action run some jobs had run until complete even been cancelled because of using `always()`.

### Relevant logcat output

_No response_

### Code of Conduct

Jaehwa-Noh:
**What I have done and why**

`Always` is run even workflow had been cancelled.
[documentation](https://docs.github.com/en/actions/managing-workflow-runs/canceling-a-workflow#steps-github-takes-to-cancel-a-workflow-run)
> To cancel the workflow run, the server re-evaluates if conditions for all currently running jobs. If the condition evaluates to true, the job will not get canceled. For example, the condition if: always() would evaluate to true and the job continues to run. When there is no condition, that is the equivalent of the condition if: success(), which only runs if the previous step finished successfully.

And GitHub action recommend use `!cancelled` rather than `always`.
[documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#always)
> [!Warning]
> Avoid using always for any task that could suffer from a critical failure, for example: getting sources, otherwise the workflow may hang until it times out. If you want to run a job or step regardless of its success or failure, use the recommended alternative: if: ${{ !cancelled() }}

And github action in here looked sufficient to use `!cancelled`.

Fix #1439


keyboardsurfer:

Jaehwa-Noh:
@keyboardsurfer I've resolved the conflict. Kindly checking the CI result and merge it.

keyboardsurfer (approved):

keyboardsurfer:
Thanks again for this PR!

PR #1444, opened 2024-05-16
Extract testOptions from compose plugin and make its own plugin

### LINKED ISSUE #1443: Should be divide testOptions from compose plugin.

### What happened?

I don't think the testOptions is the parts of Compose plugin, but it is Robolectric, and some test which needs access android resource.

https://github.com/android/nowinandroid/blob/b85cf7d4766ca84a424700f86a96f5eab0a3cfbd/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt#L48-L53

### Relevant logcat output

_No response_

### Code of Conduct

Jaehwa-Noh:
**What I have done and why**

Make its own plugins.

Fixes #1443

### DIFF
--- a/build-logic/convention/src/main/kotlin/AndroidApplicationTestOptionsConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/AndroidApplicationTestOptionsConventionPlugin.kt
@@ -23,6 +23,8 @@ import org.gradle.kotlin.dsl.getByType
 class AndroidApplicationTestOptionsConventionPlugin : Plugin<Project> {
     override fun apply(target: Project) {
         with(target) {
+            pluginManager.apply("com.android.application")
+
             val extension = extensions.getByType<ApplicationExtension>()
             configureTestOptions(extension)
         }
--- a/build-logic/convention/src/main/kotlin/AndroidApplicationTestOptionsConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/AndroidApplicationTestOptionsConventionPlugin.kt
@@ -1,0 +1,30 @@
+import com.android.build.api.dsl.ApplicationExtension
+import com.google.samples.apps.nowinandroid.configureTestOptions
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.getByType
+
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/TestOptions.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.android.build.api.dsl.CommonExtension
+import org.gradle.api.Project
+
+internal fun Project.configureTestOptions(
+    commonExtension: CommonExtension<*, *, *, *, *, *>,
+) {
+    commonExtension.apply {
+        testOptions {
+            unitTests {
+                isIncludeAndroidResources = true
```
jaelsonwagner, line 28: I would like to suggest keeping the following comment for future reference.
// For Robolectric

jaelsonwagner: What about the `isReturnDefaultValues = true`  statement?

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/TestOptions.kt
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.samples.apps.nowinandroid
+
+import com.android.build.api.dsl.CommonExtension
+import org.gradle.api.Project
+
+internal fun Project.configureTestOptions(
```
jaelsonwagner, line 22: As we are configuring the unitTests, I would suggest to use the following name:
`configureTestOptionsUnitTests`

build-logic/convention/src/main/kotlin/AndroidLibraryTestOptionsConventionPlugin.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+import com.android.build.api.dsl.LibraryExtension
+import com.google.samples.apps.nowinandroid.configureTestOptions
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.configure
+
+class AndroidLibraryTestOptionsConventionPlugin : Plugin<Project> {
```
jaelsonwagner, line 23: As we are configuring the unitTests, I would suggest to use the following name:
`AndroidLibraryTestOptionsUnitTestsConventionPlugin`

build-logic/convention/src/main/kotlin/AndroidApplicationTestOptionsConventionPlugin.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+import com.android.build.api.dsl.ApplicationExtension
+import com.google.samples.apps.nowinandroid.configureTestOptions
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.configure
+
+class AndroidApplicationTestOptionsConventionPlugin : Plugin<Project> {
```
jaelsonwagner, line 23: As we are configuring the unitTests, I would suggest to use the following name:
AndroidApplicationsTestOptionsUnitTestsConventionPlugin

Issue #350, opened 2022-10-21
[FR]: Offline warning snackbar is too naggy


dturner:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

We have a snackbar which is constantly on when the user is offline. This is probably a bit too naggy.

### Describe the solution

We should consider only showing the snackbar for a few seconds and/or allowing the user to dismiss the snackbar. This frees up screen real estate for the user to browse the offline content.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tunjid:
The UX supporting the persistent snackbar is that it's easy to forget you're offline in an offline first app as the app performs fine without internet access. It may not be easily apparent that certain features are unavailable when you're offline.

Apps that show a persistent piece of UI when offline include:

Chase Bank
Spotify
GameChanger

I think we should keep it persistent, maybe we can make it less prominent?

calren:
+1 to keeping it persistent somehow. I think it'll be an accessibility issue to have content that automatically goes away.

SimonMarquis:
Should we close this issue if we feel like keeping the SnackBar displayed is important for accessibility?

PR #1090, opened 2023-12-10
Handle `ActivityNotFoundException` when launching Chrome custom tab

SimonMarquis:
Fixes #1068

[Screen_recording_20231210_133959.webm](https://github.com/android/nowinandroid/assets/1921278/3969b588-96e8-424f-9cc6-8b13e60c2835)

PR #1457, opened 2024-05-20
Create a general error handler which handles multiple errors - Nia 1435

### LINKED ISSUE #1435: Create a general error handler which handles multiple errors
Introduce a general error handler which displays errors using a SnackBar one after each other ([following the Material guidelines](https://m3.material.io/components/snackbar/guidelines#84cd5bb6-354f-4fd0-bdd9-daf9684ba4bf)).

Context: https://github.com/android/nowinandroid/pull/1090

tjmtic:
Fixes #1435  

Simple and Generic Error Handler, with a few extensible error types and an OFFLINE error (per reference to #350 ). Not exactly sure if there was something more specific to be implementing or errors to cover, so will be curious if there is a different approach here.

Included in core/ui due to the dependency reference to SnackbarHostState.

The errorHandler can be passed down through composables from NiaApp, the same way as onShowSnackbar, and then called when necessary by the ui. I am not confident in the instantiation in NiaApp, but am not sure it would be possible before then since the handler requires the snackbarHostState in this implementation.


core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/SnackbarErrorHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+        // Log the error or show a generic error message
+        when (error) {
+            is Specific -> {
+                snackbarHostState.showSnackbar(error.data.toString())
+            }
+            is Exception -> {
+                snackbarHostState.showSnackbar("Exception: ${error.exception.message}")
+            }
+            is Offline -> {
+                snackbarHostState.showSnackbar("No internet connection")
```
lihenggui, line 37: Strings should be in the `strings.xml` file

anhtuannd:
But still only one error can be displayed at a time. When the device is offline, will other errors be shown and replaced offline error?

app/src/main/kotlin/com/google/samples/apps/nowinandroid/navigation/NiaNavHost.kt
```diff
@@ -39,6 +40,7 @@ import com.google.samples.apps.nowinandroid.ui.interests2pane.interestsListDetai
 fun NiaNavHost(
     appState: NiaAppState,
     onShowSnackbar: suspend (String, String?) -> Boolean,
+    snackbarErrorHandler: SnackbarErrorHandler,
```
anhtuannd, line 43: This is unused.

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/SnackbarErrorHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import androidx.compose.material3.SnackbarHostState
+import com.google.samples.apps.nowinandroid.core.ui.SnackbarError.Default
+import com.google.samples.apps.nowinandroid.core.ui.SnackbarError.Exception
+import com.google.samples.apps.nowinandroid.core.ui.SnackbarError.Offline
+import com.google.samples.apps.nowinandroid.core.ui.SnackbarError.Specific
+
+class SnackbarErrorHandler(private val snackbarHostState: SnackbarHostState) {
+
+    suspend fun handleError(error: SnackbarError<*>) {
```
anhtuannd, line 27: It can be top level function

anhtuannd (commented):
I think the PR is not completed yet.

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/SnackbarErrorHandler.kt
tjmtic: Yes, moved these, thank you.

app/src/main/kotlin/com/google/samples/apps/nowinandroid/navigation/NiaNavHost.kt
tjmtic: Yes, removed unused, thank you.

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/SnackbarErrorHandler.kt
tjmtic: Thank you, updated this implementation.

tjmtic:
Going to close this PR. 

After looking more at the [ comment in the linked issue ](https://github.com/android/nowinandroid/pull/1090#discussion_r1433979886)
I think I see what the intended approach was and will update with a new PR.

PR #1460, opened 2024-05-21
Remove the printModulePaths task and read modules direct from settings

dturner:
**What I have done and why**
I've updated the `generateModuleGraphs.sh` script to remove the dependency on the `printModulePaths` gradle task. This task was unnecessarily run during the gradle configuration phase and has now been removed. Instead, the bash script just reads the module paths direct from `settings.gradle.kts`. 

Also referenced here: https://github.com/android/nowinandroid/pull/1293

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -49,13 +49,3 @@ plugins {
     alias(libs.plugins.room) apply false
     alias(libs.plugins.module.graph) apply true // Plugin applied to allow module graph generation
 }
-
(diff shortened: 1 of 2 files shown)


Jaehwa-Noh:
There are different way to define include projects. 
`include(":app", ":core:common", ":core:database", ":feature:foryou", ":core:datastore")`

This way wouldn't be your regex work properly.
But current project doesn't use that way. so that It looks good choice alternative way to get modules path.

Issue #1465, opened 2024-05-23
[Bug]: Multi back stacks support doesn't restore screen when navigate with system back button


ken-kentan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

When you have the Saved tab open and press the system back button to return to the For you tab, the expected behavior is that the screen previously opened in the For you tab will be displayed. However, instead, the home screen is displayed. Clicking the For you tab again will restore the state.

Expected behavior: When pressing the system back button to return to the For you tab, the screen previously opened in the tab will be restored.

<video src="https://github.com/android/nowinandroid/assets/10807357/06b43af5-e26d-43fa-a98d-1ee36d974ec7" width="300" />

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anhtuannd:
This seems to be fixed by #1470

Issue #1463, opened 2024-05-22
[Bug]: Cross linking navigation doesn't work correctly


dturner:
Steps to repro:

- On the ForYou screen, tap on any topic chip within a news resource card. The topic detail screen opens for that topic. It is in the Interests nav graph. 
- Tap on ForYou in the nav bar/rail

Expected result: 

- The ForYou screen is shown again

Actual result: 

- Nothing happens

jeonhei:
I have the same problem.This problem is caused by the Navigation library and I've never been able to figure it out.

jeonhei:
It worked!Thank a lot!

PR #1470, opened 2024-05-24
Avoid popUpTo and navigateTo same route

anhtuannd:
Fix for #1463, #1465 and #1597

The issue arises when both popUpTo() and navigate() target the same route. This can lead to unexpected behavior in navigation within NIA.

If start destination and target destination are same, we can just set saveState and restoreState to false.

### DIFF
--- a/app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
+++ b/app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
@@ -33,11 +33,8 @@ import com.google.samples.apps.nowinandroid.core.data.repository.UserNewsResourc
 import com.google.samples.apps.nowinandroid.core.data.util.NetworkMonitor
 import com.google.samples.apps.nowinandroid.core.data.util.TimeZoneMonitor
 import com.google.samples.apps.nowinandroid.core.ui.TrackDisposableJank
-import com.google.samples.apps.nowinandroid.feature.bookmarks.navigation.BookmarksRoute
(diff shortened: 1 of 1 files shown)


anhtuannd:
@dturner This PR is ready for review

anhtuannd:
Rebased to latest type-safe navigation

anhtuannd:
Also fixes https://github.com/android/nowinandroid/issues/1597

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
```diff
@@ -142,17 +142,22 @@ class NiaAppState(
     fun navigateToTopLevelDestination(topLevelDestination: TopLevelDestination) {
         trace("Navigation: ${topLevelDestination.name}") {
             val topLevelNavOptions = navOptions {
+                val startDestination = navController.graph.findStartDestination()
+
+                // Avoid popUpTo and navigate to same screen.
+                val shouldNotSavedAndRestored = startDestination.hasRoute(topLevelDestination.route)
```
keyboardsurfer, line 148: You could remove the double negative `!shouldNotSavedAndRestored` by renaming the property to `saveStateForThisDestination`. This would also make the code more legible.

keyboardsurfer:
Thanks for this PR. Once you have addressed the comments and CI passes, I'd like to see this merged into main.

anhtuannd:
This is no longer needed after we migrate to type-safe navigation.

Issue #1471, opened 2024-05-26
[FR]: Suggest changing the UI layer injection method of Analytics Instance in Analytics Module. 


rhkrwngud445:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

Suggest changing the UI layer injection method of Analytics Instance in Analytics Module. 

Instead of using CompositionLocal, how about using Hilt's EntryPoint?

It has these advantages.
1. It has the advantage of reducing MainActivity injection code.
2. Management outside HiltScope can be reduced.
3. Analytics can eliminate compose dependency.

Additionally, in analytics/AnalyticsModule, how about creating an instance as a singleton?

### Describe the solution

Change Injection of CompositionLocal to Hilt's EntryPoint

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1472, opened 2024-05-26
refactor: On ui layer analytics injected by entrypoint

rhkrwngud445:
Issue: #1471

Suggest changing the UI layer injection method of Analytics Instance in Analytics Module. 

Instead of using CompositionLocal, how about using Hilt's EntryPoint?

It has these advantages.
1. It has the advantage of reducing MainActivity injection code.
2. Management outside HiltScope can be reduced.
3. Analytics can eliminate compose dependency.

Additionally, in analytics/AnalyticsModule, how about creating an instance as a singleton?

### DIFF
--- a/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
+++ b/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
@@ -37,7 +37,7 @@ import androidx.compose.ui.tooling.preview.Devices
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.unit.dp
-import com.google.samples.apps.nowinandroid.core.analytics.LocalAnalyticsHelper
+import com.google.samples.apps.nowinandroid.core.analytics.analyticsInstance
 import com.google.samples.apps.nowinandroid.core.designsystem.theme.NiaTheme
 import com.google.samples.apps.nowinandroid.core.model.data.UserNewsResource
 
@@ -62,7 +62,7 @@ fun LazyStaggeredGridScope.newsFeed(
                 contentType = { "newsFeedItem" },
             ) { userNewsResource ->
                 val context = LocalContext.current
-                val analyticsHelper = LocalAnalyticsHelper.current
+                val analyticsHelper = analyticsInstance(context)
                 val backgroundColor = MaterialTheme.colorScheme.background.toArgb()
 
                 NewsResourceCardExpanded(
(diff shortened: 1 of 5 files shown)


core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
```diff
@@ -62,7 +62,7 @@ fun LazyStaggeredGridScope.newsFeed(
                 contentType = { "newsFeedItem" },
             ) { userNewsResource ->
                 val context = LocalContext.current
-                val analyticsHelper = LocalAnalyticsHelper.current
+                val analyticsHelper = analyticsInstance(context)
```
yongsuk44, line 65: When the Composable is recomposed, accessing and retrieving from the Hilt dependency graph can incur costs. How about optimizing this by using `remember`? 🤔

rhkrwngud445: I agree that reducing cost
Thank you for your opinion

[commit](https://github.com/android/nowinandroid/pull/1472/commits/4fad93549a67035614f618ba7f647401992066ff)

yongsuk44:
This is just a chat, but if a Composable that calls `analyticsInstance()` leaves the Composition and later AnalyticsHelper is needed again, it seems necessary to fetch the object from the dependency graph again. In the existing implementation, it is managed by the Activity Lifecycle, so I think there wouldn't be such unnecessary overhead. What do you think?

Additionally, according to the comments in the existing code, `NoOpAnalyticsHelper()` is used to run the Preview. Do you think the modified code can replace it?

rhkrwngud445:
@yongsuk44 Thanks for your feedback

> Additionally, according to the comments in the existing code, `NoOpAnalyticsHelper()` is used to run the Preview. Do you think the modified code can replace it?
I think it is changed by the lambda code so that preview will be OK

> This is just a chat, but if a Composable that calls analyticsInstance() leaves the Composition and later AnalyticsHelper is needed again, it seems necessary to fetch the object from the dependency graph again. In the existing implementation, it is managed by the Activity Lifecycle, so I think there wouldn't be such unnecessary overhead. What do you think?

I also agree that cost existed. 
It was confirmed that there was a difference of more than 5000 NanoTime. I think that it seems to be a trade-off area with the advantages I suggested.

yongsuk44:
1. I think it would be simpler to write `TrackScreenViewEvent` in the `InterestsRoute` rather than hoisting. I'm curious if it's possible to replace the previews of components like `userNewsResourceCardItems` and `newsFeed`.

2. It seems to me that comparing how much the frame changes through `ScrollForYouFeedBenchmark` would make things clearer. What do you think?

rhkrwngud445:
@yongsuk44 
I think improvements can be made with a nested callback structure, but I don't think it's a good structure.

I suggested CompositionLocal because I was wondering why, and I found the answer.
Your suggestions and questions were helpful. Thank you.

dturner:
We cannot merge this PR unless the `Build` workflow completes successfully. Additionally, you have not enabled contributions from maintainers on your branch so I cannot do this manually. 

Please run the following commands: 

```
git commit --allow-empty -m "Retrigger Build workflow"
git push
```

If we don't hear back from you within 1 week this PR will be closed.

PR #1485, opened 2024-05-31
Remove desugaring from this project

Jaehwa-Noh:
**What I have done and why**

We can remove desugaring from this project.
The only usage of desugar API is just here.
https://github.com/android/nowinandroid/blob/85129e4660f7a27c7081f4ac21169d19db89fbb6/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt#L242-L247

When we remove the desugaring, we can remove `androidDesugarJdkLibs` dependency.
Output APK's size is reduced 1MB.
image

And there's no need to use JAVA 11 whether even use the desugaring, the JAVA version bumped at  #1392 PR

### DIFF
--- a/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
+++ b/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
@@ -67,10 +67,7 @@ import com.google.samples.apps.nowinandroid.core.model.data.FollowableTopic
 import com.google.samples.apps.nowinandroid.core.model.data.NewsResource
 import com.google.samples.apps.nowinandroid.core.model.data.UserNewsResource
 import kotlinx.datetime.Instant
-import kotlinx.datetime.toJavaInstant
-import kotlinx.datetime.toJavaZoneId
-import java.time.format.DateTimeFormatter
-import java.time.format.FormatStyle
+import java.text.DateFormat
 import java.util.Locale
 
 /**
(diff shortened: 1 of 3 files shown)


core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -240,11 +237,9 @@ fun NotificationDot(
...
 
 @Composable
-fun dateFormatted(publishDate: Instant): String = DateTimeFormatter
-    .ofLocalizedDate(FormatStyle.MEDIUM)
-    .withLocale(Locale.getDefault())
-    .withZone(LocalTimeZone.current.toJavaZoneId())
-    .format(publishDate.toJavaInstant())
+fun dateFormatted(publishDate: Instant): String = DateFormat
+    .getDateInstance(DateFormat.MEDIUM, Locale.getDefault())
+    .format(publishDate.toEpochMilliseconds())
```
yogurtearl, line 242: `DateFormat` is not thread-safe or immutable and `DateTimeFormatter` is. 

ideally, the formatter would be `remember`'d (and thread-safe and immutable)

Jaehwa-Noh: Thank you to review it.
Let me says my opinion.

I checked the java [documentation](https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html), it said
> Synchronization
>
> Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.

I don't think my DateFormat is called on multiple thread, it's only called on main thread.
And It has been created new DateFormat instance as `NewsResourceCardExpanded` is composed.

alexvanyo:
We use desugaring in a lot more places than just that one with our usage of [`kotlinx-datetime`](https://github.com/Kotlin/kotlinx-datetime) for time APIs on older SDK versions. This PR will crashes at startup on an API 21 device.

Using desugaring adds size to the APK (like any library), but for our minified release it is around 0.2 MB, which seems well worth it to be able to use much better time APIs.

PR #1486, opened 2024-05-31
Fixed issue with module graphs not generating on Linux

neil3000:
**What I have done and why**

In commit e5008c655ea0373ece8c8dfef6e3aa073991249e a condition was added to check if the grep command supports the -P parameter.

However, from my testing at least, this doesn't seem to work as it uses /dev/null to check for "", probably based on the assumption that grep will return positively even if it doesn't find the matching string (here ""), which isn't true when using /dev/null.

At least that is my understanding of the issue, feel free to correct me if it isn't exactly that <3
What is sure however is that it doesn't work on my setup, even though I do have the grep -P option and it works fine (worked fine before e5008c655ea0373ece8c8dfef6e3aa073991249e, and still works when removing the condition).

Instead of doing a `grep -P "" /dev/null` I changed it to a `echo "" | grep -P ""`, the result will still be positive as grep doesn't return an error when no matches are found, but the string isn't ""null"" as with /dev/null, and so the condition doesn't always return false.

### DIFF
--- a/generateModuleGraphs.sh
+++ b/generateModuleGraphs.sh
@@ -26,15 +26,15 @@ if ! command -v dot &> /dev/null
 then
     echo "The 'dot' command is not found. This is required to generate SVGs from the Graphviz files."
     echo "Installation instructions:"
-    echo "  - On macOS: You can install Graphviz using Homebrew with the command: 'brew install graphviz'"
     echo "  - On Ubuntu: You can install Graphviz using APT with the command: 'sudo apt-get install graphviz'"
+    echo "  - On macOS: You can install Graphviz using Homebrew with the command: 'brew install graphviz'"
     exit 1
 fi
 
 # Check for a version of grep which supports Perl regex.
 # On MacOS the OS installed grep doesn't support Perl regex so check for the existence of the
 # GNU version instead which is prefixed with 'g' to distinguish it from the OS installed version.
-    if grep -P "" /dev/null > /dev/null 2>&1; then
+if echo "" | grep -P "" > /dev/null 2>&1; then
     GREP_COMMAND=grep
 elif command -v ggrep &> /dev/null; then
     GREP_COMMAND=ggrep
@@ -114,10 +114,10 @@ echo "$module_paths" | while read -r module_path; do
           -Pmodules.graph.of.module="${module_path}" </dev/null
 
         # Convert to SVG using dot, remove unnecessary comments, and reformat
-        dot -Tsvg "/tmp/${file_name}.gv" |
-          sed 's/<!--/\x0<!--/g;s/-->/-->\x0/g' | grep -zv '^<!--' | tr -d '\0' |
-          xmllint --format - \
-          > "docs/images/graphs/${file_name}.svg"
+                dot -Tsvg "/tmp/${file_name}.gv" |
+                  sed 's/<!--/\x0<!--/g;s/-->/-->\x0/g' | grep -zv '^<!--' | tr -d '\0' |
+                  xmllint --format - \
+                  > "docs/images/graphs/${file_name}.svg"
         # Remove the temporary .gv file
         rm "/tmp/${file_name}.gv"
     fi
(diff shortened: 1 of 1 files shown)


Fe1777:
To address the issue of module graphs not generating on Linux within the Nowinandroid project, we need to ensure that all scripts and configurations are compatible with Linux environments. This could involve checking for case sensitivity issues, line endings, executable permissions, and any specific Linux-related settings that differ from other operating systems.

Here's a possible approach to fixing this issue:

### Step-by-Step Process

1. **Check for Executable Permissions:**
   On Linux systems, scripts often need explicit executable permissions. Ensure all script files related to generating module graphs have the correct permissions.

   ```sh
   chmod +x path/to/your/script.sh
   ```

2. **Handle Line Endings:**
   Line endings differences between Windows and Unix-like systems can cause scripts to fail. Ensure that all script files use Unix line endings (`\n`).

   In Git, you can configure this by setting up `.gitattributes`:

   ```text
   *.sh text eol=lf
   ```

3. **Update the Script for Linux Compatibility:**
   If the script involves any paths, make sure they use Unix path separators (`/`).

4. **Check for Case Sensitivity:**
   File and directory names are case-sensitive on Linux but not on Windows. Verify that all references to files and directories in your scripts exactly match the case used in the filesystem.

5. **Install Required Tools:**
   Ensure that all necessary tools and dependencies are installed in the Linux environment. This might include Java, Gradle, or specific libraries.

6. **Modify the Graph Generation Script:**
   Review and modify the script or Gradle tasks responsible for generating the module graphs to ensure compatibility with Linux.

### Example Fix and Commit

Below is an example modification assuming you have a script called `generateGraph.sh`:

#### Example `generateGraph.sh`:

```bash
#!/bin/bash

# Navigate to the project directory
cd "$(dirname "$0")" || exit

# Ensure the current working directory is correct
pwd

# Run the graph generation command
./gradlew generateModuleGraph
```

#### Making the Script Executable:

```sh
chmod +x path/to/generateGraph.sh
```

#### Commit the Fix:

```sh
git add path/to/generateGraph.sh
git commit -m "Fixed issue with module graphs not generating on Linux by updating script permissions and compatibility."
git push origin fix-graphmodules
```

### Example Pull Request Description

Here's an example of a comprehensive PR description that explains the changes made:

**Title:** Fixed issue with module graphs not generating on Linux

**Body:**
```
## Description

Fixed the issue where module graphs were not generating on Linux systems. The changes include:

- Modified `generateGraph.sh` to ensure compatibility with Unix-like environments.
- Ensured the script has the correct executable permissions.
- Updated `.gitattributes` to enforce Unix line endings for shell scripts.
- Verified all file path references for case sensitivity issues.
- Confirmed necessary tools and dependencies are installed on the Linux system.

## Issue
Resolves #1486

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
- [x] Tested the script on a Linux environment to verify that the module graphs generate correctly.

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
```

### Conclusion

By following these steps, you address the module graph generation issue on Linux by ensuring compatibility with the Linux environment. Ensure the script has executable permissions, correct line endings, and handles path and case sensitivity issues. Once these adjustments are made, commit and push your changes, and finally create a pull request with a detailed description of the changes and testing conducted.

neil3000:
Is that... a Gemini/chatGPT answer from a bot? .-.

Fe1777:
> Is that... a Gemini/chatGPT answer from a bot? .-.
> 
> Anyway, not helpful at all here.

Sad right?... You dont train you chatgpt? For coding?... DL, Ml?... Thats y you getting fired... See my level kiddo...

dturner:
Should be fixed with #1506. Please file an issue if not.

SimonMarquis:
#1506 did not change this grep command, and is still used here

https://github.com/android/nowinandroid/blob/644e618af1aaebbffd7f0b80e24e9e55b692b3c1/generateModuleGraphs.sh#L73-L74

This should probably be restored and merged.
(not sure why the order of the error message has been changed though 😅)

PR #1408, opened 2024-05-01
hilt compiler migration

Abhimanyu14:
**What I have done and why**

Migrating Hilt compiler from "hilt-android-compiler" to "hilt-compiler".

"hilt-android-compiler" is renamed to "hilt-compiler"
https://github.com/google/dagger/releases/tag/dagger-2.29.1

**How I'm testing it**

_Choose at least one:_
- Unit tests
- UI tests
- Screenshot tests
- N/A _(provide justification)_

**Do tests pass?**
- [ ] Run local tests on `DemoDebug` variant: `./gradlew testDemoDebug`
- [ ] Check formatting: `./gradlew --init-script gradle/init.gradle.kts spotlessApply`

**Is this your first pull request?**
- [ ] [Sign the CLA](https://cla.developers.google.com/)
- [ ] Run `./tools/setup.sh`
- [ ] Import the code formatting style as explained in [the setup script](/tools/setup.sh#L40).

PR #1489, opened 2024-06-04
Migrate from "hilt-android-compiler" to "hilt-compiler"

seve-andre:
As stated [here](https://github.com/google/dagger/releases/tag/dagger-2.29.1) and previously merged [here](https://github.com/android/compose-samples/pull/1384), `hilt-compiler` is just a rename of `hilt-android-compiler`. Since the latter will be deprecated in a future release, it is recommended to use `hilt-compiler` going forward.


Jaehwa-Noh:
It seems to be duplicated with #1408

seve-andre:
> It seems to be duplicated with #1408

Thanks, I'm gonna close this one then.

PR #1496, opened 2024-06-08
Add api 21 into CI

### LINKED ISSUE #1490: [FR]: Add android API 21 in CI

### Describe the problem

Some code change not cover the lower android APIs and there's no way to check it on CI.

### Describe the solution

Add API 21 at CI.

### Additional context

In this commit, the android API 23 was removed from CI.
https://github.com/android/nowinandroid/commit/ab78c44e791fd5755cfc7abf27638c2dac73e593
I don't know exactly the history and what happened at that time, but I felt the needs of cover lower API in CI from https://github.com/android/nowinandroid/pull/1485#issuecomment-2150505641 .

If android 21 API was added in CI, we can find crash as early as possible, and can remove the threat.

### Code of Conduct

Jaehwa-Noh:
**What I have done and why**

Cover api 21 with Continuous Integration to detect early crash from `minSdk` api version.

Using `decodeFromString` instead `decodeFromStream` api 21 below version, reference by https://github.com/Kotlin/kotlinx.serialization/issues/2457#issuecomment-1786923342

Fix #1490


keyboardsurfer:

dturner:
The older emulators are notoriously flaky and should not be added to the `Build` workflow. Please remove the API 21 emulator. Please also update the PR description to more accurately describe what this fixes.

Jaehwa-Noh:
@dturner Could you check this PR #1485 
The developers are hard to notice all dependencies and can't ensure that works lowest api 21, be supported in this project,  well.
We can early catch the crash on API 21.

SimonMarquis:
To be honest, I don't think it is worth to have an entire CI pipeline just to catch one specific issue on this specific library that was not properly documented in the first place.
A comment should suffice, or a lint detector if we really want to not reuse this. But [API 21 is like 0.1%](https://github.com/SimonMarquis/Android-Version-Distribution) of active users on the play store...

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/demo/DemoNiaNetworkDataSource.kt
```diff
@@ -57,6 +76,24 @@ class DemoNiaNetworkDataSource @Inject constructor(
     override suspend fun getNewsResourceChangeList(after: Int?): List<NetworkChangeList> =
         getNewsResources().mapToChangeList(NetworkNewsResource::id)
 
+    /**
+     * Convert [InputStream] to [String].
+     */
+    private suspend fun convertStreamToString(inputStream: InputStream): String = withContext(
```
SimonMarquis, line 82: We don't need to reinvent the wheel 😛 
Kotlin already has extensions to `readText()` from input streams.

Jaehwa-Noh:
@SimonMarquis From your comment, why don't we change lowest api  21 to 22 or so?

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/demo/DemoNiaNetworkDataSource.kt
Jaehwa-Noh: Thanks, I'll test it.

Jaehwa-Noh:
When nobody cares lowest api 21, It would be better to shift up the lowest api.

SimonMarquis:
It depends I guess. Sometimes we have to bump the lowest compatible version when using dependencies that are themselves forcing a higher limit. Using the same value as the default Android template from Android Studio might be a good fit (not sure what it is nowadays).
But anyways, this will be the maintainer's choice.

Jaehwa-Noh:
@dturner I am just as missing your point, the purpose of this PR is API 21 on CI makes work.
and the reason of needing API21 CI is this comment.
https://github.com/android/nowinandroid/pull/1496#issuecomment-2553053136

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/demo/DemoNiaNetworkDataSource.kt
Jaehwa-Noh: readText is working.
https://github.com/Jaehwa-Noh/nowinandroid/actions/runs/12426949721/job/34695993732

core/network/src/main/kotlin/JvmUnitTestDemoAssetManager.kt
```diff
@@ -38,4 +39,5 @@ internal object JvmUnitTestDemoAssetManager : DemoAssetManager {
     private val assets = File(properties["android_merged_assets"].toString())
 
     override fun open(fileName: String): InputStream = File(assets, fileName).inputStream()
+    override fun readText(fileName: String): String = File(assets, fileName).readText(StandardCharsets.UTF_8)
```
SimonMarquis, line 42: We don't need to expose an additional api in the AssetManager. Simply use the already existing `open` function, and call either `decodeFromStream` and `decodeFromString(input.bufferedReader().readText())`

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/demo/DemoNiaNetworkDataSource.kt
```diff
@@ -42,13 +43,27 @@ class DemoNiaNetworkDataSource @Inject constructor(
...
         withContext(ioDispatcher) {
-            assets.open(TOPICS_ASSET).use(networkJson::decodeFromStream)
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                assets.open(TOPICS_ASSET).use(networkJson::decodeFromStream)
+            } else {
+                // Use decodeFromString to capability with API 24 below.
+                // https://github.com/Kotlin/kotlinx.serialization/issues/2457#issuecomment-1786923342
+                val topicsJsonString = assets.readText(TOPICS_ASSET)
+                networkJson.decodeFromString(topicsJsonString)
+            }
```
SimonMarquis, line 53:
```suggestion
            assets.open(TOPICS_ASSET).use {
                if (SDK_INT >= N) networkJson.decodeFromStream(it)
                // https://github.com/Kotlin/kotlinx.serialization/issues/2457#issuecomment-1786923342
                else it.bufferedReader().use(BufferedReader::readText).let(networkJson::decodeFromString)
            }
```

core/network/src/main/kotlin/JvmUnitTestDemoAssetManager.kt
Jaehwa-Noh: Thanks I'll remove exposed readText.

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/demo/DemoNiaNetworkDataSource.kt
Jaehwa-Noh: Thank you. I accept your suggestion but not totally same. I made a private function with generic type for two method, `getTopics` and `getNewsResources`.

Jaehwa-Noh:
This code works on the CI with API 21, 22, 23.
https://github.com/Jaehwa-Noh/nowinandroid/actions/runs/12443126186/job/34741972448

Jaehwa-Noh:
@dturner @keyboardsurfer Ready to get review.

dturner (approved):

PR #1498, opened 2024-06-10
Add DeviceConfigurationOverride for insets to allow screenshot testing insets

alexvanyo:
This adds a way to test for #1295, by creating a `DeviceConfigurationOverride.WindowInsets` that allows injecting arbitrary window insets for content under test.

By using this override in host-side screenshot tests to inject a non-zero amount of system bar insets on all sides, we can visually show how the snackbar is not being inset properly, and can be obscured by the system insets.

This is a prototype for the upstream request for a `DeviceConfigurationOverride.WindowInsets`: https://issuetracker.google.com/343474632


Jaehwa-Noh:
I have a question.
I used `WindowInsets.safeDrawing.getBottom` to avoid the System UI in 
https://github.com/android/nowinandroid/pull/1296/files#diff-3fcb081475d70c62cae861a437429902dd21cd40067b7b52ec0c9c5184a1bf06R90-R95
>    val density = LocalDensity.current
     val bottomSystemUiHeight = density.run {
         WindowInsets.safeDrawing
             .getBottom(density = density)
             .toDp()
     }

If I set the windowInsets like that to test, the windowInsets would be set to the safeDrawing?
Is there any way to set SafeDrawing manually?

dturner (approved):

PR #1499, opened 2024-06-11
Refactor chrome tab toolbar dark mode on/off

yongsuk44:
Modified to change the Chrome toolbar color according to the system scheme during dark mode transitions. Although specific colors can be applied using [setColorSchemeParams](https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsIntent.Builder#setColorSchemeParams(int,androidx.browser.customtabs.CustomTabColorSchemeParams)), using the default values should be sufficient.

https://github.com/android/nowinandroid/assets/38935359/042b8863-f0ca-45f3-bbaf-7f26ad42ef87

[After.webm](https://github.com/android/nowinandroid/assets/38935359/a2f10a8a-c9d5-4ab3-802f-55daba2ff10f)

### DIFF
--- a/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
+++ b/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsFeed.kt
@@ -18,8 +18,6 @@ package com.google.samples.apps.nowinandroid.core.ui
 
 import android.content.Context
 import android.net.Uri
-import androidx.annotation.ColorInt
-import androidx.browser.customtabs.CustomTabColorSchemeParams
 import androidx.browser.customtabs.CustomTabsIntent
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.layout.padding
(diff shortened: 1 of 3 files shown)


Jaehwa-Noh:
There's no way to change dark mode for content? Your demo seems changed just status and action bar.

yongsuk44:
> There's no way to change dark mode for content? Your demo seems to change just status and action bar.

@Jaehwa-Noh 
It will change on web pages that support dark mode. :)

https://github.com/android/nowinandroid/assets/38935359/2fa722ff-48ca-4163-9c92-192e79b923ef

Ibodwe:
If I use this method, wouldn't there be inconsistencies when the app selects a light or dark theme instead of the system theme?

yongsuk44:
@Ibodwe 
If it is a WebView that needs to interact with the app, it should indeed change according to the app theme as you mentioned. However, since it is a WebView solely for web pages, it has been implemented as it is now.

dturner:
This is an interesting idea, however, the original code served a purpose: it sets the chrome custom tab toolbar color to the app's background color. It's a way of passing the app's branding to the webview. It shouldn't be removed. Ideally what should happen when day/night mode changes is the custom tab toolbar color should change to whatever the background color of the app is, however, I don't think that's possible. 

I will leave this open just in case it is.

keyboardsurfer:

PR #1502, opened 2024-06-15
Remove extra ProGuard rules after OkHttp update

Goooler:
https://square.github.io/okhttp/changelogs/changelog_4x/#version-4110


dturner:
Thanks for doing this.

PR #1506, opened 2024-06-16
Replace bash commands with proper svgo cli

### LINKED ISSUE #1503: [Bug]: `generateModuleGraphs.sh` seems to be not working

### What happened?

When I run the `generateModuleGraphs.sh`, build was failed and gave merely blank text output.
I don't know Is this error just on me? Nobody get same errors and blank outputs?

image

### Relevant logcat output

```shell
BUILD FAILED in 1s
6 actionable tasks: 2 executed, 4 up-to-date
Configuration cache entry stored.
Error: dot: can't open /tmp/dep_graph_ui_test_hilt_manifest.gv: No such file or directory
-:1: parser error : Document is empty

^
rm: /tmp/dep_graph_ui_test_hilt_manifest.gv: No such file or directory
```

### Code of Conduct

SimonMarquis:
Although this adds another requirement to execute this script, it will prevent OS-specific issues on `sed` and `grep` as described in #1503, and make sure anyone can generate the same SVG output files.

`--pretty` is kept for convenience but can be removed if needed.

Closes #1503


dturner:
Not overly keen on the dependency on node.js introduced by `svgo`, however, this is better than the current situation of platform-specific issues with `sed` and `grep`. Thanks.

dturner (approved):

PR #1513, opened 2024-06-22
Cleanup unused reference in `libs.versions.toml`

SimonMarquis:
Following Kotlin 2 update, the compose compiler version is already set by default to follow Kotlin's versioning.


yongsuk44:
How about removing `androidx-compose-ui-tooling` and `androidx-compose-ui-tooling-preview` as well? 
They also seem to be declared but not used.

image

SimonMarquis:
@yongsuk44 they are still used indirectly:
https://github.com/android/nowinandroid/blob/85129e4660f7a27c7081f4ac21169d19db89fbb6/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt#L42-L43

dturner:
For the benefit of anyone else upgrading, here are the full details: https://android-developers.googleblog.com/2024/04/jetpack-compose-compiler-moving-to-kotlin-repository.html

dturner (approved):

PR #1517, opened 2024-06-27
Apply version catalog in the convention plugins

Jaehwa-Noh:
Apply version catalog in the convention plugins.

### DIFF
--- a/build-logic/convention/src/main/kotlin/AndroidApplicationComposeConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/AndroidApplicationComposeConventionPlugin.kt
@@ -16,6 +16,7 @@
 
 import com.android.build.api.dsl.ApplicationExtension
 import com.google.samples.apps.nowinandroid.configureAndroidCompose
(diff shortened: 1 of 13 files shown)


dturner:
Please explain why this is a good idea. At first glance, the lines like this: 

```
libs.findPlugin("compose").get().get().pluginId
```

look less readable than the original

```
apply(plugin = "org.jetbrains.kotlin.plugin.compose")
```

Jaehwa-Noh:
@dturner I think this is the good for SSOT, and when we would change plugins, we can change name or group in just one place `libs.versions.toml`.
I agreed with you it is hard to read then origin, but I think that we can remove `.get().get()` using extension function. 

Kindly leave your comment here. I'll change or close this PR.
Thank you.

Jaehwa-Noh:
@dturner 
I have improved the style more readable.
It have had more clear context about the source, it comes from version catalog.
```
apply(plugin = libs.getPlugin("android.application"))
```

Is there anything more do I improve?

PR #1518, opened 2024-06-27
Instrumented tests for InterestsListDetailScreen

### LINKED ISSUE #1340: [Testing FR] [feature:interests] Add large screen tests for 2-pane support
Add screenshot tests and explore the possibility of navigation tests (see the different types in :app)

jdkoren:
Tests for user journeys on the InterestsListDetailScreen on both single-pane and two-pane configurations. Fixes #1340

### DIFF
--- a/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/UiTestExtentions.kt
+++ b/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/UiTestExtentions.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.samples.apps.nowinandroid.ui
+
+import androidx.annotation.StringRes
+import androidx.compose.ui.test.junit4.AndroidComposeTestRule
+import kotlin.properties.ReadOnlyProperty
+
+fun AndroidComposeTestRule<*, *>.stringResource(@StringRes resId: Int) =
+    ReadOnlyProperty<Any, String> { _, _ -> activity.getString(resId) }
--- a/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
+++ b/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
@@ -74,12 +74,13 @@ class InterestsListDetailScreenTest {
     // Overrides for device sizes.
     private enum class TestDeviceConfig(widthDp: Float, heightDp: Float) {
         Compact(412f, 915f),
-        Expanded(1200f, 840f);
+        Expanded(1200f, 840f),
+        ;
 
         val sizeOverride = DeviceConfigurationOverride.ForcedSize(DpSize(widthDp.dp, heightDp.dp))
         val adaptiveInfo = WindowAdaptiveInfo(
             windowSizeClass = WindowSizeClass.compute(widthDp, heightDp),
-            windowPosture = Posture()
+            windowPosture = Posture(),
         )
     }
 
--- a/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
+++ b/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
@@ -16,6 +16,7 @@
 
 package com.google.samples.apps.nowinandroid.ui.interests2pane
 
+import androidx.activity.compose.BackHandler
 import androidx.compose.ui.test.DeviceConfigurationOverride
 import androidx.compose.ui.test.ForcedSize
 import androidx.compose.ui.test.assertIsDisplayed
@@ -27,7 +28,6 @@ import androidx.compose.ui.test.performClick
 import androidx.compose.ui.unit.DpSize
 import androidx.compose.ui.unit.dp
 import androidx.test.espresso.Espresso
-import androidx.test.espresso.NoActivityResumedException
 import com.google.samples.apps.nowinandroid.core.data.repository.TopicsRepository
 import com.google.samples.apps.nowinandroid.core.designsystem.theme.NiaTheme
 import com.google.samples.apps.nowinandroid.core.model.data.Topic
@@ -43,7 +43,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TemporaryFolder
 import javax.inject.Inject
-import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
 import com.google.samples.apps.nowinandroid.feature.topic.R as FeatureTopicR
 
 @HiltAndroidTest
@@ -156,10 +156,16 @@ class InterestsListDetailScreenTest {
 
     @Test
     fun expandedWidth_backPressFromTopicDetail_leavesInterests() {
+        var unhandledBackPress = false
         composeTestRule.apply {
             setContent {
                 DeviceConfigurationOverride(override = expandedWidth) {
                     NiaTheme {
+                        // Back press should not be handled by the two pane layout, and thus
+                        // "fall through" to this BackHandler.
+                        BackHandler {
+                            unhandledBackPress = true
+                        }
                         InterestsListDetailScreen()
                     }
                 }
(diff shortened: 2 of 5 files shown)


app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/UiTestExtentions.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package com.google.samples.apps.nowinandroid.ui
+
+import androidx.annotation.StringRes
+import androidx.compose.ui.test.junit4.AndroidComposeTestRule
+import kotlin.properties.ReadOnlyProperty
+
+fun AndroidComposeTestRule<*, *>.stringResource(@StringRes resId: Int) =
+    ReadOnlyProperty<Any, String> { _, _ -> activity.getString(resId) }
```
SimonMarquis, line 24: nit: IMO we should try to keep explicit return types for all public functions

```suggestion
fun AndroidComposeTestRule<*, *>.stringResource(
    @StringRes resId: Int,
): ReadOnlyProperty<Any, String> = ReadOnlyProperty { _, _ -> activity.getString(resId) }
```

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
```diff
@@ -0,0 +1,231 @@
...
+        val adaptiveInfo = WindowAdaptiveInfo(
+            windowSizeClass = WindowSizeClass.compute(widthDp, heightDp),
+            windowPosture = Posture()
+        )
+    }
+
+    @Before
+    fun setup() {
+        hiltRule.inject()
+        composeTestRule.activityRule
```
SimonMarquis, line 89: I don't think this line is needed

```suggestion
```

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    @Test
+    fun expandedWidth_initialState_showsTwoPanesWithPlaceholder() {
+        composeTestRule.apply {
```
SimonMarquis, line 94: I feel like these patterns of `composeTestRule.apply` followed by `setContent` could be replaced with a more simple `composeTestRule.setContent` and also to reduce indentation.


Or even simpler, if all these tests are actually doing more or less the same thing, to extract the setup phase into a private function:

```kotlin
private fun setup(
    config: TestDeviceConfig,
    then: AndroidComposeTestRule<*, *>.() -> Unit,
) = with(composeTestRule) {
    setContent {
        DeviceConfigurationOverride(override = config.sizeOverride) {
            NiaTheme {
                InterestsListDetailScreen(windowAdaptiveInfo = config.adaptiveInfo)
            }
        }
    }
    then()
}
```

to be called like this in the tests themselves:

```kotlin
setup(TestDeviceConfig.Compact) {
    onNodeWithTag(...).assertIsDisplayed()
    ...
}
```

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
```diff
@@ -0,0 +1,231 @@
...
+                with(TestDeviceConfig.Compact) {
+                    DeviceConfigurationOverride(override = sizeOverride) {
+                        NiaTheme {
+                            InterestsListDetailScreen(windowAdaptiveInfo = adaptiveInfo)
+                        }
+                    }
+                }
+            }
+
+            val firstTopic = runBlocking {
```
SimonMarquis, line 219: nit: this could be extracted into a single function to be reused in the 4 tests

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
```diff
@@ -0,0 +1,232 @@
...
+    private val Topic.testTag
+        get() = "topic:${this.id}"
+
+    // Overrides for device sizes.
+    private enum class TestDeviceConfig(widthDp: Float, heightDp: Float) {
+        Compact(412f, 915f),
+        Expanded(1200f, 840f),
+        ;
+
+        val sizeOverride = DeviceConfigurationOverride.ForcedSize(DpSize(widthDp.dp, heightDp.dp))
```
alexvanyo, line 80: Glad that the ability to store this as a property is helpful for the parameterization!

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
```diff
@@ -0,0 +1,232 @@
...
+            onNodeWithText(firstTopic.name).performClick()
+
+            Espresso.pressBack()
+
+            assertTrue(unhandledBackPress)
+        }
+    }
+
+    @Test
+    fun notExpandedWidth_backPressFromTopicDetail_showsListPane() {
```
alexvanyo, line 208: Nit: I think all `notExpandedWidth` be `compactWidth`, which is more descriptive, since `notExpandedWidth` could also mean `mediumWidth`
```suggestion
    fun compactWidth_backPressFromTopicDetail_showsListPane() {
```

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/UiTestExtentions.kt
alexvanyo, line 1: Typo in file name: `UiTestExtentions` -> `UiTestExtensions`

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
alexvanyo: Agreed that we could probably use `= composeTestRule.apply {` to save a level of indention which might be nice.

I'm not opposed to a setup function if it is the same, but since the `expandedWidth_backPressFromTopicDetail_leavesInterests` test has a slightly different setup with the `BackHandler` to test that behavior, trying to have that configurability in a setup would probably just be more complicated.

I personally don't mind some duplication in tests for the ease of reading and simplicitly. If a test fails, each setup, helper, or other utility method adds an additional place where you have to dive deeper and understand what's going on in the test itself, before figuring out what the actual code is doing.

alexvanyo (changes requested):
Overall looks great, just some minor comments.

Also, since I think this fixes #1340, I think you can add that to the description with `Fixes #1340` and once this is merged the issue will be closed automatically.

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
jdkoren: Yeah, it did look a little odd having one test didn't use the helper function. I think the UI setup is small enough that I don't mind repeating it. We can revisit it later if necessary.

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
jdkoren: Agreed, fixed

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
jdkoren: Fixed

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/UiTestExtentions.kt
jdkoren: Fixed

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/UiTestExtentions.kt
jdkoren: Fixed

alexvanyo (approved):
Looks good to me!

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
Hackermrj: Really Mr.jdkoren hiw to do that

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreenTest.kt
jdkoren: Done. The helper returns the whole list because we might add a test which accesses multiple topics.

PR #1522, opened 2024-06-30
Change `set` to `=` operator in Kotlin DSL

Jaehwa-Noh:
**What I have done and why**

From this documentation https://docs.gradle.org/current/userguide/kotlin_dsl.html#kotdsl:assignment, kotlin DSL preferred assign lazily using `=` operator.

> Using the = operator is the preferred way to call set() in the Kotlin DSL.


dturner (approved):
Nice change, thanks!

PR #1528, opened 2024-07-06
Use the version catalog as the source of truth for plugin IDs

davidburstrom:
The IDs stated in the build-logic build script must exactly match the plugin IDs in the version catalog file. This solution reduces the inherent duplication.

To avoid any IDE errors about missing plugin accessors when adding new plugins to the build-logic script, it's recommended to add the new ID in the version catalog first, then invoke Gradle (sync) in order to generate the accessor classes.

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -58,59 +58,59 @@ tasks {
 gradlePlugin {
     plugins {
         register("androidApplicationCompose") {
-            id = "nowinandroid.android.application.compose"
+            id = libs.plugins.nowinandroid.android.application.compose.get().pluginId
             implementationClass = "AndroidApplicationComposeConventionPlugin"
         }
         register("androidApplication") {
-            id = "nowinandroid.android.application"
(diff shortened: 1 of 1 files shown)


build-logic/convention/build.gradle.kts
```diff
@@ -58,59 +58,59 @@ tasks {
 gradlePlugin {
     plugins {
         register("androidApplicationCompose") {
-            id = "nowinandroid.android.application.compose"
+            id = libs.plugins.nowinandroid.android.application.compose.get().pluginId
```
baksha97, line 61: Can we also use these as the single source of truth for all `id` matchers used in the project?

davidburstrom: Which "matchers" are you referring to? In the `plugins {}` blocks?

dturner (approved):
Thanks for doing this 👍

build-logic/convention/build.gradle.kts
dturner: Since there has been no answer from @baksha97 I'm resolving this comment.

PR #1529, opened 2024-07-06
Explain why the plugins get listed in the root project build script

davidburstrom:
It dispels a bit of the magic that happens during a build. See the patch for the thorough explanation.


dturner (approved):
Thanks for the clear explanation

Issue #1267, opened 2024-03-11
[Bug]: Fts Database data increase infinitely


Jaehwa-Noh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

Both TopicFts and NewsResourceFts databases are always inserted new data not replaced.
<img width="1252" alt="newsResourcesFts" src="https://github.com/android/nowinandroid/assets/48680511/88725396-9a90-41ec-86b4-0cb934060681">
<img width="1244" alt="topicsFts" src="https://github.com/android/nowinandroid/assets/48680511/3004337d-8be1-45ac-a071-155618f6f524">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1531, opened 2024-07-06
Fix an uninstall app after instrumented tests problem

### LINKED ISSUE #1530: [Bug]: The app uninstalled after instrumented tests

### What happened?

1. Install app with demoDebug build variant on local device.
2. Run instrumented test on local device.
3. The demoDebug installed app was uninstalled after the instrumented test done.

### Relevant logcat output

_No response_

### Code of Conduct

Jaehwa-Noh:
**What I have done and why**

Fix #1530 

https://issuetracker.google.com/issues/295039976
Add `android.injected.androidTest.leaveApksInstalledAfterRun=true` in `gradle.properties`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -42,3 +42,7 @@ android.defaults.buildfeatures.shaders=false
 
 # Run Roborazzi screenshot tests with the local tests
 roborazzi.test.verify=true
+
+# Prevent uninstall app after instrumented tests
+# https://issuetracker.google.com/issues/295039976
+android.injected.androidTest.leaveApksInstalledAfterRun=true


dturner:
Whilst this retains the original behavior of _not_ uninstalling the `demoDebug` APK if it was previously installed, I wonder whether uninstalling the APK after an instrumented test is actually a problem. 

In the past we've had problems with CI emulators running out of disk space, largely due to APKs not being uninstalled. I don't want to run into that again. The new behavior of uninstalling the demoDebug APK (created for the instrumented tests or not) might actually be better for this project. 

Is there a specific reason why having `demoDebug` uninstalled is a problem?

Jaehwa-Noh:
@dturner Oh, I'd not followed up the out of disk space issue.
My target is making instrumented test app doesn't override the already installed app for using, and remove it after test done.

Why I do is, when we install the app aab or apk on device, and we've use that installed app for a while, and after that we ran the instrumented test on same device, the origin installed app was uninstalled.
we need to preserve installed demoDebug or whatever build variables to still remain on the device, for not install new initialized app again.

So that, I think we need to separate the installed app from the instrumented test app.

If the out of disk space problem is bigger than this, I'll close this PR and the issue.

Jaehwa-Noh:
@dturner 
Maybe the out of the space problem is related with this #1267 issue.
I think the bunch of data had been inserted again and again, over again as it ran on AndroidTest.

dturner (approved):

PR #1540, opened 2024-07-09
[Do not merge] Override WindowMetrics in InterestsListDetailScreenTest

jdkoren:
Creates and uses a test rule to override the WindowMetrics reported by WindowMetricsCalculator with an arbitrary size.

### DIFF
--- a/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/FakeWindowMetricsCalculatorRule.kt
+++ b/app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/FakeWindowMetricsCalculatorRule.kt
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/FakeWindowMetricsCalculatorRule.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val calculator = FakeWindowMetricsCalculator()
+
+    fun setWindowSize(size: Size) {
+        calculator.windowSize = size
+    }
+
+    override fun apply(base: Statement?, description: Description?): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                WindowMetricsCalculator.overrideDecorator(
```
alexvanyo, line 46: (commenting for visibility)

This is a `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` method, so while it may work to call, it's not an API explicitly supported by the library for us to use until that is removed in the library.

PR #1541, opened 2024-07-11
Change display name each flavor on debug

Jaehwa-Noh:
**What I have done and why**

prodDebug and demoDebug is hardly distinguish each app on the screen.
I changed the app name based on the flavor on `debug`, except `release`.

| before | after |
|-------|------|
|Screenshot_20240711_163737|Screenshot_20240711_191553|

Now, we can distinguish both at once glance on the screen.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -37,7 +37,6 @@ kotlin.code.style=official
 
 # Disable build features that are enabled by default,
 # https://developer.android.com/build/releases/gradle-plugin#default-changes
-android.defaults.buildfeatures.resvalues=false
 android.defaults.buildfeatures.shaders=false
 
 # Run Roborazzi screenshot tests with the local tests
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -37,7 +37,6 @@ kotlin.code.style=official
 
 # Disable build features that are enabled by default,
 # https://developer.android.com/build/releases/gradle-plugin#default-changes
-android.defaults.buildfeatures.resvalues=false
```
SimonMarquis, line 40: IMO re-enabling `resValues` for that is not good. We already have specific sourceSet for all flavors.
Also, I'm not sure that Prod variant is actually expected to be installed by regular devs, since it needs production keys etc.
This makes the actual Demo app name appearance worse.

Jaehwa-Noh: @SimonMarquis 
It seems hard it to do with resValue from the error logs. I can do same thing with another way without resValues.
Before refactor that I also wonder `prod` usage, this project just needs the `prodRelease`?

dturner:
If I ever get into the state where I've deployed both `prodDebug` and `demoDebug` to the same device, I just uninstall both and reinstall the one I want to test. It might have happened once a year. Not worth updating the code for.

PR #1545, opened 2024-07-12
Hotfix broken build with a test

### LINKED ISSUE #1544: [Bug]: Build Broken

### What happened?

Removed dependency.

image

### Relevant logcat output

_No response_

### Code of Conduct

Jaehwa-Noh:
**What I have done and why**

Fix #1544


machado001 (commented):
Imho we could add a new line to libs.versions.toml to maintain project consistence.

in libs.version.toml: 

```
[libraries]
//another libs...
kotlin-test = { group = "org.jetbrains.kotlin", name = "kotlin-test", version.ref = "kotlin" }
```

then:

```
dependencies {
    ...
    testImplementation(libs.kotlin.test)
    ...
 
}
```

What do you think?

Jaehwa-Noh:
@machado001 
Thanks for your comment.

From this documentation,
https://kotlinlang.org/docs/jvm-test-using-junit.html#add-dependencies
   
Kotlin DSL
> dependencies {
    // Other dependencies.
    testImplementation(kotlin("test"))
}

Groovy DSL
> dependencies {
    // Other dependencies.
    testImplementation 'org.jetbrains.kotlin:kotlin-test'
}


If we use Groovy DSL, It is good choice to use kotlin-test with version catalog, but we use Kotlin DSL and it gives us kotlin alias `kotlin("test")`

kotlin("test") is more shorten and clear than version catalog, and it reduces that what we have to manage a dependency as we delegate it to kotlin.

machado001:
Hi @Jaehwa-Noh, thank you for your reply.

I agree that using `kotlin("test")` is clean enough. Just to provide another perspective, I suggested adding the library to the version catalog to maintain project consistency and have a centralized place to manage and document dependencies.

New developers can start contributing to the project by looking at the version catalog file to get a general overview of the libraries and plugins used in the project.

I checked another `build.gradle.kts` file, particularly from the `:lint` module, and it looks like `kotlin("test")` is already used there. This could become a recurrent dependency that isn't declared in the version catalog, requiring developers to check each module individually to discover this.

Thus, another advantage is that, from the version catalog, developers can simply Ctrl + click (or Ctrl + B) to check which modules use the dependency.

Jaehwa-Noh:
@machado001 Yes, you have the reason. Simple way is you can found all usage in search all files Ctrl + Shift + F (Find all files).
On this topic, I can't determine which way is better on my own.
Let us hear the other developers and maintainers opinions.

trietbui85:
I would prefer the explicit declaration of `kotlin-test` dependency in `libs.versionss.toml`.

Jaehwa-Noh:
Following #1560, not only fixed #1544 but also replaced `kotlin("test")` to version catalog `kotlin.test` for consistent.

Jaehwa-Noh:
Superseded #1667

Issue #1526, opened 2024-07-03
[Bug]: Scrollbar is hardly stable


Jaehwa-Noh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

Scrollbar is hardly stable.
When I drag down and drag up, you can see the scrollbar position jump other place.
[Screen_recording_20240703_165416.webm](https://github.com/android/nowinandroid/assets/48680511/716d20b6-0603-427e-aabe-4c27d99d2f37)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SimonMarquis:
Interesting, is it only present on large screen devices, or also on regular phone width?

Jaehwa-Noh:
@SimonMarquis 
I'd not tested it on regular and tablet yet.
What I used a device in the video was `Resizable API34` emulator with Foldable size.

anhtuannd:
I think it happens due to there is one spacer at the end of stagger list.
https://github.com/android/nowinandroid/blob/50b13ecb21138b76e14ff0b2f4cf1dcfda9d0f43/feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt#L200
Since ScrollBar (thumb position and thumb size) is calculated based on first visible items and percentage of visible items, not by real position, and all other items are long layout, it causes miscalculating when you scroll to end of the list.
https://github.com/android/nowinandroid/blob/50b13ecb21138b76e14ff0b2f4cf1dcfda9d0f43/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/ScrollbarExt.kt#L204

rodrigodiasferreira:
Hi @anhtuannd , @Jaehwa-Noh , @SimonMarquis ,
I've done some investigation about this issue, the reason for this behavior is indeed the `itemsVisible` and the `Spacer` as commented by @anhtuannd . 
When it is close to the end, and in one of the lanes there is no additional news resource, then the `StaggeredGridLayoutInfo.visibleItemsInfo` list the Spacer, which is way down, which makes the calculation logic of the items visible broke, since the Spacer is out of the viewport. And it happens also for some news resource that are before the viewport and the `StaggeredGridLayoutInfo.visibleItemsInfo` also list them.
[...]

Issue #1549, opened 2024-07-14
[FR]: In the file ScrollbarExt.kt there is room for reducing the logic and code repeating in order to improve the reusability and maintainability


rodrigodiasferreira:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

In the file ScrollbarExt.kt file there is room for reducing the logic and code repeating in order to improve the reusability and maintainability.
Moreover, this improvement will make it better for developers who use the nia sample to study, and learn ways to use Kotlin to increase resusability.
The `scrollbarState` extension function for the LazyListState, LazyGridState, LazyStaggeredGridState have a lot of common logic that could be unified.

### Describe the solution

My suggestion is to unify into a more generic extension function:
`<LazyState : ScrollableState, LazyStateItem> LazyState.genericScrollbarState`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

rodrigodiasferreira:
Hi there, 
I would like to contribute to Now In Android sample.
I've already developed such improvement, but I'm not able to upload it.

I've already signed the Google Individual CLA.
I can upload new PRs to my repositories, but for Now in Android it says to check if I have the correct access rights.

I am trying to upload the branch where I developed the improvement: `scrollbarext-improvement`
`git push --set-upstream origin scrollbarext-improvement`

This is the error I got:
```
ERROR: Permission to android/nowinandroid.git denied to rodrigodiasferreira.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.`
```

Anyone can help me to check if I am doing something wrong?

Thanks.

Jaehwa-Noh:
Follow these steps.
1. Fork this repository into your repository.
2. Open 'your forked repository' on the Android Studio.
3. Edit your code, and push to your forked repository.
4. Make PR at this origin repository.

rodrigodiasferreira:
Hi @Jaehwa-Noh, thanks a lot for the tips.
I've just upload the PR: https://github.com/android/nowinandroid/pull/1550, I mentioned this issue on the PR, but maybe it should be linked through the development right side bar on this issue, but I don't have permission to do that.
I also do not have permission to self assign this issue to me.

It's my first contribution, so if there is any adjustment, or anything that I should improve, please let me know, I will do my best to conform to your standards.
Thanks!

rodrigodiasferreira:
[...]

PR #1550, opened 2024-07-14
Improving the reusability and maintainability in ScrollbarExt.kt file and Fix the scroll bar thumb jumping a little bit due to lack of smothness on interpolatedIndex

rodrigodiasferreira:
**What I have done and why**

Improved the reusability and maintainability of the file ScrollbarExt.kt file by implementing a generic: `<LazyState : ScrollableState, LazyStateItem> LazyState.genericScrollbarState`, and reducing the repetitive logic and code.

Related issue:
- https://github.com/android/nowinandroid/issues/1549
- It also fixes the issue: https://github.com/android/nowinandroid/issues/1526 (the part related to the interpolated index).

### DIFF
--- a/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/ScrollbarExt.kt
+++ b/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/ScrollbarExt.kt
@@ -33,84 +33,6 @@ import kotlinx.coroutines.flow.filterNotNull
 import kotlin.math.min
 
 /**
- * Calculates a [ScrollbarState] driven by the changes in a [LazyState].
- *
- * @param itemsAvailable the total amount of items available to scroll in the [LazyState].
- * @param itemSize a lookup function for the size of an item in the layout.
- * @param offset a lookup function for the offset of an item relative to the start of the view port.
- * @param nextItemOnMainAxis a lookup function for the next item on the main axis in the direction
- *  of the scroll.
- * @param itemSize next [LazyStateItem] on the main axis of [LazyState], null if none.
- * @param index a lookup function for index of an item in the layout relative to
- * @param viewportStartOffset a lookup function for the start offset of the view port
- * @param viewportEndOffset a lookup function for the end offset of the view port
- * @param reverseLayout a lookup function for the reverseLayout of [LazyState].
- */
-@Composable
-internal inline fun <LazyState : ScrollableState, LazyStateItem> LazyState.genericScrollbarState(
-    itemsAvailable: Int,
-    crossinline visibleItems: () -> List<LazyStateItem>,
-    crossinline itemSize: LazyState.(LazyStateItem) -> Int,
-    crossinline offset: LazyState.(LazyStateItem) -> Int,
-    crossinline nextItemOnMainAxis: LazyState.(LazyStateItem) -> LazyStateItem?,
-    crossinline index: (LazyStateItem) -> Int,
-    crossinline viewportStartOffset: () -> Int,
-    crossinline viewportEndOffset: () -> Int,
-    crossinline reverseLayout: () -> Boolean = { false },
-): ScrollbarState {
-    val state = remember { ScrollbarState() }
-    LaunchedEffect(this, itemsAvailable) {
-        snapshotFlow {
-            if (itemsAvailable == 0) return@snapshotFlow null
-
-            val visibleItemsInfo = visibleItems()
-            if (visibleItemsInfo.isEmpty()) return@snapshotFlow null
-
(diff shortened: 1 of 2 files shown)


core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/scrollbar/ScrollbarExt.kt
```diff
@@ -32,42 +33,57 @@ import kotlinx.coroutines.flow.filterNotNull
...
+ *  of the scroll.
+ * @param itemSize next [LazyStateItem] on the main axis of [LazyState], null if none.
+ * @param index a lookup function for index of an item in the layout relative to
+ * @param viewportStartOffset a lookup function for the start offset of the view port
+ * @param viewportEndOffset a lookup function for the end offset of the view port
+ * @param reverseLayout a lookup function for the reverseLayout of [LazyState].
  */
 @Composable
-fun LazyListState.scrollbarState(
+internal inline fun <LazyState : ScrollableState, LazyStateItem> LazyState.genericScrollbarState(
```
machado001, line 50: Is worth to make this function inline? inline keyword isn't recommended for huge functions, since it can make the bytecode to grow

rodrigodiasferreira: Sure. I agree, I removed the inline qualifier, and also made it more restrict the access to this function, from internal to private, since this function is only used in that file.
Thanks for the review @machado001.

yongsuk44:
Sometimes, I believe that dividing similar logic into multiple extension functions is done to reduce the learning curve of the logic.

When I look at the refactored code, it seems that jumping from the extension function to `genericScrollbarState` and then jumping back to the extension function to check arguments could occur. This might lower the readability for developers who are learning the code, and I am concerned that it could increase the learning curve. 🥲

rodrigodiasferreira:
Hi @yongsuk44, thanks for your comment making this discussion richer.
I disagree. When I was studying the Now In Android project, I had to compare the 3 functions in order to check what they had in common, and then I figure it out, it is pretty much the same logic, but with the differences that each type brings: `LazyListState, LazyGridState, LazyStaggeredGridState`. This refactoring unify the logic into a unique function, and makes very clear that they perform the same logic, which makes easier for study in my opinion, because you don't need to read the 3 different functions to realize that they are the same logic.

Moreover, if there will be some need to fix something in that logic in the future, it won't be necessary to change on the 3 different functions, only in one place, which improves also the maintainability. Besides one could even forget to change on these other places, which would cause different behavior for the scroll bar for the different type.

In summay, in my opinion, this refactoring is a good improvement for Now In Android code base in aspects of study, reusability and maintainability.

anhtuannd:
Hi,
I think LazyStaggeredGridState's ScrollBar needs some improvement on how we calculate thumb size and position, as seen in  https://github.com/android/nowinandroid/issues/1526

So merging them into once might not the best approach.

rodrigodiasferreira:
Hi @anhtuannd , thanks for raising this concern.
As I commented on the issue: https://github.com/android/nowinandroid/issues/1526, I believe I've fixed that issue on the PR: https://github.com/android/nowinandroid/pull/1553, and the solution was done in a common place, so the unifying done on this PR, won't affect the fix of that issue. Please check when you have some time, and let me know, if you agree...

tunjid:
Hi hi! Original author of the Scrollbar here. It used to have a generic abstraction, but we split up out for readability.

You can see the original PR where the generic code was split up favoring readability over copy pasting here: https://github.com/android/nowinandroid/pull/947#discussion_r1335920196

rodrigodiasferreira:
Hi TJ / @tunjid , how are you?
I watched several videos that you recorded on the Android Developers site.
It's an honor to meet you!

I understand what you are saying, but on my opinion I still think that the generic abstraction eliminates some duplicated logic and improves the maintainability. And also I prefer to read the more concise code, which I know for sure that the parts that use the same generic abstraction, they are the same, no need to compare the code and checking that they do the same thing.

But since you guys made this decision consciously, I will open a new PR with the fix on the commit: https://github.com/android/nowinandroid/pull/1550/commits/31b77d51ba329297ec39cbac66b73f2202157cce, and close this one.

Thanks a lot for the feedback. :-)

rodrigodiasferreira:
Hi @tunjid,

I've uploaded the commit: https://github.com/android/nowinandroid/commit/31b77d51ba329297ec39cbac66b73f2202157cce into the PR: https://github.com/android/nowinandroid/pull/1553. So the fix related to issue: https://github.com/android/nowinandroid/issues/1526 is concentrated on the same PR.
I will close this one.

Once again, thanks for the feedback.

PR #1556, opened 2024-07-22
Drag and drop

tiwiz:
Add Drag&Drop feature to feed title in both For You and Saved fields

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 accompanist = "0.34.0"
 androidDesugarJdkLibs = "2.0.4"
 # AGP and tools should be updated together
-androidGradlePlugin = "8.5.1"
+androidGradlePlugin = "8.4.0"
 androidTools = "31.4.1"
 androidxActivity = "1.8.2"
 androidxAppCompat = "1.7.0"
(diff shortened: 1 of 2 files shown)


riggaroo:
Possible to add a screen recording?

app/.gitignore
```diff
@@ -1 +1,5 @@
-/build
\ No newline at end of file
+/build
+/.idea/.gitignore
+/.idea/gradle.xml
+/.idea/misc.xml
+/.idea/vcs.xml
```
SimonMarquis, line 5: This is not an expected change, please discard it.

gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 accompanist = "0.34.0"
 androidDesugarJdkLibs = "2.0.4"
 # AGP and tools should be updated together
-androidGradlePlugin = "8.4.0"
+androidGradlePlugin = "8.5.1"
```
SimonMarquis, line 5: Does your change need this?
If yes, we should update it in a distinct review.
If no, discard this change.

SimonMarquis: - https://github.com/android/nowinandroid/pull/1536

tiwiz: Discarded

tiwiz:
Closing PR - will open another one with only drag

PR #1557, opened 2024-07-22
Add drag feature

tiwiz:
Add Drag feature to article title. This will allow for sharing the content across apps.

https://github.com/user-attachments/assets/15e6fbf2-9fe4-4def-9c5b-50748c8c6581

### DIFF
--- a/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
+++ b/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
@@ -16,8 +16,15 @@
 
 package com.google.samples.apps.nowinandroid.core.ui
 
+import android.content.ClipData
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
+import android.view.View
 import androidx.compose.foundation.Canvas
+import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.Image
+import androidx.compose.foundation.draganddrop.dragAndDropSource
+import androidx.compose.foundation.gestures.detectTapGestures
 import androidx.compose.foundation.horizontalScroll
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
@@ -45,6 +52,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draganddrop.DragAndDropTransferData
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalInspectionMode
@@ -77,6 +85,7 @@ import java.util.Locale
  * [NewsResource] card used on the following screens: For You, Saved
  */
 
+@OptIn(ExperimentalFoundationApi::class)
 @Composable
 fun NewsResourceCardExpanded(
     userNewsResource: UserNewsResource,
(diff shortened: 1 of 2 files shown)


Jaehwa-Noh:
This feature is for multi-window user, is there any way to check the window state? For just activate it when user in multi-window.

tiwiz:
There is an API to detect if the app is running in [multi-window environment](https://developer.android.com/reference/android/app/Activity#isInMultiWindowMode()), but I would rather not use it: many OEMs have features (like shelves, or similar) that allow users to Drag and item to a temporary part and drop it once they change app, so only enabling this feature in multi-window would limit those use-cases as well

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -112,7 +128,22 @@ fun NewsResourceCardExpanded(
                     Row {
                         NewsResourceTitle(
                             userNewsResource.title,
-                            modifier = Modifier.fillMaxWidth((.8f)),
+                            modifier = Modifier.fillMaxWidth((.8f))
+                                .dragAndDropSource {
```
alexvanyo, line 132: I wonder if the whole card should be draggable, instead of just the title?

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -112,7 +128,22 @@ fun NewsResourceCardExpanded(
...
+                                    detectTapGestures(
+                                        onLongPress = {
```
alexvanyo, line 134: Is there a test we can add to verify this?

tiwiz: Unfortunately, not. We have no testing APIs for this

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
tiwiz: I'd rather avoid using the whole card, as we have no clear indication of what is draggable and what is not, and it might confuse users when they tap longer on the card

Jaehwa-Noh:
[Screen_recording_20240730_102043.webm](https://github.com/user-attachments/assets/c44d5760-698a-493b-9be9-dd1ea270d17c)

I found a bug, drag and drop works only once. Is this bug origin from drag and drop function?

tiwiz:
Good catch! I think it might have something to do with scrollable container, rather than `draggableSource`, as I was able to init the feature multiple times. Let me investigate further

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
alexvanyo: Is there a bug to track for those?

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -112,7 +128,22 @@ fun NewsResourceCardExpanded(
...
+                                            startTransfer(
+                                                DragAndDropTransferData(
+                                                    ClipData.newPlainText(
+                                                        "Feed sharing",
+                                                        userNewsResource.title + ": " + userNewsResource.url,
```
alexvanyo, line 139: Since this is the string manipulation that the user can see, should this be using string resources to perform the manipulation in case that would need to be translated?

alexvanyo (approved):
Looks good, just one more question about the string resource for the content we show to the user.

core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
tiwiz: That would require Context. Would you be OK in passing it down from the Composable itself?

danybony:
Nice one! LGTM

PR #1559, opened 2024-07-30
Update DeviceConfigurationOverride.WindowInsets to match upstream implementation

alexvanyo:
Updates the local version of `DeviceConfigurationOverride.WindowInsets` to match the implementation of the one in upstream Compose merged in https://android-review.git.corp.google.com/c/platform/frameworks/support/+/3143004

This fixes behavior on devices running SDK 29 and below.


app/src/testDemo/kotlin/com/google/samples/apps/nowinandroid/ui/DeviceConfigurationOverrideWindowInsets.kt
```diff
@@ -17,18 +17,17 @@
 package com.google.samples.apps.nowinandroid.ui
 
 import android.view.WindowInsets
-import android.widget.FrameLayout
```
Phatphatnguyen, line 20: -import androidx.widget.Framelayout

dturner (approved):

PR #1563, opened 2024-08-06
Remove ImageLoader from NiaApplication

AndrewReitz:
Removed ImageLoader from NiaApplication as it's not needed. 
The ImageLoader is provided to the rest of the app using dagger. 

This appears to make it into the app because of the example on the coil website which is how you can use the ImageLoader without a DI framework.

### DIFF
--- a/app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
+++ b/app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
@@ -17,8 +17,6 @@
 package com.google.samples.apps.nowinandroid
 
 import android.app.Application
-import coil.ImageLoader
-import coil.ImageLoaderFactory
 import com.google.samples.apps.nowinandroid.sync.initializers.Sync
 import com.google.samples.apps.nowinandroid.util.ProfileVerifierLogger
 import dagger.hilt.android.HiltAndroidApp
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
```diff
@@ -28,9 +26,7 @@ import javax.inject.Inject
  * [Application] class for NiA
  */
 @HiltAndroidApp
-class NiaApplication : Application(), ImageLoaderFactory {
```
hoc081098, line 31: It is used when we do not specify an ImageLoader in Coil AsyncImage Composable

hoc081098: https://github.com/coil-kt/coil/blob/main/coil-compose%2Fsrc%2FcommonMain%2Fkotlin%2Fcoil3%2Fcompose%2FSingletonAsyncImage.kt#L68-L68

hoc081098: https://github.com/coil-kt/coil/blob/main/coil%2Fsrc%2FcommonMain%2Fkotlin%2Fcoil3%2FSingletonImageLoader.common.kt#L111-L113

Issue #1523, opened 2024-07-02
[Bug]: Quickly switching the bottom tabs can cause the screens of two tabs to overlap


xyz-fly:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

Quickly switching the bottom tabs can cause the screens of two tabs to overlap
![3ef5193d-374a-4fba-8bb4-7753e222787e](https://github.com/android/nowinandroid/assets/4801424/5fcb0c3f-4ade-40a2-a7d8-dedc83d8f163)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

royerdavid:
I experience the same issue.

Songgyubin:
I'm experiencing the same problem in my app, but it seems to be a problem related to navOption.

PR #1573, opened 2024-08-14
Fix Overlap When Moving Between Tabs

Songgyubin:
When you quickly switch between bottom tabs, the screens of the two tabs overlap.
An issue regarding this was also created, and it was resolved by updating the navigation library version.
[Issues raised](https://github.com/android/nowinandroid/issues/1523) 
Screenshot_20240814_221708

The above issue was resolved in navigation 2.8.0-beta04.
[jetpack navigation release note](https://developer.android.com/jetpack/androidx/releases/navigation?hl=ko#2.8.0-beta04)
(Currently, NowInAndroid is using jetpack navigation version 2.8.0-alpha06)

The latest version is 2.8.0-beta07, but compileSdk version 35 is required, so the version was updated to beta06.


Jaehwa-Noh:
I think the resolve is about `Predictive back` feature.

Songgyubin:
@Jaehwa-Noh 
Yes, I read it again and there is a screenshot of the release notes. 
 I removed the highlighted part because it did not seem to fit perfectly with the issue I raised.

However, we have confirmed that the issues raised have been resolved starting with version `beta-04`

I think it was probably fixed along with fixing the bug in the `Predictive back` function. What do you think?

Jaehwa-Noh:
@Songgyubin 
I'm not sure the issue was resolved as I read a release note.

Songgyubin:
@Jaehwa-Noh 

Yes, I tried it again, but there is no information about it in the release notes.

However, when I ran it and tested it, it seemed to have been resolved.

So, to get an accurate understanding, I [commented](https://issuetracker.google.com/issues/338975163#comment11) on the [same post](https://issuetracker.google.com/issues/338975163) as the issue I was raising.

Can I keep an eye on it and update this PR?

https://issuetracker.google.com/issues/338975163
https://issuetracker.google.com/issues/338975163#comment11

Songgyubin:
@Jaehwa-Noh 
[As a result of the inquiry](https://issuetracker.google.com/issues/338975163#comment12), it was said that it was the same problem that caused animation problems due to `prediction back cancellation`, and that the [fix](https://issuetracker.google.com/issues/346608857) was applied in beta04.

Therefore, the issue of screen overlapping when switching tabs quickly has been corrected starting with version beta04.

Jaehwa-Noh:
https://android-review.googlesource.com/c/platform/frameworks/support/+/3138255/5/navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavHost.kt#36
```
val transition = rememberTransition(transitionState, label = "entry")

        if (inPredictiveBack) {
            LaunchedEffect(progress) {
                val previousEntry = currentBackStack[currentBackStack.size - 2]
                transitionState.seekTo(progress, previousEntry)
            }
        } else {
            LaunchedEffect(backStackEntry) {
                // This ensures we don't animate after the back gesture is cancelled and we
                // are already on the current state
                if (transitionState.currentState != backStackEntry) {
                    transitionState.animateTo(backStackEntry)
                } else {
                    // convert from nanoseconds to milliseconds
                    val totalDuration = transition.totalDurationNanos / 1000000
                    // When the predictive back gesture is cancel, we need to manually animate
                    // the SeekableTransitionState from where it left off, to zero and then
                    // snapTo the final position.
                    animate(
                        transitionState.fraction,
                        0f,
                        animationSpec = tween((transitionState.fraction * totalDuration).toInt())
                    ) { value, _ ->
                        this@LaunchedEffect.launch {
                            if (value > 0) {
                                // Seek the original transition back to the currentState
                                transitionState.seekTo(value)
                            }
                            if (value == 0f) {
                                // Once we animate to the start, we need to snap to the right state.
                                transitionState.snapTo(backStackEntry)
                            }
                        }
                    }
                }
            }
        }
```

I've checked it, yes, there animation is fixed. This issue seems to be fixed that library version.
By the way, when I changed the navigation library to beta-06, I can't start the interests screen.

Could you possible to show the interests screen on this PR?

Songgyubin:
@Jaehwa-Noh 
I also can't open the interests screen.
This is an unexpected situation.

The app crashes.
```
java.lang.NoSuchMethodError: No static method localLookaheadPositionOf-RE3cj74$default(Landroidx/compose/ui/layout/LookaheadScope;Landroidx/compose/ui/layout/LayoutCoordinates;Landroidx/compose/ui/layout/LayoutCoordinates;JILjava/lang/Object;)J in class Landroidx/compose/ui/layout/LookaheadScope;
or its super classes (declaration of 'androidx.compose.ui.layout.LookaheadScope' appears in /data/app/~~A60uN9g5j3zSh8LW1c2u6A==/com.google.samples.apps.nowinandroid.demo.debug-FbvxGIMsxBPgNmQdZ-yKSA==/base.apk!classes12.dex)
```

Songgyubin:
@Jaehwa-Noh 
The problem was with androidx.compose.material3.adaptive

It has been resolved starting from `androidx.compose.material3.adaptive version beta03`, and I raised an issue and [received a confirmation](https://issuetracker.google.com/issues/360717840#comment2) that the problem was resolved.
(current version: `androidxComposeMaterial3Adaptive = "1.0.0-beta01"`)

The latest version is `1.1.0-alpha01`, but since there are many changes, we don't know what problems will arise, and there is not enough time to test, so we updated it to `1.0.0-rc01` to fix the bug in the PR.

**compose-material3-adaptive release note:** [https://developer.android.com/jetpack/androidx/releases/compose-material3-adaptive](https://developer.android.com/jetpack/androidx/releases/compose-material3-adaptive)

## After

<video src="https://github.com/user-attachments/assets/d9628837-cec9-4556-9498-19632e9c45bb" width="300" />

alexvanyo (approved):
The updates look great, thank you for investigating this and making the change!

PR #1583, opened 2024-09-02
change collectJob to backgroundScope

saeedishayan76:
Hello, 
‌Based on the latest document available in developer.android, we can use backgroundScope to collect flows instead of making a job object and canceling it after the final assertion.
With that, we do not need to manually cancel the created job, and the cancellation of this coroutine is done automatically before the end of the test. The following links are about the example on the site of developer.android and kotllinlang:

https://developer.android.com/kotlin/flow/test

Screenshot 2024-09-02 at 8 04 01 PM

https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-test-scope/background-scope.html


saeedishayan76:
I have signed the agreement.

saeedishayan76 (commented):
changes are stable and ready for merge

PR #1594, opened 2024-09-05
Fix duplicate composition trace markers

yongsuk44:
The composition trace markers are enabled by default, and we have corrected the issue of markers being displayed redundantly.

img  

image2


dturner:
@mlykotom Happy to merge this?

dturner:
Please fix formatting by running `./gradlew spotlessApply --init-script gradle/init.gradle.kts`

mlykotom (approved):
There's no problem leaving the custom traces in the code, because they might be used from the `TraceSectionMetric`, or just be a custom trace when you don't use the composition tracing. 

But given we don't have any use case for this, we can make the contributor happy :)

PR #1616, opened 2024-09-19
Replace BoxWithConstraints with Box

dturner:
Using `BoxWithConstraints` without constraints needlessly introduces a performance penalty. Replaced with `Box`.

Note: this is to fix our internal build which has stricter rules on lint checks.


bentrengrove (approved):

alexvanyo (approved):

Jaehwa-Noh:
I just wonder, why don't you remove `Box` on this situation? `ForYouScreen` needs additional `Box` compose on top of it?

PR #1621, opened 2024-09-20
make view model instance as private in MainActivity

momintahir:
The `viewModel` of `MainActivity` was only being used in this class, and the compiler also gave a warning that it could be made private.


dturner:
Please sign the CLA so we can accept this.

momintahir:
> Please sign the CLA so we can accept this.

Kindly check now.

bentrengrove (approved):

PR #1548, opened 2024-07-13
Update retrofit from 2.9.0 to 2.11.0

SimonMarquis:
- Replace Jake Wharton's Kotlinx Serialization converter with new retrofit's `converter-kotlinx-serialization`.
- Remove custom Proguard rules that are now part of the dependency: https://github.com/square/retrofit/tree/trunk/retrofit-response-type-keeper#readme

Changelog: https://github.com/square/retrofit/blob/trunk/CHANGELOG.md

This is an updated version of the initial update PR that remained untouched for too long:
- #1344

PR #1623, opened 2024-09-20
Update to Retrofit 2.11.0

dturner:
See https://github.com/android/nowinandroid/pull/1548. Thanks to @SimonMarquis for doing this.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -48,7 +48,7 @@ moduleGraph = "2.5.0"
 okhttp = "4.12.0"
 protobuf = "4.26.1"
 protobufPlugin = "0.9.4"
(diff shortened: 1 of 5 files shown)


alexvanyo (approved):

gradle/libs.versions.toml
```diff
@@ -48,7 +48,7 @@ moduleGraph = "2.5.0"
 okhttp = "4.12.0"
 protobuf = "4.26.1"
 protobufPlugin = "0.9.4"
-retrofit = "2.9.0"
+retrofit = "2.11.0"
 retrofitKotlinxSerializationJson = "1.0.0"
```
lihenggui, line 52: This line could be removed since it is not in use.

PR #1547, opened 2024-07-13
Update Gradle Wrapper from 8.7 to 8.10

SimonMarquis:
Release notes:
- https://docs.gradle.org/8.8/release-notes.html
- https://docs.gradle.org/8.9/release-notes.html
- https://docs.gradle.org/8.10/release-notes.html

PR #1009, opened 2023-10-31
Fix #1008: Use capitalized variant name when registering badging tasks

lihenggui:
**What I have done and why**
Use capitalized variant names when registering badging tasks.

Before: 
image

After
image

Fixes #1008

**Do tests pass?**
- [✓] Run local tests on `DemoDebug` variant: `./gradlew testDemoDebug`
- [✓] Check formatting: `./gradlew --init-script gradle/init.gradle.kts spotlessApply`

**Is this your first pull request?**
- [ ✓] [Sign the CLA](https://cla.developers.google.com/)
- [✓ ] Run `./tools/setup.sh`
- [✓ ] Import the code formatting style as explained in [the setup script](/tools/setup.sh#L40).

PR #1644, opened 2024-09-28
Use invariant locale when capitalizing Gradle task names

cketti:
### What I have done and why

When copying the badging code from this project, we noticed during code review that the `String.capitalized()` extension function uses the default locale. However, it is undesirable for the user's locale to be used when capitalizing (English) Gradle task names. It mostly works fine, but will e.g. capitalize `install` to `İnstall` ([dotted i](https://en.wikipedia.org/wiki/%C4%B0)) when `Locale.getDefault()` returns the Turkish locale.

This PR changes the code to use the invariant locale instead.

### History
Capitalization for badging task names was added in #1009. The code used an (internal) [extension function from Gradle](https://github.com/gradle/gradle/blob/d51dfd1a6ca7ee2d779de4988241733251e61f8d/platforms/core-configuration/stdlib-kotlin-extensions/src/main/kotlin/org/gradle/configurationcache/extensions/CharSequenceExtensions.kt) that also contained this bug.

When the Gradle function was deprecated, the functionality was included in this repository via #1547.

The deprecated function in Gradle has since been [fixed](https://github.com/gradle/gradle/blob/31a34d0b4368081e2a4d784d51fe4287d67dff89/platforms/core-configuration/stdlib-kotlin-extensions/src/main/kotlin/org/gradle/configurationcache/extensions/CharSequenceExtensions.kt) and is identical to the one in this PR.

### DIFF
--- a/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Badging.kt
+++ b/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Badging.kt
@@ -108,7 +108,7 @@ abstract class CheckBadgingTask : DefaultTask() {
 }
 
 private fun String.capitalized() = replaceFirstChar {
-    if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString()
+    if (it.isLowerCase()) it.titlecase() else it.toString()
 }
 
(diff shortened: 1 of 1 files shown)


dturner (approved):

Also thanks for the super clear PR description!

Issue #1647, opened 2024-09-30
[Bug]: DataStore 1.1.1 update broke unit tests


SimonMarquis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

The following update broke unit tests:
- https://github.com/android/nowinandroid/pull/1635

The changelog mentions:

> Performance improvement on DataStore only collects update notifications when it is observed. ([b/267792241](https://issuetracker.google.com/267792241))
> Note that this change might trigger UncompletedCoroutinesError in your tests if you are using the Coroutines testing library. Make sure you pass the [TestScope.backgroundScope](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-test-scope/background-scope.html) when initializing DataStore in your tests to avoid this issue.

### Relevant logcat output

```shell
After waiting for 1m, there were active child jobs: ["coroutine#37":StandaloneCoroutine{Active}@1bbe97e2]. Use `TestScope.backgroundScope` to launch the coroutines that need to be cancelled when the test body finishes
kotlinx.coroutines.test.UncompletedCoroutinesError: After waiting for 1m, there were active child jobs: ["coroutine#37":StandaloneCoroutine{Active}@1bbe97e2]. Use `TestScope.backgroundScope` to launch the coroutines that need to be cancelled when the test body finishes
	at app//kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTest$2$1$2.invokeSuspend$lambda$0(TestBuilders.kt:354)
	at app//kotlinx.coroutines.InvokeOnCancelling.invoke(JobSupport.kt:1571)
	at app//kotlinx.coroutines.JobSupport.notifyCancelling(JobSupport.kt:1604)
	at app//kotlinx.coroutines.JobSupport.tryMakeCancelling(JobSupport.kt:826)
	at app//kotlinx.coroutines.JobSupport.makeCancelling(JobSupport.kt:786)
	at app//kotlinx.coroutines.JobSupport.cancelImpl$kotlinx_coroutines_core(JobSupport.kt:702)
	at app//kotlinx.coroutines.JobSupport.cancelCoroutine(JobSupport.kt:689)
	at app//kotlinx.coroutines.TimeoutCoroutine.run(Timeout.kt:156)
[...]

PR #1659, opened 2024-09-30
Reverting update to DataStore 1.1.1 due to failing unit tests

dturner:
The update to DataStore 1.1.1 caused the build to fail: https://github.com/android/nowinandroid/actions/runs/10990714827/job/30511531410

Downgrading to 1.0.0 to investigate. Almost certainly due to the performance improvement introduced here: https://developer.android.com/jetpack/androidx/releases/datastore#1.1.0-beta02

Although the suggested fix (also applied in this PR) to use `testScope.backgroundScope` does not resolve the issue.


SimonMarquis:
- https://github.com/android/nowinandroid/pull/1648

SimonMarquis:
And the next issue is related to the badging process: https://github.com/android/nowinandroid/issues/1647#issuecomment-2382936131

dturner:
Thanks for the fix @SimonMarquis. This is no longer required.

Issue #1614, opened 2024-09-19
[Bug]: Can't navigate to 'For You' from 'Interests' after selecting 'News' topic tag


marzaise:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

A bug happened!
**Steps to reproduce:**

- After launching the app, click on any topic tag from a news item (e.g., Headlines or Wear OS).
- Once you've navigated to this interest item, try clicking on the first tab of the Navigation Bar (For You).

**Current behavior**
Clicking on the first tab of the Navigation Bar (For You) does nothing.

**Expected behavior**
Clicking on the first tab of the Navigation Bar (For You) should navigate to 'For You' (or perhaps perform a back action in this case).

**Video**
[foryou_bug_nia.webm](https://github.com/user-attachments/assets/3b3e2284-3753-468a-8987-bb932229886d)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

padrecedano:
Hi. Show your `NiaNavHost` and `TopLevelDestination` content.

sindrenm:
FWIW, I think this is slightly related to https://issuetracker.google.com/issues/308334893, where I have reported the same thing (see [#comment8](https://issuetracker.google.com/issues/308334893#comment8)).

sindrenm:
A related issue here is that the app switches the active navigation bar tab from For You to Interests when you tap a topic from inside of the For You tab. This, IMO, should never happen. The app should instead remain on the same For You tab, but with the interest content that the user requested to see.

The problem then lies in the way the app checks if a route exists in the hierarchy. Since the `TopicScreen`'s destination only exists in a graph in `InterestsListDetailScreen`, the hierarchy check will return `true` only for the Interests top level route:

```kotlin
private fun NavDestination?.isRouteInHierarchy(route: KClass<*>) =
    this?.hierarchy?.any {
        it.hasRoute(route)
    } ?: false
```

Issue #1597, opened 2024-09-05
[Bug]: Not able to return to "For You" screen from Topic screen


devwindsw:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

[Repro Steps]
1. Click a topic in "For You" screen and verify that the topic screen is launched
2. Click "Back" button in the top bar
3. Click "For You" tab in the bottom bar

[Expected Result]
We can navigate to "For You".

[Actual Result]
We're not able to navigate to "For You" until restart the app

[Frequency] 100%

[Screen_recording_20240906_023148.webm](https://github.com/user-attachments/assets/26c15304-12e4-477e-8721-ffff63942185)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anhtuannd:
Maybe it's duplication of https://github.com/android/nowinandroid/issues/1463 and https://github.com/android/nowinandroid/issues/1465

devwindsw:
> Maybe it's duplication of #1463 and #1465

The same issue happens even with the PR https://github.com/android/nowinandroid/pull/1470/commits/4c725bfe5dd8203479518e713bb495a8f28218ad

anhtuannd:
I rebased my commit to latest. You can check again, apparently I don't see this happen.

devwindsw:
> I rebased my commit to latest. You can check again, apparently I don't see this happen.

In "with_fix", we can find "LoadingWheel" is displayed and disappears shortly but not in "without_fix".

[without_fix.webm](https://github.com/user-attachments/assets/07000bfa-605a-4559-b3e2-693bf5436c38)
[with_fix.webm](https://github.com/user-attachments/assets/6bd0636b-37e1-443a-ab02-c7b6d015ddb8)

Issue #920, opened 2023-08-24
[Bug]: about #868，I found two bugs


lelelongwang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### about #868，I found two bugs:

- bug 1: When you execute onTopicClick in foryouScreen to enter the topicScreen, and then click foryouItem on the navigationbaritem, you cannot jump to foryouScreen.

> Solution 1: It can be consistent with "Interests", when jumping from foryouScreen to topicScreen, "Foryou" is still selected

> Solution 2: After jumping from foryouScreen to topicScreen, if "Interests" is selected, clicking "Foryou" should enter foryouScreen

- bug 2: When I execute onTopicClick again in topicScreen, I do not enter the new topicScreen according to the new topic ID, which means that in the topicScreen interface, the onTopicClick interface does not change

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dturner:
Fixed in #1681

PR #1681, opened 2024-11-01
Fix topic chip navigation from ForYou screen

dturner:
**What I have done and why**
I have converted the ForYou navigation graph from a single destination containing only `ForYouScreen`, to a navigation graph with 2 destinations: `ForYouScreen` and `TopicScreen`. Tapping on a topic chip inside the ForYou screen will now navigate to the Topic screen inside the ForYou navigation graph (previously tapping would do nothing). 

The ForYou feature remains independent of the Topic feature by providing an API for: 

- a navigation lambda when a topic chip is tapped 
- a destination for the topic content

```
fun NavGraphBuilder.forYouSection(
    topicContent: NavGraphBuilder.() -> Unit
){
    navigation<ForYouBaseRoute>(...) {
        composable<ForYouRoute> { ... }
        topicContent()
    }
}
```

Fixes: #1463 #1614 #1597 #920 

I have also: 

- Deleted `NetworkNewsResourceExpanded` and associated code as it isn't used anywhere
- Renamed some of the test method names as they weren't following the camel case convention


app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
```diff
@@ -267,4 +267,28 @@ class NavigationTest {
...
+            onNodeWithText(sampleTopic).performClick()
+
+            // Get the topic ID
+            val topic = runBlocking {
+                topicsRepository.getTopics().first().filter { it.name == sampleTopic }.first()
+            }
+
+            // Tap the first topic chip
+            onNodeWithTag("topicChip:${topic.id}", useUnmergedTree = true)
+                .assertExists()
```
dturner, line 284:
```
java.lang.AssertionError: Failed: assertExists.
Reason: Expected exactly '1' node but found '2' nodes that satisfy: (TestTag = 'topicChip:1')
Nodes found:
1) Node #209 at (l=315.0, t=2707.0, r=516.0, b=2833.0)px, Tag: 'topicChip:1'
Has 1 child
2) Node #239 at (l=1293.0, t=2896.0, r=1494.0, b=3022.0)px, Tag: 'topicChip:1'
Has 1 child, 2 siblings
```

dturner: @JoseAlcerreca Would love to chat about this when you have a minute. tl;dr adding UI tests is _incredibly_ painful, think I must be doing something wrong.

dturner:
Test is failing on CI. This is likely because the API 26 emulator window height isn't sufficiently large to allow the topic tag to be visible even after scrolling down to the news resource card. Example: 

image

Jaehwa-Noh:
What about using a `performScrollTo`?

dturner:
@Jaehwa-Noh Unfortunately that doesn't work because of https://issuetracker.google.com/234597156

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/model/NetworkNewsResource.kt
```diff
@@ -35,6 +35,20 @@ data class NetworkNewsResource(
     val topics: List<String> = listOf(),
 )
 
+fun NetworkNewsResource.asExternalModel(topics: List<NetworkTopic>) =
```
alexvanyo, line 38: This looks very similar to `NetworkNewsResourceExpanded`, where there is a list of `NetworkTopic`s. Is there a good way to combine those representations?

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
```diff
@@ -267,4 +273,34 @@ class NavigationTest {
...
+            val topic = newsResource.topics.first()
+            onNodeWithText(topic.name).performClick()
+
+            // Get the news feed and scroll to the news resource
+            // Note: Possible flakiness. If the content of the news resource is long then the topic
+            // tag might not be visible meaning it cannot be clicked
+            onNodeWithTag("forYou:feed")
+                .performScrollToNode(
+                    hasTestTag("newsResourceCard:${newsResource.id}"),
+                )
```
alexvanyo, line 295: Slightly terrible workaround for the scrolling issue: once we scroll such that the correct card is aligned with the top of the viewport, if we scroll an additional amount based on the height of the card so that the bottom of the card is within the viewport, the topic tags should be visible:

```suggestion
                .fetchSemanticsNode()
                .apply {
                    val newsResourceCardNode = onNodeWithTag("newsResourceCard:${newsResource.id}")
                        .fetchSemanticsNode()
                    config[ScrollBy].action?.invoke(
                        0f,
                        // to ensure the bottom of the card is visible,
                        // manually scroll the difference between the height of
                        // the scrolling node and the height of the card
                        (newsResourceCardNode.size.height - size.height).coerceAtLeast(0).toFloat()
                    )
                }
```

alexvanyo (approved):

app/src/androidTest/kotlin/com/google/samples/apps/nowinandroid/ui/NavigationTest.kt
dturner: It ain't pretty, but it works - thanks!

core/network/src/main/kotlin/com/google/samples/apps/nowinandroid/core/network/model/NetworkNewsResource.kt
dturner: `NetworkNewsResourceExpanded` isn't actually used anywhere except for tests so it can be safely deleted. 

I believe it was added because there's an endpoint on the backend server that will supply data in this format, we just don't use it.

alexvanyo (approved):

PR #1686, opened 2024-11-07
Fix - TopAppBar flickers when moving from destination to destination

mahmed1987:
**Issue**
currentTopLevelDestination is null momentarily , causing the TopAppBar to momentarily go out of the composition tree  while navigating the top level destinations. 
In short the problem lies here

```
val currentTopLevelDestination: TopLevelDestination?
        @Composable get() {
            return TopLevelDestination.entries.firstOrNull { topLevelDestination ->
                Log.d("currentDestination","$currentDestination")
                currentDestination?.hasRoute(route = topLevelDestination.route) ?: false
            }
        }
```

The log statement above would confirm that while switching between the top level composables , the currentDestination is obtained as null at times , until it is not . 

I dont have the exact reasoning as to why that is. My hypothesis is that since there is no "gate" between the `currentDestination` composable and the `currentTopLevelDestination` , `currentTopLevelDestination` reacts to everything that occurs to `currentDestination` and perhaps `currentDestination` goes through a null while going from one destination to another.

**Solution**

I just gated the `currentTopLevelDestination` with a remember . 

[Before](https://github.com/user-attachments/assets/7ec6fed2-efde-4372-8c7e-e58b8865b26d)
[After](https://github.com/user-attachments/assets/f2c69ee5-f7b2-4728-ba66-456b5d16d8bb)

### DIFF
--- a/app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
+++ b/app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaAppState.kt
@@ -89,8 +89,11 @@ class NiaAppState(
 
     val currentTopLevelDestination: TopLevelDestination?
         @Composable get() {
-            return TopLevelDestination.entries.firstOrNull { topLevelDestination ->
-                currentDestination?.hasRoute(route = topLevelDestination.route) ?: false
+            val currentDestination = currentDestination
+            return remember(currentDestination) {
+                TopLevelDestination.entries.firstOrNull { topLevelDestination ->
+                    currentDestination?.hasRoute(route = topLevelDestination.route) ?: false
+                }
             }
         }
 


Songgyubin:
@mahmed1987 
I have confirmed that the problem you mentioned has been resolved.

However, I think there is something that needs more thought about merging into the code.

1. What was the fundamental problem?
2. Is ‘remember’ really a fundamental solution rather than a temporary measure?
3. Are there any problems such as memory leaks by using `remember`?

mahmed1987:
@Songgyubin tbh I am not fully certain what is at play here. 

Like I wrote in the PR description , this appears to be some situation that is occurring perhaps because of the compose state `currentDestination` is being read inside of a `firstOrNull` lambda ?

trambui09:
Closing this PR as we have a fix for this: https://github.com/android/nowinandroid/pull/1728

PR #1704, opened 2024-11-19
Upgrade kotlin and ksp to latest stable versions

kaeawc:
## Instructions
Upgraded Kotlin & KSP so this project can receive a KSP memory leak fix https://github.com/google/ksp/issues/2073

I profiled NowInAndroid with a dev local build of KSP to originally test the fix which made it into release [2.0.21-1.0.26](https://github.com/google/ksp/releases/tag/2.0.21-1.0.26). I just profiled it again before and after the upgrade and can verify the metaspace leak is fixed by this change.


dturner (approved):

dturner:
Thanks for this. 

Just a note for the future, if you update dependencies please also run `./gradlew dependencyGuardBaseline` to update the dependency guard files. These make it easy for us to see what's changed in the release build.

dturner:
We've spent quite a bit of time updating library dependencies recently. Both Kotlin and KSP have been updated as part of updates elsewhere.

PR #1709, opened 2024-11-23
Use TextButton instead of Button with clickable

Stillswarm:
I have replaced Text with a clickable modifier to TextButton. This doesn't change the UI much, but provides a smoother and more consistent clicking experience.


dturner:
Many thanks for this. Looks great. Since this change affects UI, we also need updated screenshots. 

To generate the screenshots automatically, enable Github Actions on your fork.

If you are unable to do this, and your local development environment is in Linux, then manually generate the screenshots by running `./gradlew recordRoborazziDemoDebug`.

Stillswarm:
Hey Don. I have turned on Github Actions on my fork as instructed.

dturner:
@Stillswarm After doing a bit of digging into this, it appears that you cannot create screenshots directly on your `main` branch. Instead, you'll need to create a pull request on your own fork. I would: 

- Reset your `main` to the state it was in before the commits in this PR 
- Create a separate branch off main and add these commits
- Create a new PR from that branch onto our repo and allow the build workflow to run, this will create another commit with the updated screenshots in it. 

We should then be able to accept that PR and close this one.

PR #1719, opened 2024-12-09
Dependency version upgrades, not API 35 yet

keyboardsurfer:


gradle/libs.versions.toml
```diff
@@ -50,7 +50,7 @@ protobuf = "4.28.2"
 protobufPlugin = "0.9.4"
 retrofit = "2.11.0"
 retrofitKotlinxSerializationJson = "1.0.0"
-robolectric = "4.11.1"
+robolectric = "4.12.2"
```
MGaetan89, line 53: What about going to Robolectric 4.14.1?

keyboardsurfer: Good spot. For some reason that didn't show up in Android Studio.

riggaroo (approved):

riggaroo (dismissed):
Looks like the top app bar changed and i dont know if we want that?

riggaroo:
It seems to have removed the cog icons and search.

keyboardsurfer:
The top app bar changes were introduced with the upgrade to API 35 and are only visible when running roborazzi. For this maintenance effort it's out of scope for me to dive into roborazzi and address this.

Instead I reverted back to API 35 for target and compile Sdk version.

@takahirom FYI

riggaroo (approved):

keyboardsurfer:
Thanks for checking this @takahirom. It seems like https://github.com/takahirom/nowinandroid/pull/6 is the approach we should take for Now in Android.
@dturner WDYT?

dturner:
It'd be more conventional to set the theme in the manifest i.e. add `android:theme="@android:style/Theme.Material.Light.NoActionBar`
to the ui-test-hilt-manifest module's `AndroidManifest.xml`

keyboardsurfer:
While @takahirom's approach seems to work for light theme, there are still issues around theming with API 35 when using roborazzi. This change is out of scope for this PR. So we're sticking with 34 for now.

app/prodRelease-badging.txt
```diff
@@ -1,6 +1,6 @@
-package: name='com.google.samples.apps.nowinandroid' versionCode='8' versionName='0.1.2' platformBuildVersionName='14' platformBuildVersionCode='34' compileSdkVersion='34' compileSdkVersionCodename='14'
+package: name='com.google.samples.apps.nowinandroid' versionCode='8' versionName='0.1.2' platformBuildVersionName='15' platformBuildVersionCode='35' compileSdkVersion='35' compileSdkVersionCodename='15'
 sdkVersion:'21'
-targetSdkVersion:'34'
+targetSdkVersion:'35'
```
dturner, line 3: Seems like this PR _does_ update to API 35. Please could you update the PR description to match.

build-logic/convention/src/main/kotlin/AndroidApplicationFirebaseConventionPlugin.kt
```diff
@@ -35,7 +36,10 @@ class AndroidApplicationFirebaseConventionPlugin : Plugin<Project> {
                 val bom = libs.findLibrary("firebase-bom").get()
                 add("implementation", platform(bom))
                 "implementation"(libs.findLibrary("firebase.analytics").get())
-                "implementation"(libs.findLibrary("firebase.performance").get())
+                "implementation"(libs.findLibrary("firebase.performance").get()) {
+                    exclude(group = "com.google.protobuf", module = "protobuf-javalite")
```
dturner, line 40: Please add a comment as to why these exclusions are necessary (are they definitely necessary with all the version changes, maybe this isn't currently a problem?)

app/prodRelease-badging.txt
dturner: Looks like you need to manually update the prodReleaseBadging. You can do this by running: 

`./gradlew updateProdReleaseBadging`

dturner (approved):
Thanks for doing this. Good to merge once the build and instrumented tests pass.

takahirom:
It seems that `val composeTestRule = createComposeRule()` has the same issue in a simple library.  
I was able to reproduce it in an emulator test (instrumentation test) as well, so I'm thinking of filing this issue on Issue Tracker.
image

PR #1725, opened 2024-12-09
Add API 34 emulator

dturner:
Testing whether this is less flaky than the API30 emulator.


SimonMarquis:
What about API 35? 😃

keyboardsurfer (approved):

PR #1739, opened 2024-12-14
Followed Kotlin DSL without type-safe accessors documentation

Jaehwa-Noh:
**What I have done and why**
Followed Kotlin DSL without type-safe accessors documentation, and reduced likely occurs typo error by replacing string to a const variable.
https://docs.gradle.org/current/userguide/kotlin_dsl.html#ex-configuring-plugins-without-type-safe-accessors


dturner:
I'm more in favour of this format: 

```
"implementation"(libs.findLibrary("hilt.core").get())
```

Why? 

- It follows the exact format recommended by https://docs.gradle.org/current/userguide/kotlin_dsl.html#ex-configuring-plugins-without-type-safe-accessors
- It's more obvious to newcomers how to add dependencies for a new source set (e.g. `androidTestImplementation`) without needing to update `StringSources.kt`
- I agree the strings vals add type safety, but they are at the expense of another level of abstraction which I feel isn't worth the tradeoff

dturner (changes requested):
I'm more in favour of this format: 

```
"implementation"(libs.findLibrary("hilt.core").get())
```

Why? 

- It follows the exact format recommended by https://docs.gradle.org/current/userguide/kotlin_dsl.html#ex-configuring-plugins-without-type-safe-accessors
- It's more obvious to newcomers how to add dependencies for a new source set (e.g. `androidTestImplementation`) without needing to update `StringSources.kt`
- I agree the strings vals add type safety, but they are at the expense of another level of abstraction which I feel isn't worth the tradeoff

Jaehwa-Noh:
@dturner Okay, I will accept your suggestion. Removing the string file and replacing it with a plain String.

Jaehwa-Noh:
@dturner I've restored to the plain string text. Kindly check it.

dturner (approved):

PR #1752, opened 2024-12-16
Feature / Added baseline and startup profile generation

LinX64:
This PR adds the generated baseline + startup profile to the generated folder and completes the NightlyBaselineProfiles build script.

1. First, it sets up the GMD and creates a snapshot
2. Generates the baseline profile and startup using `:app:generateReleaseBaselineProfile`
3. Commit and push

Two steps were missing in the current script.

@keyboardsurfer @dturner - Would appreciate your thoughts on this.


keyboardsurfer:
Baseline Profiles do not need to be part of the repository's source code. We have chosen to only generate BP for release builds (and as nightly as continuous check that nothing is broken). Over time this saves a lot of data when checking out the repository.

LinX64:
> Baseline Profiles do not need to be part of the repository's source code. We have chosen to only generate BP for release builds (and as nightly as continuous check that nothing is broken). Over time this saves a lot of data when checking out the repository.

I understand the reason why you kept it like that, so thanks. However, I should say that the NightlyBaselineProfiles script does not properly check for the purpose you mentioned. 

> as nightly as continuous check that nothing is broken

For this purpose, I'd have:

```
- name: Generate Baseline Profile
        run: ./gradlew :app:generateReleaseBaselineProfile
          -Pandroid.testInstrumentationRunnerArguments.androidx.benchmark.enabledRules=baselineprofile
          -Pandroid.testoptions.manageddevices.emulator.gpu="swiftshader_indirect"
          --stacktrace
```

To verify if everything works as expected.

keyboardsurfer (approved):

PR #1761, opened 2024-12-18
Check presence of sarif files before executing `codeql-action/upload-sarif`

### LINKED ISSUE #1759: If the Build workflow fails, the SARIF upload also fails
If the Build workflow fails, no SARIF files are created and the "Upload lint reports (SARIF)" step also fails with: 

Error: No SARIF files found to upload in "./".

For example: https://github.com/android/nowinandroid/actions/runs/12379945378/job/34555220383?pr=1757. 

When opening the failed workflow, the default view is to show the last error. This makes things confusing because it's not the SARIF upload step that is the root cause of failure. 

@SimonMarquis any chance you could take a look and change the SARIF upload step so that it only attempts the upload if the SARIF files have been created?

SimonMarquis:
because the action behavior is to throw errors when no sarif files are available...

Closes #1759

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -151,7 +151,7 @@ jobs:
           path: '**/build/reports/lint-results-*.html'
 
       - name: Upload lint reports (SARIF)
-        if: always()
+        if: ${{ !cancelled() && hashFiles('**/*.sarif') != '' }}
         uses: github/codeql-action/upload-sarif@v3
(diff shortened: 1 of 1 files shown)


dturner (approved):

dturner:
Thanks for doing this. 

Weirdly this PR has an unresolved comment from @XiaomiTpro, which I suspect is a bot that has now been deleted.

PR #1773, opened 2024-12-20
Remove last bits of GMD requirements from regular build workflow

SimonMarquis:
The removal was initiated in:
- https://github.com/android/nowinandroid/pull/1757


dturner (approved):
Thank you for tidying this up.

PR #1777, opened 2024-12-22
Disable unnecessary workflows on forks

SimonMarquis:
These workflows are not meant to be run on forks anyways.

### DIFF
--- a/.github/workflows/NightlyBaselineProfiles.yaml
+++ b/.github/workflows/NightlyBaselineProfiles.yaml
@@ -7,6 +7,7 @@ on:
 jobs:
   baseline_profiles:
     name: "Generate Baseline Profiles"
+    if: github.repository == 'android/nowinandroid'
     runs-on: ubuntu-latest
 
     permissions:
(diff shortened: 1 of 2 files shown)


dturner:
I think it's OK for these workflows to be run on forks. The whole repo is intended to be cloned as a starting point for other projects so adding a check which stops them running might cause some confusion. If you fork the main repo, you can always delete the workflows if you don't want them. 

Happy to be convinced otherwise.

SimonMarquis:
From my point of view, these specific 2 workflows are kind of invisible for most contributors (including myself) in the sense that, if I were to fix a typo in the documentation, it would waste CI time building the nightly baselines for 30 days until it is disabled by GitHub for inactivity. Running CI for no reason is environmentally friendly :/
Almost no user will be able to directly use the release workflow either since they need a fairly large amount of change to provide proper keystore (change IDs, and more). And at this point they would have to understand GitHub Actions anyways to be able to plug everything together.

Jaehwa-Noh:
I agree with this PR. May many contributors are don't know how to disable nightly baseline profiler.
It was scheduled all forked repository.
But on release workflow, I don't have an idea to be needed this change.
I hadn't seen release workflow launch on my fork ever.

dturner:
Valid argument 👍 .

dturner (approved):

PR #1786, opened 2024-12-27
Add topicdao test

tobioyelekan:
**What I have done and why**
Add test for `TopicDao` for more coverage
Also extract common db setup to a separate class.


dturner (approved):

SimonMarquis:
@dturner would you be open to convert this pattern from inheritance to composition? (through a custom test Rule for instance.)

PR #1787, opened 2024-12-27
Apply ComposeScreenshot plugin

yuuki3655:
**What I have done and why**

Apply ComposeScreenshotTest plugin and move all compose preview functions to `screenshotTest` directory.


Jaehwa-Noh:
We had had roborazzi to take screenshots. If we would use compose screenshot plugin, remove roborazzi what does duplicated work.

JoseAlcerreca:
Thanks Yuki!

@Jaehwa-Noh I asked Yuki to create the PR.


I was thinking about using previews for testing static layouts and Roborazzi for cases where we need Robolectric features or to support user actions. 

Thoughts?

Jaehwa-Noh:
@JoseAlcerreca 
I hardly see what cases need an user action and Robolectric feature on Screenshot test, so that would you give me an example of it? That would be really me helpful!

PR #1798, opened 2025-01-07
Update OfflineFirstTopicsRepositoryTest.kt

eyhgpt:
Improve the unit test to perform meaningful comparisons instead of simply comparing two empty lists.

In the unit test function  `offlineFirstTopicsRepository_topics_stream_is_backed_by_topics_dao()`

Both `topicDao.getTopicEntities().first().map(TopicEntity::asExternalModel)` and `subject.getTopics().first()` return an empty list.  This is because  `subject.syncWith(synchronizer)` is not called beforehand. 

This unit test doesn't seem meaningful by just comparing two empty lists. 

This PR adds a  `subject.syncWith(synchronizer)` call before retrieving the topic entities, so that lists with actual elements can be compared, to prove that offlineFirstTopicsRepository.getTopics() is based by topicDao.getTopicEntities().

### DIFF
--- a/core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstTopicsRepositoryTest.kt
+++ b/core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstTopicsRepositoryTest.kt
@@ -67,7 +67,6 @@ class OfflineFirstTopicsRepositoryTest {
     @Test
     fun offlineFirstTopicsRepository_topics_stream_is_backed_by_topics_dao() =
         testScope.runTest {
-            // After sync, topicDao.getTopicEntities().first() and subject.getTopics().first()
             // will return non-empty lists. 
             subject.syncWith(synchronizer)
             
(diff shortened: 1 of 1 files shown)


core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstTopicsRepositoryTest.kt
```diff
@@ -67,6 +67,10 @@ class OfflineFirstTopicsRepositoryTest {
     @Test
     fun offlineFirstTopicsRepository_topics_stream_is_backed_by_topics_dao() =
         testScope.runTest {
+            // After sync, topicDao.getTopicEntities().first() and subject.getTopics().first()
```
dturner, line 70:
```suggestion
```

core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstTopicsRepositoryTest.kt
```diff
@@ -67,6 +67,10 @@ class OfflineFirstTopicsRepositoryTest {
...
+            // will return non-empty lists.
```
dturner, line 71:
```suggestion
```

dturner (approved):

PR #1799, opened 2025-01-08
Update OfflineFirstNewsRepositoryTest.kt

eyhgpt:
Improve the unit test to perform meaningful comparisons instead of simply comparing two empty lists.

In the unit test function offlineFirstNewsRepository_news_resources_stream_is_backed_by_news_resource_dao()

Both newsResourceDao.getNewsResources().first() and subject.getNewsResources().first() return an empty list. This is because subject.syncWith(synchronizer) is not called beforehand. It doesn't seem meaningful by just comparing two empty lists.

This PR adds a subject.syncWith(synchronizer) call before retrieving the lists, similar to PR https://github.com/android/nowinandroid/pull/1798

### DIFF
--- a/core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstNewsRepositoryTest.kt
+++ b/core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstNewsRepositoryTest.kt
@@ -89,8 +89,6 @@ class OfflineFirstNewsRepositoryTest {
     @Test
     fun offlineFirstNewsRepository_news_resources_stream_is_backed_by_news_resource_dao() =
         testScope.runTest {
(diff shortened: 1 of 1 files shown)


core/data/src/test/kotlin/com/google/samples/apps/nowinandroid/core/data/repository/OfflineFirstNewsRepositoryTest.kt
```diff
@@ -89,6 +89,8 @@ class OfflineFirstNewsRepositoryTest {
     @Test
     fun offlineFirstNewsRepository_news_resources_stream_is_backed_by_news_resource_dao() =
         testScope.runTest {
+            subject.syncWith(synchronizer)
+
```
dturner, line 93:
```suggestion
```

dturner (approved):

PR #1805, opened 2025-01-09
Save state of start destination when navigating to route in nested graph

### LINKED ISSUE #1779: [Bug] Unable to navigate to Saved after selecting topic in Saved screen

### What happened?

1. In For You tab, bookmark some news
2. In Saved tab, select any topic. This will navigate to Interests tab.
3. Try to select Saved tab again. Observe that now we are unable to navigate to Saved tab.

https://github.com/user-attachments/assets/a92c351a-505f-40e7-929a-90014c8b529c

### Relevant logcat output

_No response_

### Code of Conduct

anhtuannd:
**What I have done and why**
Save state of start destination (Bookmarks) when navigating to route in Interests nested graph, so when user selects Bookmarks again, correct destination will be restored.

Closes #1779


tobioyelekan:
You should add a screen record of the fix

anhtuannd:
@tobioyelekan This PR fixed #1779 and you can check original problem (included screen recording) in the ticket, so IMHO another video here is not so necessary

PR #1807, opened 2025-01-13
add ConfirmHapticFeedback in IconButton

kosenda:
**What I have done and why**

I thought it would be better to have haptics feedback to make it more obvious that the button was tapped, so I added haptics feedback to the commonly used IconButton

**ref**
[Android Developers - Compose UI - 1.8.0-alpha06](https://developer.android.com/jetpack/androidx/releases/compose-ui?_gl=1*qirkrx*_up*MQ..*_ga*MTEzNTM3NDg3Ni4xNzM2NzQxNTEw*_ga_6HH9YJMN9M*MTczNjc0MTUwOS4xLjAuMTczNjc0MTUwOS4wLjAuNzk1MTkyMDYx#1.8.0-alpha06)
> LocalHapticFeedback now provides a default HapticFeedback implementation when the Vibrator API indicates that haptics are supported. The following have been added to the HapticFeedbackType - Confirm, ContextClick, GestureEnd, GestureThresholdActivate, Reject, SegmentFrequentTick, SegmentTick, ToggleOn, ToggleOff, VirtualKey. Wear Compose long-clickable components such as Button, IconButton, TextButton, and Card now perform the LONG_PRESS haptic when a long-click handler has been supplied. ([I5083d](https://android-review.googlesource.com/#/q/I5083db2ce3619189c0a3793de9f535d996029c75))

[Android Developers - HapticFeedbackType](https://developer.android.com/reference/kotlin/androidx/compose/ui/hapticfeedback/HapticFeedbackType)
> `Confirm`
> A haptic effect to signal the confirmation or successful completion of a user interaction..

[Android Developers - Tap and press # long-press-show ](https://developer.android.com/develop/ui/compose/touch-input/pointer-input/tap-and-press#long-press-show)

**What I have tried**
I thought that `HapticFeedbackType` had `ToggleOn` and `ToggleOff`, which seemed to be better, but on my device Pixel 6, I didn't get any haptic feedback.
So I thought `Confirm` was appropriate.

```kotlin
onCheckedChange(isChecked)
    haptics.performHapticFeedback(
        if (isChecked) HapticFeedbackType.ToggleOn else HapticFeedbackType.ToggleOff
    )
},
```

### DIFF
--- a/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/IconButton.kt
+++ b/core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/IconButton.kt
@@ -24,6 +24,8 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.hapticfeedback.HapticFeedbackType
+import androidx.compose.ui.platform.LocalHapticFeedback
 import com.google.samples.apps.nowinandroid.core.designsystem.icon.NiaIcons
 import com.google.samples.apps.nowinandroid.core.designsystem.theme.NiaTheme
 
@@ -48,11 +50,15 @@ fun NiaIconToggleButton(
     icon: @Composable () -> Unit,
     checkedIcon: @Composable () -> Unit = icon,
 ) {
+    val haptics = LocalHapticFeedback.current
     // TODO: File bug
     // Can't use regular IconToggleButton as it doesn't include a shape (appears square)
     FilledIconToggleButton(
         checked = checked,
-        onCheckedChange = onCheckedChange,
+        onCheckedChange = { isChecked ->
+            onCheckedChange(isChecked)
+            if (isChecked) haptics.performHapticFeedback(HapticFeedbackType.Confirm)
+        },
         modifier = modifier,
         enabled = enabled,
         colors = IconButtonDefaults.iconToggleButtonColors(


core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/IconButton.kt
```diff
@@ -48,11 +50,15 @@ fun NiaIconToggleButton(
...
 ) {
+    val haptics = LocalHapticFeedback.current
     // TODO: File bug
     // Can't use regular IconToggleButton as it doesn't include a shape (appears square)
     FilledIconToggleButton(
         checked = checked,
-        onCheckedChange = onCheckedChange,
+        onCheckedChange = { isChecked ->
+            onCheckedChange(isChecked)
+            if (isChecked) haptics.performHapticFeedback(HapticFeedbackType.Confirm)
```
anhtuannd, line 60: I think haptic feedback should be called before line 59.

anhtuannd (commented):
Haptic feedback should be called first. Also using HapticFeedbackType.Confirm instead of  HapticFeedbackType.ToggleOn/ToggleOff is bit tricky.

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/IconButton.kt
kosenda: @anhtuannd 
Thank you for pointing this out!
If you don't mind, I'd like to know why this need to run the haptic feedback first?

Jaehwa-Noh:
It is as though you need to add a VIBRATE permission.
https://developer.android.com/reference/android/Manifest.permission#VIBRATE

core/designsystem/src/main/kotlin/com/google/samples/apps/nowinandroid/core/designsystem/component/IconButton.kt
Jaehwa-Noh: I also think it is proper to set the Haptic Toggle on and off.

kosenda:
It did work when I added permission !

blue928sky:
Probably no special permissions are needed for HapticFeedback.
`LongPress` worked on my device (Xperia 1 III Android 13) even without permissions. (`ToggleOff` and `ToggleOn` didn't seem to work though 😅 )

> Unlike other haptic APIs, using HapticFeedbackConstants with a View doesn't require the VIBRATE permission.

https://developer.android.com/develop/ui/views/haptics/haptic-feedback#use_view_prereq

PR #1818, opened 2025-01-26
[Clean-up] Add clean-up for SearchViewModel.

iamjosephmj:
**What I have done and why**

Remove `private val` modifier from `searchContentsRepository ` variable inside `SearchViewModel`


Jaehwa-Noh:
Why are you removing `private val`? 
Anything needs to be access `searchContentsRepository ?

PR #1823, opened 2025-02-02
NavHost replaced with if statement

Magirus:
Removed NavHost for ListDetailPaneScaffold

Creating NavHost object every time user goes to a new interest object is expensive. We may do it with more effective way. Especially when there are only two possible composable choices that can be switched with the only if statement.

### DIFF
--- a/feature/topic/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicScreen.kt
+++ b/feature/topic/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicScreen.kt
@@ -72,20 +72,24 @@ import com.google.samples.apps.nowinandroid.feature.topic.R.string
 
 @Composable
 fun TopicScreen(
+    topicId: String,
     showBackButton: Boolean,
     onBackClick: () -> Unit,
     onTopicClick: (String) -> Unit,
     modifier: Modifier = Modifier,
-    viewModel: TopicViewModel = hiltViewModel(),
+    viewModel: TopicViewModel = hiltViewModel()
 ) {
+    viewModel.updateTopic(topicId)
+
     val topicUiState: TopicUiState by viewModel.topicUiState.collectAsStateWithLifecycle()
     val newsUiState: NewsUiState by viewModel.newsUiState.collectAsStateWithLifecycle()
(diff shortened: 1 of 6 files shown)


feature/topic/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicViewModel.kt
```diff
@@ -29,52 +27,74 @@ import com.google.samples.apps.nowinandroid.core.model.data.Topic
...
+                    ),
+                    newsUiState(
+                        topicId = topicId,
+                        userDataRepository = userDataRepository,
+                        userNewsResourceRepository = userNewsResourceRepository,
+                    ),
+                ) { topicIUState, newsUIState ->
+                    _topicUIState.update { topicIUState }
+                    _newsUIState.update { newsUIState }
+                }.stateIn(viewModelScope)
```
alexvanyo, line 76: The `combine`s for old `topicId`s will continue to run right now, since these won't be cancelled if the topic id changes.

feature/topic/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/topic/TopicScreen.kt
```diff
@@ -72,20 +72,24 @@ import com.google.samples.apps.nowinandroid.feature.topic.R.string
...
 fun TopicScreen(
+    topicId: String,
     showBackButton: Boolean,
     onBackClick: () -> Unit,
     onTopicClick: (String) -> Unit,
     modifier: Modifier = Modifier,
-    viewModel: TopicViewModel = hiltViewModel(),
+    viewModel: TopicViewModel = hiltViewModel()
 ) {
+    viewModel.updateTopic(topicId)
```
alexvanyo, line 82: This line feels the strangest with this usage, since we are reusing a single instance of the `ViewModel` scoped to the overall navigation entry with many different `topicId`s that we are swapping between.

One side-effect is that we can't display two different versions of the `TopicScreen` at the same time, which might be ideal to animate between two different selected topics.

alexvanyo:
Thank you for proposing this for consideration - https://github.com/android/nowinandroid/pull/1835 removed the nested `NavHost` with a slightly different approach, so closing this PR

PR #1515, opened 2024-06-23
TestFixtures refactoring

SimonMarquis:
Support for Kotlin TestFixtures on Android modules was added in AGP 8.5.0!

A lot more cleanup could be done, especially on test classes that redefine over and over again topics, newsResources, etc.

This is a new take on this PR:
- https://github.com/android/nowinandroid/pull/452

To close this issue from 2022:
- https://github.com/android/nowinandroid/issues/451

---

The main changes are split by commits:

- AGP 8.5.0 update with `enableTestFixturesKotlinSupport` flag
- `:core:data-test`, `:sync:test` and `:core:datastore-test` have been merged as testFixtures sourceSet of their respective module.
- `:core:testing` content has been migrated when possible to already existing modules.
- `:sync:work` has been move to it's parent `:sync` module that was previously empty.

---

- [ ] Regenerate graphs

PR #1834, opened 2025-03-01
Enable test fixtures and migrate datastore test setup

emartynov:
**What I have done**

I've enabled Kotlin test fixtures and converted test-related modules to test fixtures:
1. Added flag in `gradle.properties`
2. Configured test fixtures in feature module
3. Moved code from the test module to the test fixtures folder in the feature module
4. Replaced all references from test module with test fixtures from feature module
5. Deleted the test module

**Why**

I believe test fixtures have two advantages from the test module:
1. It doesn't need to be a separate module, and the test fixture code lies close to the definition of dependency that needs to be set in tests.
6. Not for this change, but test fixtures have access to internal code in the module, making the public module API cleaner and smaller.

**Further changes**

* Put official documentation about the feature.
* Migrate a couple of other test-related modules to test fixtures.


SimonMarquis:
FYI some work was already been done in:
- https://github.com/android/nowinandroid/issues/451
- https://github.com/android/nowinandroid/pull/1515

But we are waiting on the maintainers feedback [here](https://github.com/android/nowinandroid/pull/1515#issuecomment-2535443453) and [there](https://github.com/android/nowinandroid/pull/1515#issuecomment-2553972217)

emartynov:
Ah, I didn't notice it, sorry! Let me see PRs. 

And I think I will close it.

emartynov:
Closing since there is full set in #1515

PR #1840, opened 2025-03-18
Remove Gradle Isolated Project violation

liutikas:
**What I have done and why**

Calls to Project.properties is not Gradle Isolated Projects safe as it causes Gradle to look through parent projects to find these. Move to using providers.gradleProperty as that does not have the same behavior.

Test: ./gradlew build --dry-run -Dorg.gradle.unsafe.isolated-projects=true
      fewer warnings after the PR


alexvanyo (approved):
Looks good!

PR #1843, opened 2025-03-18
Disable Gradle configuration cache on unsupported DependencyGuard task

SimonMarquis:
Fix Gradle Configuration Cache issue with Dependency Guard:

https://github.com/android/nowinandroid/actions/runs/13933394498/job/38995815537

```
Configuration cache state could not be cached: field `__monitoredConfigurationsMap__` of task `:app:dependencyGuard` of type `com.dropbox.gradle.plugins.dependencyguard.internal.list.DependencyGuardListTask`: error writing value of type 'org.gradle.api.internal.provider.DefaultMapProperty'
```

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -49,7 +49,7 @@ jobs:
       - name: Check Dependency Guard
         id: dependencyguard_verify
         continue-on-error: true
-        run: ./gradlew dependencyGuard
+        run: ./gradlew dependencyGuard --no-configuration-cache
 
       - name: Prevent updating Dependency Guard baselines if this is a fork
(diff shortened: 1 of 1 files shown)


SimonMarquis:
Another way to fix this could be to explicitlt disable configuration cache for the corresponding task type, but this would require a bit more changes. Let me know what you prefer.

SimonMarquis:
That being said, it would be great to still be able to use configuration cache for this task, because most of the time, it will remain UP-TO-DATE.
I can report the issue on the DependencyGuard plugin if necessary, but I don't think this plugin is still in active development. Maybe it would be the time to migrate to Gradle's builtin dependency locking mechanism 😏

PR #1855, opened 2025-03-30
Replace deprecated `centerAlignedTopAppBarColors` with `topAppBarColors`

SimonMarquis:
from a recent material3 update: https://developer.android.com/jetpack/androidx/releases/compose-material3#1.4.0-alpha06


alexvanyo (approved):
Thanks for the fix!

PR #1856, opened 2025-03-30
Cleanup Gradle dependencies and miscellaneous improvements

SimonMarquis:
- Inline versions that were used only once, allowing us to use the compact form for libraries & plugins
- Merge TOML declarations with `group` + `name` into the unique `module` component
- Expose new BoMs to enforce versioning across packages (`coil-kt-bom`, `kotlin-bom`, `kotlinx-coroutines-bom`, `kotlinx-serialization-bom`, `protobuf-bom`, `okhttp-bom`, `retrofit-bom`)
- Refactor `jacoco` version declaration into a proper dependency declaration to enable actual dependency tracking
- Rename `androidx.hilt.navigation.compose` to `hilt.ext.navigation.compose` to match the already existing `hiltExt` group.
- Replace duplicated `*.gradlePlugin` dependencies with converter that converts plugin alias to usable dependencies.
- Sort dependencies
- Remove unused `kotlinx-coroutines-android` dependency


gradle/libs.versions.toml
```diff
@@ -1,166 +1,118 @@
 [versions]
-accompanist = "0.37.0"
```
alexvanyo, line 2: I'm not a fan of inlining versions that only appear once - it does remove lines from this file, but I find it a lot clearer to have all of the versions in continuous place.

For future maintenance, this also seems like this would be more likely to have a dependency get copied in the same version group without moving that shared version back to the `versions` section.

alexvanyo (commented):
Thanks for the cleanup! I like all of these changes except for the inlining of versions out of the `[versions]` block if they're only used in one place - I feel like that ends being a bit messier even if slightly longer.

gradle/libs.versions.toml
Jaehwa-Noh: I think the inline version makes [version] category meaningless, and it would ruin the TOML convention.
If you want to remove the one time used version, you should be using this below.
This snippet from a gradle version catalog documentation.
https://docs.gradle.org/current/userguide/version_catalogs.html
image

SimonMarquis: This is clearly a downgrade. Also, Gradle recently published guidelines that go against this.
https://docs.gradle.org/current/userguide/best_practices_dependencies.html#single-gav-string

SimonMarquis: > the inline version makes [version] category meaningless, and it would ruin the TOML convention.

This is the whole point. Not using what is not necessary.
Which convention exactly? Extracting version numbers does not serve any purpose here.

dturner: I'm not wholly against this change. I think it does make it easier to read the dependencies, however, I also think it needs further investigation (does Android Studio break this format by automatically adding in new versions for single dependencies? Does this format work with our own version update tools/processes?)

@SimonMarquis Given that the rest of this PR is good to merge, would it be possible to split this change into a new PR we can discuss further there?

SimonMarquis: @dturner Ok, I'll restore the versions variables: https://github.com/android/nowinandroid/pull/1856/commits/967e7ac171e2774ddc28cb7af36ecdac99fbc0d3

> does Android Studio break this format by automatically adding in new versions for single dependencies?

What do you mean exactly? The Android Studio intent to add a missing dependency when it sees a known class? Android Studio has always been in a best-effort.
For instance, I tried to reference these 4 classes:

```kotlin
EmojiCompat
EmojiPickerView
EmojiButton
EmojiEditTextHelper
```

and here is what Android Studio did:

```toml
[versions]
emoji2 = "1.5.0"
emoji2Emojipicker = "1.5.0"
emoji2Views = "1.5.0"
emoji2ViewsHelper = "1.5.0"

[libraries]
androidx-emoji2 = { group = "androidx.emoji2", name = "emoji2", version.ref = "emoji2" }
androidx-emoji2-emojipicker = { group = "androidx.emoji2", name = "emoji2-emojipicker", version.ref = "emoji2Emojipicker" }
androidx-emoji2-views = { group = "androidx.emoji2", name = "emoji2-views", version.ref = "emoji2Views" }
androidx-emoji2-views-helper = { group = "androidx.emoji2", name = "emoji2-views-helper", version.ref = "emoji2ViewsHelper" }
```

So... I'll let you judge if we should follow what Android Studio does 😝

---

> Does this format work with our own version update tools/processes?

Yes is does, you can see it here: https://github.com/android/platform-samples/pull/219/commits/4c394aa637c5e0f13495cf9ef25935c4845420f0#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87df
It is also smart enough to apply the same version update to multiple entries if they were not using a variable.

SimonMarquis: I'll re-open a new PR with the inlined versions after this one is merged so that we can continue the discussion 👍

SimonMarquis: @dturner the requested changes have been applied. PR is ready to be reviewed (and merged?)

SimonMarquis: @dturner any last change required from your side or is it a good time to merge this?

Issue #1809, opened 2025-01-14
[Bug]: debug apk startup crash！！！！！


docwei2050:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### What happened?

      Caused by: android.os.strictmode.DiskReadViolation
                                                                                                    	at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1707)
                                                                                                    	at libcore.io.BlockGuardOs.access(BlockGuardOs.java:74)
                                                                                                    	at libcore.io.ForwardingOs.access(ForwardingOs.java:128)
                                                                                                    	at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:10127)
                                                                                                    	at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:332)
                                                                                                    	at java.io.File.exists(File.java:829)
                                                                                                    	at android.graphics.fonts.HwTypefaceEx.isUsingOnlineVariableFont(HwTypefaceEx.java:570)
                                                                                                    	at android.graphics.fonts.HwTypefaceEx.isSupportFontWeightVariable(HwTypefaceEx.java:486)
                                                                                                    	at android.graphics.fonts.SystemFonts.loadHwSystemFonts(SystemFonts.java:610)
                                                                                                    	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:8958)
                                                                                                    	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
                                                                                                    	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:3084)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:118)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:237) 
                                                                                                    	at android.os.Looper.loop(Looper.java:325) 
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:10305) 
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method) 
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:635) 
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:957) 

### Relevant logcat output

```shell
[...]

PR #1857, opened 2025-03-31
Remove death penalty from strict mode violations

keyboardsurfer:
The app was still crashing in non-debug after strict mode was enabled. Less restrictive, now only logging violations


alexvanyo (approved):

alexvanyo:
Thanks, this has been annoying me too when developing

SimonMarquis:
@keyboardsurfer can you add this in your PR body to automatically close these related issues?

Closes #1853
Closes #1809

PR #1860, opened 2025-04-05
Migrate from coil2 to coil3

SimonMarquis:
- Use BOM platform dependency.
- Remove -kt suffix from dependency name has they serve no real purpose.
- Add `coil-network` artifact to configure `CallFactory`.
- Although "First party Decoder" `SvgDecoder` is now automatically added to each new ImageLoader through a service loader, keeping it in code prevents us from removing the dependency that would appear unused otherwise.
- Replace explicit `respectCacheHeaders(false)` with `CacheStrategy::DEFAULT` which does not respect cache-control headers.
- Migrate `ImageLoaderFactory` to `SingletonImageLoader.Factory`
- Inline TOML dependencies (see https://github.com/android/nowinandroid/pull/1856)

docs: https://coil-kt.github.io/coil/upgrading_to_coil3/


core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt
```diff
@@ -118,6 +118,7 @@ fun NewsResourceCardExpanded(
         // Pass null for action to only override the label and not the actual action.
         modifier = modifier
             .semantics {
+                contentDescription = clickActionLabel
```
SimonMarquis, line 121: Let me know if you prefer to use `userNewsResource.title` or if another workaround might be better (adding a header semantic somewhere in the card maybe?)

SimonMarquis:
@dturner please take a look at this migration PR when you have time. It is a pretty straightforward migration.

PR #1861, opened 2025-04-05
Replace deprecated `ManagedDevices.devices` and suspicious receiver type

SimonMarquis:


benchmarks/build.gradle.kts
```diff
@@ -45,8 +45,8 @@ android {
         )
     }
 
-    testOptions.managedDevices.devices {
-        create<com.android.build.api.dsl.ManagedVirtualDevice>("pixel6Api33") {
+    testOptions.managedDevices.localDevices {
```
Jaehwa-Noh, line 48:
```suggestion
    testOptions.managedDevices.allDevices {
```
The documentation recommends using an `allDevices`.
https://developer.android.com/reference/tools/gradle-api/8.12/com/android/build/api/dsl/ManagedDevices#devices()
image

SimonMarquis: It does not really matter here (both versions will do the same), but if you read the definition of `localDevices`:

> Convenience container for specifying managed devices of type [ManagedVirtualDevice](https://developer.android.com/reference/tools/gradle-api/8.10/com/android/build/api/dsl/ManagedVirtualDevice).
> This list is managed in sync with [allDevices](https://developer.android.com/reference/tools/gradle-api/8.10/com/android/build/api/dsl/ManagedDevices#allDevices()). [ManagedVirtualDevice](https://developer.android.com/reference/tools/gradle-api/8.10/com/android/build/api/dsl/ManagedVirtualDevice) definitions added or removed in this container are correspondingly handled in [allDevices](https://developer.android.com/reference/tools/gradle-api/8.10/com/android/build/api/dsl/ManagedDevices#allDevices()), and vice versa.

and therefore I think this is more appropriate to use `localDevices` to create a local managed device, rather than `allDevices`.

Jaehwa-Noh: I'm thinking it is fine to use `localDevices` too, but my concern is they both, `localDevices` and `allDevices`, have a different type.

`devices`: `ExtensiblePolymorphicDomainObjectContainer<Device>`(deprecated)
`allDevices`: `ExtensiblePolymorphicDomainObjectContainer<Device>`
`localDevices`: `NamedDomainObjectContainer<ManagedVirtualDevice>`

SimonMarquis: What is your concern exactly? 
The type hierarchy is `interface ManagedVirtualDevice : Device`.

Jaehwa-Noh: Thank you for checking the hierarchy.
Good. My concern is resolved.

PR #1898, opened 2025-07-11
Simplify access to platform's `getSystemService` with reified Androidx's version

SimonMarquis:


SimonMarquis:
@dturner this tiny PR is ready for review

Issue #1842, opened 2025-03-18
[FR]: Make nowinandroid Gradle Isolated Projects compatible


liutikas:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the problem

Gradle [Isolated Projects](https://docs.gradle.org/current/userguide/isolated_projects.html) feature is an upcoming way to make project configuration faster. It would be great for nowinandroid to be compatible with this feature.

Running `./gradlew build --dry-run -Dorg.gradle.unsafe.isolated-projects=true` surfaces the following issues:
- [ ] Spotless Gradle Plugin
  - https://github.com/diffplug/spotless/issues/1979
- [ ]  `com.google.samples.apps.nowinandroid.GraphKt.configureGraphTasks(Graph.kt:130)` 
- [x] `dagger.hilt.android.plugin.HiltGradlePlugin.verifyDependencies(HiltGradlePlugin.kt:424)`
  - https://github.com/google/dagger/issues/4423
  - Fixed in 2.5.6
- [x] `androidx.room.gradle.RoomGradlePlugin$Companion.isGradleSyncRunning(RoomGradlePlugin.kt:347)` 
  - Fixed in room 2.7.0-rc02
  - No stable fixed version
- [x] `androidx.baselineprofile.gradle.utils.AgpPlugin$testMaxAgpVersion$2.invoke(AgpPlugin.kt:54)`
  - Fixed in 1.3.4 
- [x] `com.google.firebase.crashlytics.buildtools.gradle.CrashlyticsBuildtools.configure(CrashlyticsBuildtools.kt:51)` 
  - https://github.com/firebase/firebase-android-sdk/issues/6748
  - Fixed in 3.0.6
- [x] `com.google.firebase.perf.plugin.FirebasePerfPlugin.apply(FirebasePerfPlugin.java:111)`
  - https://github.com/firebase/firebase-android-sdk/issues/6748
  - Fixed in 2.0.1
- [x] REMOVED USAGE `com.google.android.libraries.mapsplatform.secrets_gradle_plugin.ExtensionsKt.loadPropertiesFile(Extensions.kt:44)`
  - https://github.com/google/secrets-gradle-plugin/issues/104
  - No fix
 - [x] `com.google.devtools.ksp.gradle.KspGradleSubplugin.applyToCompilation$configureAsAbstractKotlinCompileTool(KspSubplugin.kt:364)`
   - https://github.com/google/ksp/issues/1943
[...]

PR #1913, opened 2025-07-22
Move to a custom setup for BuildConfig.BACKEND_URL

liutikas:
secrets gradle plugin is not Gradle project isolation compatible and is using deprecated AGP variant API. This change removes the usage by using a custom provider to set this value.

Inspiration https://github.com/android/gradle-recipes/blob/agp-8.10/addCustomBuildConfigFields/build-logic/plugins/src/main/kotlin/CustomPlugin.kt

https://github.com/android/nowinandroid/issues/1842

Test: ./gradlew core:network:assemble -> BuildConfig.java has correct
      value

### DIFF
--- a/core/network/build.gradle.kts
+++ b/core/network/build.gradle.kts
@@ -14,12 +14,15 @@
  * limitations under the License.
  */
 
+import com.android.build.api.variant.BuildConfigField
+import java.io.StringReader
+import java.util.Properties
+
 plugins {
     alias(libs.plugins.nowinandroid.android.library)
     alias(libs.plugins.nowinandroid.android.library.jacoco)
     alias(libs.plugins.nowinandroid.hilt)
     id("kotlinx-serialization")
-    id("com.google.android.libraries.mapsplatform.secrets-gradle-plugin")
 }
 
 android {
@@ -32,10 +35,6 @@ android {
             isIncludeAndroidResources = true
         }
     }
-}
-
-secrets {
-    defaultPropertiesFileName = "secrets.defaults.properties"
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


SimonMarquis:
@liutikas CI fails because of this recent change on SARIF support https://github.blog/changelog/2025-07-21-code-scanning-will-stop-combining-multiple-sarif-runs-uploaded-in-the-same-sarif-file/

SimonMarquis:
Unfortunately I don't think lint currently supports the proper aggregation feature of Gradle, and the existing merging feature of Lint only merges reports from directly dependent modules.
I feel like the best solutions for now would be to either upload each file independently (but that could be error prone) or manually merge those runs together before uploading a single merged report.

Here is the implementation of what I think is the simplest for now:
- #1914

core/network/build.gradle.kts
```diff
@@ -52,3 +51,21 @@ dependencies {
...
     testImplementation(libs.kotlinx.coroutines.test)
 }
+
+val backendUrl = providers.fileContents(
+    isolated.rootProject.projectDirectory.file("local.properties")
+).asText.map { text ->
+    val properties = Properties()
+    properties.load(StringReader(text))
+    if (properties.containsKey("BACKEND_URL"))
+        (properties["BACKEND_URL"] as String)
```
dturner, line 61: nit: Can't you just return `properties["BACKEND_URL"]` as a `String?` here and have the `orElse` be called if it's null?

liutikas: Sadly this does not work in Gradle 8.x, but they have fixed it, left a comment to fix later.

PR #1922, opened 2025-08-01
Make scrollbar thumb size configurable

Akronix:
This change introduces a constant `THUMB_SIZE` to control the size of the scrollbar thumb, making it easier to adjust its appearance.

**What I have done and why:**

I just added a `const` variable to set the size of the scrollbar thumb, as I see it convenient to be able adapt it.


Jaehwa-Noh:
What about split width and height?
Currently, your modify always sets square.

Akronix:
@Jaehwa-Noh it is not square, the thumb "height" is determined from the amount of items in the list, whereas the width is what we are setting here. It just changes the other way around if the Orientation of the screen is horizontal instead of vertical.

Jaehwa-Noh:
@Akronix You're right. I am missing the when clause.

PR #1926, opened 2025-08-05
Remove unnecessary `@LooperMode` annotations

MGaetan89:
**What I have done and why**

I have removed all the `@LooperMode(LooperMode.Mode.PAUSED)` annotations, since `LooperMode.Mode.PAUSED` is the default (see https://github.com/robolectric/robolectric/blob/robolectric-4.15.1/robolectric/src/main/java/org/robolectric/plugins/LooperModeConfigurer.java#L17).


MGaetan89:
@dturner I've updated the PR for Robolectric 4.16.

MGaetan89:
@dturner the Robolectric update was done in #1981, but this still removes some unnecessary annotations.

PR #1939, opened 2025-08-09
Enable java cache.

Jaehwa-Noh:
**What I have done and why**
Enable Java cache on github action.

https://github.com/actions/setup-java?tab=readme-ov-file#caching-packages-dependencies

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -35,6 +35,7 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: 21
+          cache: 'gradle'
 
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -32,6 +32,7 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: 17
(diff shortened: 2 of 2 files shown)


SimonMarquis:
`gradle/actions/setup-gradle` already handles caching, which means we will duplicate the entire grape cache. Do you have any information about how they differ?

Jaehwa-Noh:
@SimonMarquis Good point!
I've searched a github action documentations and finally land here.
https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#incompatibility-with-other-caching-mechanisms

```
Incompatibility with other caching mechanisms
When using setup-gradle we recommend that you avoid using other mechanisms to save and restore the Gradle User Home.

Specifically:

Avoid using actions/cache configured to cache the Gradle User Home, [as described in this example](https://github.com/actions/cache/blob/main/examples.md#java---gradle).
Avoid using actions/setup-java with the cache: gradle option, [as described here](https://github.com/actions/setup-java#caching-gradle-dependencies).
Using either of these mechanisms may interfere with the caching provided by this action. If you choose to use a different mechanism to save and restore the Gradle User Home, you should disable the caching provided by this action, as described above.
```

It doesn't need a cache since setup-gradle cache set.

PR #1950, opened 2025-08-28
Upgrade to latest stable AGP 8.12.2

liutikas:
- Upgrades to a newer version of compose bom that pulls in newer
  versions of lint checks that work with AGP 8.12.2
- Bump the minSdk to 23 because compose now requires minSdk 23
- Update Navigation_fontScale2.png due to compose upgrade
- Clean up BuildConfig set up since we already use Gradle 9.0.0

Test: ./gradlew build


SimonMarquis:
FYI https://github.com/android/nowinandroid/pull/1861

liutikas:
> FYI #1861

Fair enough, removed that part of the change.

PR #1953, opened 2025-08-31
Mermaid Graphs

SimonMarquis:
This PR replaces the current dependency graphs generated by [jraska/modules-graph-assert](https://github.com/jraska/modules-graph-assert) and then converted to SVG when executing the [generateModuleGraphs.sh](https://github.com/android/nowinandroid/blob/main/generateModuleGraphs.sh) script (relying on `dot` & `svgo`) with inlined [Mermaid](https://mermaid.js.org/) graphs that are automatically rendered on GitHub.

See https://github.com/android/nowinandroid/discussions/1554 for more details.

---

There are still some questions left to answer before merging:
- [ ] `graphDump` and `graphUpdate` are currently being registered in a few different places, like "base" plugins. We could instead create a new plugin to automatically register them, or register them once from the root project with `allprojects` or keep it like this for now.
- [ ] There is no CI validation, meaning any change in the hierarchy of dependency might go unnoticed. Do we want an extra validation step? If so, we could also add a workflow step to automatically update the `README.md` files like we do for screenshots.


build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Graph.kt
```diff
@@ -0,0 +1,279 @@
...
+
+    override fun getDescription() = "Updates Markdown file with the corresponding dependency graph."
+
+    @TaskAction
+    operator fun invoke() = with(output.get().asFile) {
+        if (!exists()) {
+            createNewFile()
+            writeText(
+                """
+                # `${projectPath.get()}`
```
dturner, line 260:
```suggestion
                # `${projectPath.get()}` module dependency graph
```

dturner:
Firstly, thanks for this really excellent PR. The mermaid graphs are much clearer than the GraphViz SVGs.

> graphDump and graphUpdate are currently being registered in a few different places, like "base" plugins. We could instead create a new plugin to automatically register them, or register them once from the root project with allprojects or keep it like this for now.

Each module should have a dependency graph unless there's a special reason not to, for this reason `allprojects` makes sense to me. Put another way, when adding a new module it should have to explicitly opt-out of having a module graph generated for it. 

> There is no CI validation, meaning any change in the hierarchy of dependency might go unnoticed. Do we want an extra validation step? If so, we could also add a workflow step to automatically update the README.md files like we do for screenshots.

Sounds good to me. 

This change should be documented somewhere. The [Modularization learning journey]( https://github.com/android/nowinandroid/blob/main/docs/ModularizationLearningJourney.md) is the most obvious place to do this. Changes: 

- [x] Update the [modularization graph example](https://github.com/android/nowinandroid/blob/main/docs/ModularizationLearningJourney.md#types-of-modules-in-now-in-android) to use mermaid instead of the current GraphViz SVG
- [x] Document the graph format - what the colors/lines mean
- [x] Document the `graphDump` and `graphUpdate` commands
- [x] Document the CI workflow

SimonMarquis:
I've added a `RootPlugin` that is now responsible to register the graph related tasks on all sub modules.

---

For the CI validation step, I've just copy/pasted the dependency graph block and adapted it to the graphs, but I feel like this should probably need a refactor. For instance, having an early "validation" job that takes care of these kind of update, and only after that, we execute the build/test/androidTest jobs. But it is probably out of the scope of this PR.

I still need to test this (for example if new modules are added or deleted)

---

- [x] Update the [modularization graph example](https://github.com/android/nowinandroid/blob/main/docs/ModularizationLearningJourney.md#types-of-modules-in-now-in-android) to use mermaid instead of the current GraphViz SVG

Any suggestion? Should we try to map the current svg as close as possible?

- [x] Document the graph format - what the colors/lines mean

Do you think of documenting this in the same document?

---

- [x] Document the `graphDump` and `graphUpdate` commands

I'm not sure if it makes sense to document both of them. `graphDump` is an internal (intermediate) tasks that devs should probably never execute themselves. Only `graphUpdate` seems useful to document. Where would you want it documented exactly? IIn the Modularization learning journey docs (under which item)?

---

- [x] Document the CI workflow

I don't think workflows are currently documented anywhere. Any suggestion?

dturner:
> Any suggestion? Should we try to map the current svg as close as possible?

Yes, I would do that. The goal of that chart is to demonstrate the different types (`:app`, `:feature`, `:core`). It doesn't, and shouldn't show every module. 

 > Do you think of documenting this in the same document?
 
Yes, unless you can think of a more suitable place? If it was easy to add a legend to each module graph then that could be a better option. 

> Only graphUpdate seems useful to document. Where would you want it documented exactly? IIn the Modularization learning journey docs (under which item)?

Let's add a section "Module graphs" at the end and mention: 

- Each module has its own README containing a module graph, link to the app module as an example
- When module dependencies change, the module graphs are automatically updated, link to the CI workflow
- You can run the updates yourself by running the `graphUpdate` command

> I don't think workflows are currently documented anywhere. Any suggestion?

As above. 

Happy to do any/all of this if you want to delegate to me.

OleksandrKucherenko:
as an alternative: https://github.com/dorongold/gradle-task-tree/issues/41#issuecomment-3046425285

OleksandrKucherenko:
https://gist.github.com/OleksandrKucherenko/58478be9d17de80617ccabb6977325ea

OleksandrKucherenko:
https://github.com/OleksandrKucherenko/dependencies.graphviz

SimonMarquis:
Hi @OleksandrKucherenko thanks for your input but this has nothing to do with task graphs.
If you want to discuss module graph, there is already a discussion open here: https://github.com/android/nowinandroid/discussions/1554

OleksandrKucherenko:
> Hi @OleksandrKucherenko thanks for your input but this has nothing to do with task graphs.
> If you want to discuss module graph, there is already a discussion open here: https://github.com/android/nowinandroid/discussions/1554

Yep, I notice that too late.

Check another comment for more related to the dependencies graph.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Graph.kt
SimonMarquis: As this is a `h1` (header of level 1) this kind of prevent us to add further documentation about the module. That is the main reason why I kept it "generic".
Do you still want to proceed with this?
Or maybe we could do this instead?

```kotlin
writeText(
    """
    # `${projectPath.get()}`

    ## Module dependency graph

    <!--region graph--> <!--endregion-->

    """.trimIndent(),
)
```

SimonMarquis:
# Colors

Do you want to reuse the colors of the SVG or something like that (or something different)?

- `#536DFE` Android application
- `#3BD482` Android library
- `#607D8B` Android test
- `#7F52FF` JVM

---

# Types of modules

Here is my attempt at re-creating the SVG, without trying to go too far with invisble block to try mimic-ing the original layout.

As you can see, there are not differences between regular libraries and "features". Do we want to display them differently? In this case, I can adapt the plugin code to look for `nowinandroid.android.feature` plugin id instead.

```mermaid
graph TB
  subgraph :core
    direction TB
    :core:data[data]:::android-library
    :core:database[database]:::android-library
    :core:model[model]:::jvm
    :core:network[network]:::android-library
    :core:ui[ui]:::android-library
  end
  subgraph :feature
    direction TB
    :feature:topic[topic]:::android-library
    :feature:foryou[foryou]:::android-library
    :feature:interests[interests]:::android-library
    :feature:foo[...]:::android-library

  end
  :app[app]:::android-application

  :app -.-> :feature:foryou
  :app -.-> :feature:interests
  :app -.-> :feature:topic
  :core:data ---> :core:database
  :core:data ---> :core:network
  :core:database ---> :core:model
  :core:network ---> :core:model
  :core:ui ---> :core:model
  :feature:topic -.-> :core:data
  :feature:topic -.-> :core:ui

classDef android-application fill:#536DFE,stroke:#fff,stroke-width:2px,color:#fff;
classDef android-library fill:#3BD482,stroke:#fff,stroke-width:2px,color:#fff;
classDef jvm fill:#7F52FF,stroke:#fff,stroke-width:2px,color:#fff;
```
---

# Legends

### something very simple

```mermaid
graph TB
  application:::android-application -. implementation .-> feature:::android-library
  library:::android-library -- api --> jvm:::jvm

classDef android-application fill:#536DFE,stroke:#fff,stroke-width:2px,color:#fff;
classDef android-library fill:#3BD482,stroke:#fff,stroke-width:2px,color:#fff;
classDef jvm fill:#7F52FF,stroke:#fff,stroke-width:2px,color:#fff;
```

### or a more complex wiht real names?

```mermaid
graph TB
  :app[:app]:::android-application
  :feature:topic[:feature:topic]:::android-library
  :app -. implementation .-> :feature:topic

  :core:database[:core:database]:::android-library
  :core:model[:core:model]:::jvm
  :core:database -- api --> :core:model

classDef android-application fill:#536DFE,stroke:#fff,stroke-width:2px,color:#fff;
classDef android-library fill:#3BD482,stroke:#fff,stroke-width:2px,color:#fff;
classDef jvm fill:#7F52FF,stroke:#fff,stroke-width:2px,color:#fff;
```

---

# Docs

```markdown
## Module graphs

Each module has its own `README.md` file containing a module graph (e.g. [`:app` module graph](../app/README.md#module-dependency-graph)).  
When modules dependencies change, module graphs are automatically updated by the [Build.yaml](../.github/workflows/Build.yaml) workflow.  
You can also manually update the graphs by running the `graphUpdate` task.
```

dturner:
> Colors
There's not much contrast between some of the colors. We could use this pastel color scheme from some of the other Android sample repositories: 

```
val PastelRed = Color(0xFFFFADAD)
val PastelOrange = Color(0xFFFFD6A5)
val PastelYellow = Color(0xFFFDFFB6)
val PastelGreen = Color(0xFFCAFFBF)
val PastelBlue = Color(0xFF9BF6FF)
val PastelMauve = Color(0xFFA0C4FF)
val PastelPurple = Color(0xFFBDB2FF)
val PastelPink = Color(0xFFFFC6FF)
```

> Types of modules 

Looks good. I would remove the `:sync` block as it's not important for understanding the module types. 

Yes to displaying the feature modules differently, just choose any unallocated color from the above theme. 

> Legends

Simple looks good

> Docs

Looks great 👍

SimonMarquis:
I've updated the color mapping with this:

| Type        | Plugin ID                          | Color                  |
|-------------|------------------------------------|------------------------|
| Application | `nowinandroid.android.application` | `#CAFFBF` PastelGreen  |
| Feature     | `nowinandroid.android.feature`     | `#FFD6A5` PastelOrange |
| Library     | `nowinandroid.android.library`     | `#9BF6FF` PastelBlue   |
| Test        | `nowinandroid.android.test`        | `#A0C4FF` PastelMauve  |
| JVM         | `nowinandroid.jvm.library`         | `#BDB2FF` PastelPurple |
| Unknown     | `?`                                | `#FFADAD` PastelRed    |

`graphUpdate` is now documented in ModularizationLearningJourney.

dturner:
Once again, thanks for all your hard work here. I filed a follow up feature request #1962 so this PR could be merged without any further delay.

PR #1959, opened 2025-09-14
🚧 AGP 9

SimonMarquis:
This neat website can be used to track the current issues with AGP 9: https://agp-status.frybits.com/agp-9.0.0/

---

- [ ] https://github.com/firebase/firebase-android-sdk/issues/7293
- [ ] [AGP 9.0.0-alpha03 produces circular dependency on KSP tasks](https://issuetracker.google.com/issues/442128557)
  fixed in KSP [2.2.20-2.0.3](https://github.com/google/ksp/releases/tag/2.2.20-2.0.3)
- [ ] https://github.com/gradle/android-cache-fix-gradle-plugin/pull/1886
  Temporary workaround with `android.newDsl=false`
  ```
  > Failed to apply plugin 'org.gradle.android.cache-fix'.
     > Could not get unknown property 'unitTestVariants' for object of type com.android.build.gradle.internal.dsl.ApplicationExtensionImpl$AgpDecorated.
  ```
- [ ] https://github.com/google/dagger/issues/4944
  Temporary workaround with `android.newDsl=false`

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -37,7 +37,7 @@ kotlin {
 }
 
 dependencies {
-    compileOnly(libs.android.gradleApiPlugin)
+    compileOnly(libs.android.gradlePlugin)
     compileOnly(libs.android.tools.common)
     compileOnly(libs.compose.gradlePlugin)
     compileOnly(libs.firebase.crashlytics.gradlePlugin)
--- a/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
+++ b/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
@@ -28,12 +28,10 @@ import org.jetbrains.kotlin.compose.compiler.gradle.ComposeCompilerGradlePluginE
  * Configure Compose-specific options
  */
 internal fun Project.configureAndroidCompose(
-    commonExtension: CommonExtension<*, *, *, *, *, *>,
+    commonExtension: CommonExtension,
 ) {
     commonExtension.apply {
-        buildFeatures {
-            compose = true
-        }
+        buildFeatures.compose = true
 
         dependencies {
             val bom = libs.findLibrary("androidx-compose-bom").get()
--- a/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
+++ b/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
@@ -32,14 +32,12 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
  * Configure base Kotlin with Android options
  */
 internal fun Project.configureKotlinAndroid(
-    commonExtension: CommonExtension<*, *, *, *, *, *>,
+    commonExtension: CommonExtension,
 ) {
     commonExtension.apply {
         compileSdk = 35
 
-        defaultConfig {
-            minSdk = 23
-        }
+        defaultConfig.minSdk = 23
 
         compileOptions {
             // Up to Java 11 APIs are available through desugaring
--- a/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaFlavor.kt
+++ b/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaFlavor.kt
@@ -20,7 +20,7 @@ enum class NiaFlavor(val dimension: FlavorDimension, val applicationIdSuffix: St
 }
 
 fun configureFlavors(
-    commonExtension: CommonExtension<*, *, *, *, *, *>,
+    commonExtension: CommonExtension,
     flavorConfigurationBlock: ProductFlavor.(flavor: NiaFlavor) -> Unit = {},
 ) {
     commonExtension.apply {
@@ -28,15 +28,13 @@ fun configureFlavors(
             flavorDimensions += flavorDimension.name
         }
 
-        productFlavors {
-            NiaFlavor.values().forEach { niaFlavor ->
-                register(niaFlavor.name) {
-                    dimension = niaFlavor.dimension.name
-                    flavorConfigurationBlock(this, niaFlavor)
-                    if (this@apply is ApplicationExtension && this is ApplicationProductFlavor) {
-                        if (niaFlavor.applicationIdSuffix != null) {
-                            applicationIdSuffix = niaFlavor.applicationIdSuffix
-                        }
+        NiaFlavor.values().forEach { niaFlavor ->
+            productFlavors.register(niaFlavor.name) {
+                dimension = niaFlavor.dimension.name
+                flavorConfigurationBlock(this, niaFlavor)
+                if (this@apply is ApplicationExtension && this is ApplicationProductFlavor) {
+                    if (niaFlavor.applicationIdSuffix != null) {
+                        applicationIdSuffix = niaFlavor.applicationIdSuffix
                     }
                 }
             }
--- a/core/datastore/src/main/kotlin/com/google/samples/apps/nowinandroid/core/datastore/di/DataStoreModule.kt
+++ b/core/datastore/src/main/kotlin/com/google/samples/apps/nowinandroid/core/datastore/di/DataStoreModule.kt
@@ -37,11 +37,11 @@ import javax.inject.Singleton
 
 @Module
 @InstallIn(SingletonComponent::class)
-object DataStoreModule {
+internal object DataStoreModule {
 
     @Provides
     @Singleton
-    internal fun providesUserPreferencesDataStore(
+    fun providesUserPreferencesDataStore(
         @ApplicationContext context: Context,
         @Dispatcher(IO) ioDispatcher: CoroutineDispatcher,
         @ApplicationScope scope: CoroutineScope,
--- a/gradle.properties
+++ b/gradle.properties
@@ -63,3 +63,5 @@ roborazzi.test.verify=true
 # Prevent uninstall app after instrumented tests
 # https://issuetracker.google.com/issues/295039976
 android.injected.androidTest.leaveApksInstalledAfterRun=true
+
+android.newDsl=false
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,8 +2,8 @@
 accompanist = "0.37.0"
 androidDesugarJdkLibs = "2.1.4"
 # AGP and tools should be updated together
-androidGradlePlugin = "8.12.2"
-androidTools = "31.12.2"
+androidGradlePlugin = "9.0.0-alpha05"
+androidTools = "32.0.0-alpha05"
 androidxActivity = "1.9.3"
 androidxAppCompat = "1.7.0"
 androidxBrowser = "1.8.0"
@@ -151,7 +151,7 @@ truth = { group = "com.google.truth", name = "truth", version.ref = "truth" }
 turbine = { group = "app.cash.turbine", name = "turbine", version.ref = "turbine" }
 
 # Dependencies of the included build-logic
-android-gradleApiPlugin = { group = "com.android.tools.build", name = "gradle-api", version.ref = "androidGradlePlugin" }
+android-gradlePlugin = { group = "com.android.tools.build", name = "gradle-api", version.ref = "androidGradlePlugin" }
 android-tools-common = { group = "com.android.tools", name = "common", version.ref = "androidTools" }
 compose-gradlePlugin = { module = "org.jetbrains.kotlin:compose-compiler-gradle-plugin", version.ref = "kotlin" }
 firebase-crashlytics-gradlePlugin = { group = "com.google.firebase", name = "firebase-crashlytics-gradle", version.ref = "firebaseCrashlyticsPlugin" }
--- a/app-nia-catalog/dependencies/releaseRuntimeClasspath.txt
+++ b/app-nia-catalog/dependencies/releaseRuntimeClasspath.txt
@@ -124,10 +124,10 @@ io.coil-kt:coil-compose:2.7.0
 io.coil-kt:coil:2.7.0
 jakarta.inject:jakarta.inject-api:2.0.1
 javax.inject:javax.inject:1
-org.jetbrains.kotlin:kotlin-stdlib-common:2.1.10
+org.jetbrains.kotlin:kotlin-stdlib-common:2.2.10
 org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.9.0
 org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.0
-org.jetbrains.kotlin:kotlin-stdlib:2.1.10
+org.jetbrains.kotlin:kotlin-stdlib:2.2.10
 org.jetbrains.kotlinx:kotlinx-coroutines-android:1.8.1
 org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.8.1
 org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.8.1
--- a/app/dependencies/prodReleaseRuntimeClasspath.txt
+++ b/app/dependencies/prodReleaseRuntimeClasspath.txt
@@ -226,10 +226,10 @@ javax.inject:javax.inject:1
 org.checkerframework:checker-qual:3.12.0
 org.jetbrains.kotlin:kotlin-android-extensions-runtime:1.9.22
 org.jetbrains.kotlin:kotlin-parcelize-runtime:1.9.22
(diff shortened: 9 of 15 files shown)


build-logic/convention/build.gradle.kts
```diff
@@ -37,7 +37,7 @@ kotlin {
 }
 
 dependencies {
-    compileOnly(libs.android.gradleApiPlugin)
+    compileOnly(libs.android.gradlePlugin)
```
liutikas, line 40: Can you share why we need to make this change?

SimonMarquis: This is obviously not a required change, but I decided to remove (hide) the fact that is was an "api" version of the dependency since the "impl" is not supposed to be used in future releases. Keeping the distinction between the api and the impl visible in the dependency name is not that useful.

liutikas: Ah! Missed the version catalog part of the change!

gradle/libs.versions.toml
```diff
@@ -46,7 +46,7 @@ kotlin = "2.2.20"
 kotlinxCoroutines = "1.10.1"
 kotlinxDatetime = "0.6.1"
 kotlinxSerializationJson = "1.8.0"
-ksp = "2.2.20-2.0.4"
+ksp = "2.3.1-SNAPSHOT"
```
liutikas: You should be able to undo this part, I upgrades us to 2.3.1

gradle/libs.versions.toml
```diff
@@ -34,7 +34,7 @@ coil = "2.7.0"
 dependencyGuard = "0.5.0"
 firebaseBom = "33.7.0"
 firebaseCrashlyticsPlugin = "3.0.6"
-firebasePerfPlugin = "2.0.1"
+firebasePerfPlugin = "2.0.2"
```
liutikas, line 41: Will handle this upgrade in https://github.com/android/nowinandroid/pull/1983

liutikas: Merged.

liutikas:
> release build type has isDebuggable forced to true, even when explicitly turned off.

Can you point me to how to repro this?

SimonMarquis:
I've not been able to reproduce it on my other projects, but on this one, simply build a `release` app with `gradlew :app:assembleDemoRelease`.

SimonMarquis:
I found the issue, fixed in 52b118a1e073c07c4b7843e3ca93ea70333e41a0.
`enableAndroidTestCoverage` forces the app to be debuggable, probably because of a new logic internal to AGP.

I've refactored the code to configure only the `debug` buildTypes:
https://github.com/SimonMarquis/nowinandroid/blob/52b118a1e073c07c4b7843e3ca93ea70333e41a0/build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt#L68-L71

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
```diff
@@ -28,12 +28,10 @@ import org.jetbrains.kotlin.compose.compiler.gradle.ComposeCompilerGradlePluginE
...
  */
 internal fun Project.configureAndroidCompose(
-    commonExtension: CommonExtension<*, *, *, *, *, *>,
+    commonExtension: CommonExtension,
 ) {
     commonExtension.apply {
-        buildFeatures {
-            compose = true
-        }
+        buildFeatures.compose = true
```
liutikas: Please undo the flattening of these calls. It is unrelated to this change. Please do the same for all the places.

I prefer we land this change with minimal diff. This kind of clean up should be a separate PR.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt
```diff
@@ -59,8 +62,14 @@ private fun String.capitalize() = replaceFirstChar {
  * tests on CI using a different Github Action or an external device farm.
  */
 internal fun Project.configureJacoco(
+    commonExtension: CommonExtension,
     androidComponentsExtension: AndroidComponentsExtension<*, *, *>,
 ) {
+    commonExtension.buildTypes.named("debug") {
```
liutikas, line 69: Add a comment on why this is needed.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
```diff
@@ -32,22 +32,18 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
...
-    commonExtension: CommonExtension<*, *, *, *, *, *>,
+    commonExtension: CommonExtension,
 ) {
     commonExtension.apply {
         compileSdk = 35
 
-        defaultConfig {
-            minSdk = 23
-        }
+        defaultConfig.minSdk = 23
```
liutikas: Please undo the flattening of these calls.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaFlavor.kt
```diff
@@ -20,23 +20,21 @@ enum class NiaFlavor(val dimension: FlavorDimension, val applicationIdSuffix: St
...
-        productFlavors {
-            NiaFlavor.values().forEach { niaFlavor ->
-                register(niaFlavor.name) {
-                    dimension = niaFlavor.dimension.name
-                    flavorConfigurationBlock(this, niaFlavor)
-                    if (this@apply is ApplicationExtension && this is ApplicationProductFlavor) {
-                        if (niaFlavor.applicationIdSuffix != null) {
-                            applicationIdSuffix = niaFlavor.applicationIdSuffix
-                        }
+        NiaFlavor.values().forEach { niaFlavor ->
```
liutikas: Undo this rewrite, it is unrelated to the upgrade of AGP

core/datastore/src/main/kotlin/com/google/samples/apps/nowinandroid/core/datastore/di/DataStoreModule.kt
```diff
@@ -33,15 +33,17 @@ import dagger.hilt.android.qualifiers.ApplicationContext
...
 
+@VisibleForTesting
 @Module
 @InstallIn(SingletonComponent::class)
 object DataStoreModule {
 
     @Provides
     @Singleton
-    internal fun providesUserPreferencesDataStore(
+    fun providesUserPreferencesDataStore(
```
liutikas: Why is this change needed? Is it related to the AGP upgrade?

gradle.properties
```diff
@@ -63,3 +63,5 @@ roborazzi.test.verify=true
 # Prevent uninstall app after instrumented tests
 # https://issuetracker.google.com/issues/295039976
 android.injected.androidTest.leaveApksInstalledAfterRun=true
+
+android.newDsl=false
```
liutikas, line 70: Add a comment on why we need to opt out linking to the hilt bug.

sync/work/src/demo/kotlin/com/google/samples/apps/nowinandroid/sync/di/SyncModule.kt
```diff
@@ -24,7 +24,9 @@ import dagger.Binds
 import dagger.Module
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import org.jetbrains.annotations.VisibleForTesting
 
+@VisibleForTesting
```
liutikas: This seems to be unrelated to AGP upgrade. Please undo

sync/work/src/prod/kotlin/com/google/samples/apps/nowinandroid/sync/di/SyncModule.kt
```diff
@@ -28,8 +28,10 @@ import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import org.jetbrains.annotations.VisibleForTesting
 import javax.inject.Singleton
 
+@VisibleForTesting
```
liutikas: This seems to be unrelated to AGP upgrade. Please undo

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
SimonMarquis: I'd love to, but AGP broke these nice APIs with lambdas: https://github.com/android/nowinandroid/pull/1959/commits/b060fd0b71cf7cca93c761b4ecb0ee0935998dc2

core/datastore/src/main/kotlin/com/google/samples/apps/nowinandroid/core/datastore/di/DataStoreModule.kt
SimonMarquis: Yes, see this commit: https://github.com/android/nowinandroid/pull/1959/commits/fd939d5a48898f263c9e737fcd8428395932d8bd

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaFlavor.kt
SimonMarquis: Same AGP API breakage.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/KotlinAndroid.kt
SimonMarquis: Same AGP API breakage.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
bddckr: The [upgrade docs](https://developer.android.com/build/releases/agp-preview#android-gradle-plugin-changed-dsl) suggest to use Kotlin's `.apply { ... }`.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/Jacoco.kt
SimonMarquis: https://github.com/android/nowinandroid/pull/1959/commits/1ebc99d32c78318b9210e341687f5dfe7425a799

gradle.properties
SimonMarquis: This is not meant to be merged as-is, as mentioned in the PR first message.
That being said, if you really want to merge this PR before alls bugs are fixed & workarounds are removed, I can add a comment for this. Let me know.

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/AndroidCompose.kt
SimonMarquis: Changing receivers is sometimes a very sneaky source of bugs...
So, unless @liutikas really want this `.apply{}` version everywhere, I'll keep it as-is.

bddckr: The receiver wouldn't be changing compared to `main`/the pre-9-AGP-APIs. It was the same previously.

SimonMarquis: Yes I know. I meant it as a general rule. Switching context receivers va a proposer dsl that is enforced by the IDE is always better 😀 but that is not an option for now.

SimonMarquis: I'll wait for the final call and update if necessary

SimonMarquis: Since the reviewer asked for minimal changes, I reverted the changes with `.apply`.

core/datastore/src/main/kotlin/com/google/samples/apps/nowinandroid/core/datastore/di/DataStoreModule.kt
liutikas: This looks like a bug, please file one. Potentially a hilt issue?

gradle.properties
liutikas: Let's add a link to the issue.

core/datastore/src/main/kotlin/com/google/samples/apps/nowinandroid/core/datastore/di/DataStoreModule.kt
SimonMarquis: Probably due to mangling. Unfortunate I'm be afk for a few weeks and I won't be able to create a proper bug report until I'm back.

liutikas:
Can you please rebase the merge conflicts?

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaFlavor.kt
```diff
@@ -20,20 +20,20 @@ enum class NiaFlavor(val dimension: FlavorDimension, val applicationIdSuffix: St
...
         }
 
-        productFlavors {
+        productFlavors.apply flavors@{
             NiaFlavor.values().forEach { niaFlavor ->
                 register(niaFlavor.name) {
                     dimension = niaFlavor.dimension.name
                     flavorConfigurationBlock(this, niaFlavor)
-                    if (this@apply is ApplicationExtension && this is ApplicationProductFlavor) {
+                    if (this@flavors is ApplicationExtension && this is ApplicationProductFlavor) {
```
imknown: 1. `productFlavors { } ` still works using `AGP 9.0.0-rc02`. No `apply` needed.

2. `this@flavors` may be incorrect, should it be:
``` koltin
commonExtension.apply extension@{
    // ...
    productFlavors flavors@{
        // ...
        if (this@extension is ApplicationExtension ...) {
        // ...
    }
}
```

core/datastore/src/main/kotlin/com/google/samples/apps/nowinandroid/core/datastore/di/DataStoreModule.kt
SimonMarquis: This issue seems to be fixed in the current KSP/Hilt version, reverting the commit.

gradle.properties
SimonMarquis: Added in [`8cbbd45` (#1959)](https://github.com/android/nowinandroid/pull/1959/commits/8cbbd454ebb9901f05e5f37c44564d82d67ede0c)

build-logic/convention/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaFlavor.kt
SimonMarquis: The extension was removed from AGP but kotlin-dsl provides one that fits: [`7839613` (#1959)](https://github.com/android/nowinandroid/pull/1959/commits/7839613a5b2fab6c16a0f1beb9ee2cb693d25816)

SimonMarquis:
<strike>

AGP 9.0.0 does not seem to have added the corresponding `android {}` extensions to configure android modules. Which means we get these warnings:~~


```
'fun Project.android(configure: Action<BaseAppModuleExtension>): Unit' is deprecated. Replaced by com.android.build.api.dsl.ApplicationExtension.
This class is not used for the public extensions in AGP when android.newDsl=true, which is the default in AGP 9.0, and will be removed in AGP 10.0.
```

```
'fun Project.android(configure: Action<TestExtension>): Unit' is deprecated. Replaced by com.android.build.api.dsl.TestExtension.
This class is not used for the public extensions in AGP  when android.newDsl=true, which is the default in AGP 9.0, and will be removed in AGP 10.0.
```

```
 'fun Project.android(configure: Action<LibraryExtension>): Unit' is deprecated. Replaced by com.android.build.api.dsl.LibraryExtension.
This class is not used for the public extensions in AGP when android.newDsl=true, which is the default in AGP 9.0, and will be removed in AGP 10.0.
```

~~We could add them ourselves in the build-logic and expose them as public API until they offer a solution, or ignore the warning for now, wdyt?~~

```kotlin
fun Project.androidApp(configure: ApplicationExtension.() -> Unit) = configure<ApplicationExtension>(configure)
fun Project.androidLibrary(configure: LibraryExtension.() -> Unit) = configure<LibraryExtension>(configure)
fun Project.androidTest(configure: TestExtension.() -> Unit) = configure<TestExtension>(configure)
```

</strike>

**EDIT:** It seems like this only happens when `android.newDsl=false`, so I guess we should not try to work around it and wait for the proper solution of enabling the new DSL.

PR #1963, opened 2025-09-20
🧜‍♀️ Mermaid Graphs legends

### LINKED ISSUE #1962: [FR]: Add dependency graph legend to each README
Add the dependency graph legend to each README. It's currently shown only on https://github.com/android/nowinandroid/blob/main/docs/ModularizationLearningJourney.md#types-of-modules-in-now-in-android, however, if you're not familiar with that page then you won't know what the colors/lines mean. Since the module READMEs are autogenerated, it should be easy to update the graph output to include the legend as well.

@SimonMarquis FYI you're probably best to take this if you don't mind.

SimonMarquis:
Closes #1962

I've used an additionnal task output that will generate the `legend.txt` which will then be used when reconstructing the markdown content.

Side note:  
Android Studio Otter (2025.2.1) is finally getting a Markdown renderer 🎉  
Unfortunately, the rendering engine is forced to `Compose (experimental)` which means, it only parses basic stuff, and `<details>` are not rendered, as well as mermaid content, even with the proper plugin installed.  
Hopefully they will support them soon.

### DIFF
--- a/app-nia-catalog/README.md
+++ b/app-nia-catalog/README.md
@@ -22,9 +22,9 @@ graph TB
 
   :app-nia-catalog -.-> :core:designsystem
   :app-nia-catalog -.-> :core:ui
-  :core:ui ---> :core:analytics
-  :core:ui ---> :core:designsystem
-  :core:ui ---> :core:model
+  :core:ui --> :core:analytics
+  :core:ui --> :core:designsystem
+  :core:ui --> :core:model
 
 classDef android-application fill:#CAFFBF,stroke:#000,stroke-width:2px,color:#000;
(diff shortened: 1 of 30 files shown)


Qingcaj:
> Closes #1962
> 
> I've used an additionnal task output that will generate the `legend.txt` which will then be used when reconstructing the markdown content.
> 
> Side note:  
> Android Studio Otter (2025.2.1) is finally getting a Markdown renderer 🎉  
> Unfortunately, the rendering engine is forced to `Compose (experimental)` which means, it only parses basic stuff, and `<details>` are not rendered, as well as mermaid content, even with the proper plugin installed.  
> Hopefully they will support them soon.
> 
> image
>

PR #1964, opened 2025-09-20
Reduce recompositions on search screen

ivk1800:
**What I have done and why**

I refactored the SearchRoute → SearchScreen → SearchToolbar → SearchTextField chain to reduce unnecessary recompositions when the search query changes.

Before: entering a search query triggered recomposition of
SearchRoute, SearchScreen, SearchToolbar, SearchTextField, EmptySearchResultBody, RecentSearchesBody.

After: only SearchTextField and EmptySearchResultBody are recomposed.
SearchRoute, SearchScreen, and SearchToolbar stay stable.

This was achieved by:

1) Passing State<String> instead of raw String as a parameter down the composable chain.
2) Adapting previews accordingly.

The approach is similar to [this recomposition visualization](https://ivk1800.github.io/recomposition-visualization/?sampleId=7)(before) and [this one](https://ivk1800.github.io/recomposition-visualization/?sampleId=6) (after).
In the "before" example, intermediate functions get recomposed unnecessarily because they take a changing value as a parameter. In the "after" example, they take a stable lambda, so only the leaf node (RedBox) is recomposed.

Why:
1) Reduces the number of recompositions.
2) Improves performance by keeping intermediate composables stable.

I followed the official Compose best practice — “Defer reads as long as possible” (see [Android Performance Best Practices](https://developer.android.com/develop/ui/compose/performance/bestpractices#defer-reads)) — by deferring access to changing data until it’s needed, thus avoiding unnecessary recompositions.

### DIFF
--- a/feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
+++ b/feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
@@ -51,7 +51,9 @@ import androidx.compose.material3.TextField
 import androidx.compose.material3.TextFieldDefaults
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.State
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


hoangchungk53qx1:
Better yet, don't expect States in you components. Expect the final value as a parameter

hoangchungk53qx1:
Every time you read state in a Composable, you're creating a subscription. That Composable will recompose whenever that state changes. Read state too early in your composition tree, and you might be triggering unnecessary recompositions across large portions of your UI.

feature/search/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/search/SearchScreen.kt
```diff
@@ -126,7 +128,7 @@ internal fun SearchRoute(
 @Composable
 internal fun SearchScreen(
     modifier: Modifier = Modifier,
-    searchQuery: String = "",
+    searchQueryState: State<String> = remember { mutableStateOf("") },
```
hoangchungk53qx1, line 131: bad practice for passing State objects as parameter
Deffer state in compose by lambda

PR #1968, opened 2025-09-25
Disable Gradle parallel execution

edechamps-Google:
Since the move to Gradle 9 this seems to cause breakage in some setups, e.g. microsoft/vscode-gradle#1712.


liutikas:
vscode is not a supported workflow for nowinandroid. It is meant to be developed with Android Studio.

PR #1969, opened 2025-09-28
Make the logic that get BackendUrl simplify.

Jaehwa-Noh:
**What I have done and why**

Remove `map` and replace `Provider<Any>` to plain String.


liutikas:
How is this simplifying anything?

Jaehwa-Noh:
@liutikas This is a passing string type of url directly without any additional behavior, as currently using a map to get string.

Jaehwa-Noh:
Made a code can be read easily.

liutikas:
This is not any easier to read. I will keep the existing code.

PR #1985, opened 2025-11-06
Fix #1984 where TestMethodNameDetector does not scan some tests

### LINKED ISSUE #1984: [Bug]: TestMethodNameDetector does not scan all tests

- [x] I have searched the existing issues

- [x] I have searched StackOverflow

### What happened?

1. Go to `InterestListsDetailsScreenTest.kt`
2. Change one of the test names to include the prefix `test_`

Expected:
TestMethodNameDetector should issue a warning, both in the IDE  and in tasks like `gradle :app:lintAnalyzeDemoDebugUnitTest`
Actual:
No warning

### Relevant logcat output

```shell
The reason this is happening is because the annotations on the tests in `InterestListsDetailsScreenTest.kt` have `AnnotationUsageType.DEFINITION` which the `TestMethodNameDetector.kt` does not currently opt-in to scan for. Will submit a patch.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

drawers:
**What I have done and why**

Fix #1984 where `TestMethodNameDetector` does not scan some tests.

The reason this is happening is because the annotations in the unit tests are of usage type `AnnotationUsageType.DEFINTION` but the detector does not opt in to scan these.

I have based my solution on [`IgnoreWithoutReasonDetector`](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/IgnoreWithoutReasonDetector.kt;l=40) which is conceptually similar.

### DIFF
--- a/lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodNameDetector.kt
+++ b/lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodNameDetector.kt
@@ -18,6 +18,7 @@ package com.google.samples.apps.nowinandroid.lint
 
 import com.android.tools.lint.detector.api.AnnotationInfo
 import com.android.tools.lint.detector.api.AnnotationUsageInfo
+import com.android.tools.lint.detector.api.AnnotationUsageType
 import com.android.tools.lint.detector.api.Category.Companion.TESTING
 import com.android.tools.lint.detector.api.Detector
 import com.android.tools.lint.detector.api.Implementation
@@ -31,6 +32,7 @@ import com.android.tools.lint.detector.api.SourceCodeScanner
 import com.android.tools.lint.detector.api.TextFormat.RAW
 import com.intellij.psi.PsiMethod
 import org.jetbrains.uast.UElement
+import org.jetbrains.uast.UMethod
 import java.util.EnumSet
 import kotlin.io.path.Path
 
@@ -43,13 +45,18 @@ class TestMethodNameDetector : Detector(), SourceCodeScanner {
 
     override fun applicableAnnotations() = listOf("org.junit.Test")
 
+    // Restrict this detector to `AnnotationUsageType.DEFINITION` following
+    // similar examples in AOSP like `IgnoreWithoutReasonDetector`.
+    override fun isApplicableAnnotationUsage(type: AnnotationUsageType): Boolean =
+        type == AnnotationUsageType.DEFINITION
+
     override fun visitAnnotationUsage(
         context: JavaContext,
         element: UElement,
         annotationInfo: AnnotationInfo,
         usageInfo: AnnotationUsageInfo,
     ) {
-        val method = usageInfo.referenced as? PsiMethod ?: return
+        val method = annotationInfo.annotation.uastParent as? UMethod ?: return
 
         method.detectPrefix(context, usageInfo)
         method.detectFormat(context, usageInfo)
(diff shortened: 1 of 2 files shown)


lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodNameDetector.kt
```diff
@@ -43,13 +45,18 @@ class TestMethodNameDetector : Detector(), SourceCodeScanner {
 
     override fun applicableAnnotations() = listOf("org.junit.Test")
 
+    // Restrict this detector to `AnnotationUsageType.DEFINITION` following
+    // similar examples in AOSP like `IgnoreWithoutReasonDetector`.
+    override fun isApplicableAnnotationUsage(type: AnnotationUsageType): Boolean =
```
drawers, line 50: If we don't override this function from the superclass, we get the following implementation which skips `AnnotationUsageType.DEFINITION`:

```kotlin
  open fun isApplicableAnnotationUsage(type: AnnotationUsageType): Boolean {
    // Some annotation usages are off by default (typically because they were introduced
    // later and might introduce false positives in third party checks; checks should
    // opt into these.
    return when (type) {
      AnnotationUsageType.BINARY,
      AnnotationUsageType.EQUALITY,
      AnnotationUsageType.DEFINITION,
      AnnotationUsageType.IMPLICIT_CONSTRUCTOR,
      AnnotationUsageType.IMPLICIT_CONSTRUCTOR_CALL -> false
      else -> true
    }
  }
```

Hence the problem where test functions with an annotation that is not `AnnotationUsageType.METHOD_RETURN` aren't scanned.

Docs for `AnnotationUsageType.DEFINITION` suggest opting in like this:

```kotlin
override fun isApplicableAnnotationUsage(type: AnnotationUsageType): Boolean {
    return type == AnnotationUsageType.DEFINITION || super.isApplicableAnnotationUsage(type)
}
```

but if I do that, I get duplicate reports. In other words, the same problem is reported twice. So I have based the solution on [this one](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/IgnoreWithoutReasonDetector.kt;l=78) that restricts annotations to just `AnnotationUsageType.DEFINITION`

lint/src/main/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodNameDetector.kt
```diff
@@ -43,13 +45,18 @@ class TestMethodNameDetector : Detector(), SourceCodeScanner {
...
+        type == AnnotationUsageType.DEFINITION
+
     override fun visitAnnotationUsage(
         context: JavaContext,
         element: UElement,
         annotationInfo: AnnotationInfo,
         usageInfo: AnnotationUsageInfo,
     ) {
-        val method = usageInfo.referenced as? PsiMethod ?: return
+        val method = annotationInfo.annotation.uastParent as? UMethod ?: return
```
drawers, line 59: I had to change this line because for some annotations `usageInfo.referenced` will return `null`.

I based the new interpretation on similar logic in `IgnoreWithoutReasonDetector`. One doubt would be what happens when there are multiple annotations. If I do `context.uastFile.asRecursiveLogString()` to look at the UAT nodes, these are still siblings within a single parent where the parent is the `UMethod`.

```text
UMethod (name = test_foo2)
    UAnnotation (fqName = org.junit.Test)
    UAnnotation (fqName = kotlin.Deprecated)
    UBlockExpression
```

lint/src/test/kotlin/com/google/samples/apps/nowinandroid/lint/TestMethodNameDetectorTest.kt
```diff
@@ -40,6 +40,11 @@ class TestMethodNameDetectorTest {
                         fun test_foo() = Unit
                         @Test
                         fun `test foo`() = Unit
+
+                        @Test
+                        fun test_foo2() {
```
drawers, line 45: The previous version of the TestMethodNameDetector would not have reported on this method. In other words, a false negative.

PR #1999, opened 2025-11-24
Bump number of core on CI emulator

Jaehwa-Noh:
**What I have done and why**

This PR follows up #1816.

The standard public repository runner has 4 cores, so that I use that all cores as running emulator on github action.

Number of core documentation.
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

image

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -241,6 +241,8 @@ jobs:
           api-level: ${{ matrix.api-level }}
           arch: x86_64
           disable-animations: true
+          # https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories
+          # Number of core of default public repository github action runner is 4 in all.
           cores: 4
           disk-size: 6000M
           heap-size: 600M


.github/workflows/Build.yaml
```diff
@@ -241,6 +241,9 @@ jobs:
           api-level: ${{ matrix.api-level }}
           arch: x86_64
           disable-animations: true
+          # https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories
+          # Number of core of default public repository github action runner is 4 in all.
+          cores: 4
```
liutikas, line 246: Do I understand correctly that this is giving 4 cores to emulator, but we'll still be running Gradle at the same time also using up cores?

Jaehwa-Noh: No, this PR just gives 4 cores to emulator.
You can check CI Gradle workers at here
https://github.com/android/nowinandroid/blob/ade8065442099e1f63f6c4c716bafc84b053ca09/.github/ci-gradle.properties#L19


| | Emulator | Gradle|
|-|----------|-------|
|PR #1999| 4 cores | 2 cores |
|PR #1816 | 2 cores| 4 cores|

Is there any problems when set the cores as below cases?
Case1: 4 and 2 cores (as #1999 merged)
Case2: 2 and 4 cores (as #1816 merged)
Case3: 4 and 4 cores (as all #1999 #1816 PR merged)

PR #2002, opened 2025-11-25
[WIP] Enable project isolation

liutikas:
Fixes https://github.com/android/nowinandroid/issues/1842

### DIFF
--- a/build-logic/convention/src/main/kotlin/RootPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/RootPlugin.kt
@@ -17,10 +17,16 @@
 import com.google.samples.apps.nowinandroid.configureGraphTasks
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.configuration.BuildFeatures
+import javax.inject.Inject
 
-class RootPlugin : Plugin<Project> {
+abstract class RootPlugin : Plugin<Project> {
+    @get:Inject abstract val buildFeatures: BuildFeatures
+
     override fun apply(target: Project) {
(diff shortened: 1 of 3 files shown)


settings.gradle.kts
```diff
@@ -26,6 +26,7 @@ pluginManagement {
         }
         mavenCentral()
         gradlePluginPortal()
+        maven("https://central.sonatype.com/repository/maven-snapshots/")
```
dturner: How much does this affect the build time? When adding the snapshot repo for AndroidX it massively slows down the gradle sync process which degrades the developer experience a lot.

liutikas: Sorry forgot to mark it WIP. This is not yet ready

liutikas: Removed.

build-logic/convention/src/main/kotlin/RootPlugin.kt
```diff
@@ -17,10 +17,16 @@
...
+import javax.inject.Inject
+
+abstract class RootPlugin : Plugin<Project> {
+    @get:Inject abstract val buildFeatures: BuildFeatures
 
-class RootPlugin : Plugin<Project> {
     override fun apply(target: Project) {
         require(target.path == ":")
-        target.subprojects { configureGraphTasks() }
+        if (!buildFeatures.isolatedProjects.active.orElse(false).get()) {
```
Emmanu12, line 28:
```suggestion
        if (!buildFeatures.isolatedProjects.active.orElse(false).get()) {
```

Emmanu12:
Si

PR #2012, opened 2025-12-05
Fix splash screen theme not respecting user preference on Android 12+

### LINKED ISSUE #633: [Bug]: Splashscreen theme doesn't change after switching the color theme

300Ech:
## Description
Fixes #633

The splash screen was displaying the system theme instead of respecting the user's saved app theme preference on Android 12+ devices.

## Problem
On Android 12+, the system shows the splash screen before Application.onCreate() finishes. The app saved the theme setting, but UiModeManager wasn’t updated, so the splash screen ignored the user's theme.

## Solution
- Initialize night mode in `NiaApplication.onCreate()` using `runBlocking` to ensure it completes before the splash screen is displayed
- Observe theme preference changes to immediately notify when the user changes the theme, ensuring subsequent cold starts use the correct theme
- Use `UiModeManager.setApplicationNightMode()` on API 31+, fall back to `AppCompatDelegate` on older Android versions
- Add extension functions to UiExtensions.kt

## Why the Observer is Needed
After a theme change, the first cold start shows the wrong splash screen since UiModeManager isn’t updated right away. Watching for changes fixes that instantly.

## Testing
- ✅ Set device to Light mode, app to Dark theme → Splash screen shows dark
- ✅ Set device to Dark mode, app to Light theme → Splash screen shows light  
- ✅ Set app to Follow System → Splash screen matches system theme
- ✅ Tested on Android 16 (API 36) and Android 11 (API 30) devices
- ✅ All existing tests pass

## Checklist
- [x] Code follows the project's coding style
- [x] Changes are backwards compatible (API 21+)
- [x] Tested on multiple Android versions
- [x] References issue #633
- [x] Solution follows existing project patterns (@ApplicationScope, extension functions)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -99,6 +99,7 @@ dependencies {
     implementation(libs.androidx.compose.runtime.tracing)
     implementation(libs.androidx.core.ktx)
     implementation(libs.androidx.core.splashscreen)
+    implementation(libs.androidx.appcompat)
     implementation(libs.androidx.lifecycle.runtimeCompose)
     implementation(libs.androidx.lifecycle.viewModel.navigation3)
     implementation(libs.androidx.profileinstaller)
--- a/app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
+++ b/app/src/main/kotlin/com/google/samples/apps/nowinandroid/NiaApplication.kt
@@ -20,11 +20,17 @@ import android.app.Application
 import android.content.pm.ApplicationInfo
 import android.os.StrictMode
 import android.os.StrictMode.ThreadPolicy.Builder
+import androidx.datastore.core.DataStore
 import coil.ImageLoader
 import coil.ImageLoaderFactory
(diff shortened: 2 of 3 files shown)


seve-andre:
For reference, I implemented and recently updated a similar non-blocking solution (see #1718).
Hopefully one of these approaches can be merged soon.

300Ech:
hi! thanks for pointing that out @seve-andre, appreciate it!

Happy to leave it to the maintainers to decide what fits best, or if both ideas can be combined. Either way, glad to see this getting addressed!

PR #2019, opened 2025-12-10
Replace `extensions.getByType<T>()` with shorthand notation `the<T>()`

SimonMarquis:
docs: https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.kotlin.dsl/the.html


liutikas:
I don't agree this is better, this is way harder to read.

PR #2020, opened 2025-12-10
Replace deprecated `TabRow` with `SecondaryTabRow`

SimonMarquis:
No visual change is expected with this migration.
Screenshots in [core/designsystem/src/test/screenshots/Tabs](https://github.com/android/nowinandroid/blob/main/core/designsystem/src/test/screenshots/Tabs) are unchanged.

(Tabs are currently not used in the main app, only in the catalog)


kamaalyasin230-pixel:
ethio coder

PR #2034, opened 2026-01-14
feat: updates DevicePreviews.kt to use Devices constants

telpirion:
**What I have done and why**

The `@DevicePreview` annotation shows four previews of phone / portrait orientation. I believe that this is because the `device` field is specified as a string (which should work ... but doesn't). This PR updates the `@DevicePreview` annotation so that it uses the constants in the `androidx.compose.ui.tooling.preview.Devices` object.


telpirion:
It seems like I must've based my changes on an old branch :/. I'm going to close and reopen a new PR on a fresh branch.

PR #2036, opened 2026-01-15
Fix android#1982: Improved ConnectivityManagerNetworkMonitor

### LINKED ISSUE #1982: [FR]: Add NET_CAPABILITY_VALIDATED, Transport Type Check , distinctUntilChanged() to the Flow in ConnectivityManagerNetworkMonitor.kt

- [x] I have searched the existing issues

### 1. Add `NET_CAPABILITY_VALIDATED`

Add a check for `NET_CAPABILITY_VALIDATED` for a more reliable internet connection check on newer Android versions. A network can be "connected" but not actually have internet access (e.g., captive portals).

#### Reasoning:

- `NET_CAPABILITY_INTERNET` indicates that the network is intended to provide internet access.
- `NET_CAPABILITY_VALIDATED` confirms that the device has successfully connected to the internet through this network. Using both gives a more accurate signal of a usable internet connection.

---

### 2. Add `distinctUntilChanged()` to the Flow

The flow might emit the same value consecutively (e.g., multiple networks becoming available one after another, all causing `true` to be sent). While `conflate()` helps by dropping intermediate values, `distinctUntilChanged()` will prevent collectors from being notified if the connectivity status hasn't actually changed.

#### Reasoning:

This is a small optimization that ensures downstream collectors only react when the online status actually flips from `true` to `false` or vice-versa, reducing unnecessary recompositions or data processing.

---

### 3. Add Transport Type Check

Not all validated networks are the ones we want for general internet access. A device might have a "validated" peer-to-peer Wi-Fi network or another special-purpose network that doesn't provide the broad internet access your app expects. To improve this, we can refine our `isCurrentlyConnected` function to also check for common internet-providing transport types.

#### Reasoning:

Explicitly checking for `TRANSPORT_WIFI`, `TRANSPORT_CELLULAR`, or `TRANSPORT_ETHERNET` ensures we are connected to a standard internet uplink, filtering out misleading signals from other specialized networks. This makes our connectivity check more accurate

#### Suggestion:

Update the `isCurrentlyConnected` function to also check for NET_CAPABILITY_VALIDATED.

``` kotlin
@SuppressLint("ObsoleteSdkInt")
    @Suppress("DEPRECATION")
    private fun ConnectivityManager.isCurrentlyConnected(): Boolean {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
            val network = activeNetwork ?: return false
            val capabilities = getNetworkCapabilities(network) ?: return false
            return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
                    && capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) // Add this line
        } else {
            // For older versions
            return activeNetworkInfo?.isConnected ?: false
        }
    }
```

And similarly, update the `NetworkRequest.Builder`:

``` kotlin
// Inside the callbackFlow block
val request = Builder()
    .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
    .addCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) // Add this line
    .build()
connectivityManager.registerNetworkCallback(request, callback)
```

---

#### Suggestion:

Chain `distinctUntilChanged()` to the flow.

``` kotlin
// In ConnectivityManagerNetworkMonitor.kt
import kotlinx.coroutines.flow.distinctUntilChanged
// ... other imports

override val isOnline: Flow<Boolean> = callbackFlow {
    // ... same implementation
}
    .flowOn(ioDispatcher)
    .conflate()
    .distinctUntilChanged() // Add this line
```

---

#### Suggestion:

``` kotlin
@SuppressLint("ObsoleteSdkInt")
    @Suppress("DEPRECATION")
    private fun ConnectivityManager.isCurrentlyConnected(): Boolean {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
            val network = activeNetwork ?: return false
            val capabilities = getNetworkCapabilities(network) ?: return false
            return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
                    && capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
                     // Add this check for transport types
                    && (capabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) ||
                    capabilities.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR) ||
                    capabilities.hasTransport(NetworkCapabilities.TRANSPORT_ETHERNET))
        } else {
            // For older versions
            return activeNetworkInfo?.isConnected ?: false
        }
    }
```

---

## Final Code after suggestions:

``` kotlin
// File: nowinandroid/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt

// Github file link: https://github.com/android/nowinandroid/blob/705fd9068bb2225cb351a0c10b2154e0e56f40de/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt

internal class ConnectivityManagerNetworkMonitor @Inject constructor(
    @field:ApplicationContext private val context: Context,
    @field:Dispatcher(AppDispatchers.IO) private val ioDispatcher: CoroutineDispatcher,
) : NetworkMonitor {
    override val isOnline: Flow<Boolean> = callbackFlow {
        trace("NetworkMonitor.callbackFlow") {
            val connectivityManager = context.getSystemService<ConnectivityManager>()
            if (connectivityManager == null) {
                channel.trySend(false)
                channel.close()
                return@callbackFlow
            }

            /**
             * The callback's methods are invoked on changes to *any* network matching the [NetworkRequest],
             * not just the active network. So we can simply track the presence (or absence) of such [Network].
             */
            val callback = object : NetworkCallback() {

                private val networks = mutableSetOf<Network>()

                override fun onAvailable(network: Network) {
                    networks += network
                    channel.trySend(true)
                }

                override fun onLost(network: Network) {
                    networks -= network
                    channel.trySend(networks.isNotEmpty())
                }
            }

            trace("NetworkMonitor.registerNetworkCallback") {
                val request = Builder()
                    .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
                    .addCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
                    .build()
                connectivityManager.registerNetworkCallback(request, callback)
            }

            /**
             * Sends the latest connectivity status to the underlying channel.
             */
            channel.trySend(connectivityManager.isCurrentlyConnected())

            awaitClose {
                connectivityManager.unregisterNetworkCallback(callback)
            }
        }
    }
        .flowOn(ioDispatcher)
        .conflate()
        .distinctUntilChanged()

    @SuppressLint("ObsoleteSdkInt")
    @Suppress("DEPRECATION")
    private fun ConnectivityManager.isCurrentlyConnected(): Boolean {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
            val network = activeNetwork ?: return false
            val capabilities = getNetworkCapabilities(network) ?: return false
            return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
                    && capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
                    && (capabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) ||
                    capabilities.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR) ||
                    capabilities.hasTransport(NetworkCapabilities.TRANSPORT_ETHERNET))
        } else {
            // For older versions
            return activeNetworkInfo?.isConnected ?: false
        }
    }
}
```

#### Additional context

1. For Add `NET_CAPABILITY_VALIDATED` : 

- [https://developer.android.com/develop/connectivity/network-ops/reading-network-state](https://developer.android.com/develop/connectivity/network-ops/reading-network-state)

- [https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_VALIDATED](https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_VALIDATED)

2. For Add Transport Type Check:

- [https://medium.com/@doronkakuli/adapting-your-network-connectivity-checks-for-android-15-a-practical-guide-2b1850619294](https://medium.com/@doronkakuli/adapting-your-network-connectivity-checks-for-android-15-a-practical-guide-2b1850619294)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Syedovaiss:
## What I have done and why

**Fix #1982: Align NetworkRequest and improve transport check readability**

### Changes

1. **Added NET_CAPABILITY_VALIDATED** and transport type checks to `ConnectivityManagerNetworkMonitor`:
   - Ensures the network is actually validated and provides real internet connectivity.
   - Only supports Wi-Fi, Cellular, and Ethernet transports to avoid false positives from other transports like VPN or peer-to-peer networks.

2. **Updated NetworkRequest in registerNetworkCallback**:
   - Constrained to the same transports as `isCurrentlyConnected()` for consistency.
   - Prevents inconsistent flow emissions when network types mismatch.

3. **Refactored transport check logic**:
   - Created a `hasSupportedTransport()` helper.
   - Uses a set and `any()` for cleaner, more maintainable code.

4. **Added `distinctUntilChanged()` to the flow**:
   - Prevents duplicate emissions when connectivity state hasn't actually changed.
   - Reduces unnecessary recompositions or downstream processing.

### Motivation

- Previously, `isCurrentlyConnected()` and `NetworkRequest` used in the callback could disagree on connectivity state.
- This could lead to emitting `true` for networks that are not actually usable for internet access.
- Improving readability and maintainability helps future developers adjust supported transport types easily.

### Additional Context

- See [Android NET_CAPABILITY_VALIDATED documentation](https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_VALIDATED)
- Inspired by recommendations for reliable connectivity checks on Android 15+ ([Medium Guide](https://medium.com/@doronkakuli/adapting-your-network-connectivity-checks-for-android-15-a-practical-guide-2b1850619294))

### DIFF
--- a/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
+++ b/core/data/src/main/kotlin/com/google/samples/apps/nowinandroid/core/data/util/ConnectivityManagerNetworkMonitor.kt
@@ -22,7 +22,6 @@ import android.net.ConnectivityManager.NetworkCallback
 import android.net.Network
 import android.net.NetworkCapabilities
 import android.net.NetworkRequest
-import android.net.NetworkRequest.Builder
 import androidx.core.content.getSystemService
 import androidx.tracing.trace
 import com.google.samples.apps.nowinandroid.core.network.Dispatcher
@@ -33,12 +32,13 @@ import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.flow.conflate
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.flowOn
 import javax.inject.Inject
 
 internal class ConnectivityManagerNetworkMonitor @Inject constructor(
-    @ApplicationContext private val context: Context,
-    @Dispatcher(IO) private val ioDispatcher: CoroutineDispatcher,
+    @param:ApplicationContext private val context: Context,
+    @param:Dispatcher(IO) private val ioDispatcher: CoroutineDispatcher,
 ) : NetworkMonitor {
     override val isOnline: Flow<Boolean> = callbackFlow {
         trace("NetworkMonitor.callbackFlow") {
@@ -69,8 +69,12 @@ internal class ConnectivityManagerNetworkMonitor @Inject constructor(
             }
 
             trace("NetworkMonitor.registerNetworkCallback") {
-                val request = Builder()
+                val request = NetworkRequest.Builder()
                     .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+                    .addCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
+                    .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+                    .addTransportType(NetworkCapabilities.TRANSPORT_CELLULAR)
+                    .addTransportType(NetworkCapabilities.TRANSPORT_ETHERNET)
                     .build()
                 connectivityManager.registerNetworkCallback(request, callback)
             }
@@ -87,9 +91,31 @@ internal class ConnectivityManagerNetworkMonitor @Inject constructor(
     }
         .flowOn(ioDispatcher)
         .conflate()
+        .distinctUntilChanged()
 
     private fun ConnectivityManager.isCurrentlyConnected(): Boolean {
-        val networkCapabilities = getNetworkCapabilities(activeNetwork) ?: return false
-        return networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+        val network = activeNetwork ?: return false
+        val capabilities = getNetworkCapabilities(network) ?: return false
+
+        return hasInternetCapability(capabilities) &&
+            isValidated(capabilities) &&
+            hasSupportedTransport(capabilities)
+    }
+
+    private fun hasInternetCapability(capabilities: NetworkCapabilities): Boolean {
+        return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+    }
+
+    private fun isValidated(capabilities: NetworkCapabilities): Boolean {
+        return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
+    }
+
+    private fun hasSupportedTransport(capabilities: NetworkCapabilities): Boolean {
+        val supportedTransports = setOf(
+            NetworkCapabilities.TRANSPORT_WIFI,
+            NetworkCapabilities.TRANSPORT_CELLULAR,
+            NetworkCapabilities.TRANSPORT_ETHERNET,
+        )
+        return supportedTransports.any(capabilities::hasTransport)
     }
 }


Syedovaiss:
@dturner  would appreciate your review when available.

Key intent here was to ensure:
- validated internet connectivity only
- consistent logic between NetworkRequest and isCurrentlyConnected()
- reduced duplicate flow emissions

Open to any feedback or alternative approaches.

anhtuannd:
If we only need to validate for Internet connection, hasSupportedTransport() check seems to be redundant.

Syedovaiss:
> If we only need to validate for Internet connection, hasSupportedTransport() check seems to be redundant.

Thanks for the feedback — good point.

NET_CAPABILITY_VALIDATED does ensure that the network has verified internet access, but the additional hasSupportedTransport() check was added intentionally to align with the scope of connectivity this app considers “online.”

In practice, a network can be validated but still be backed by transports such as VPN or other specialized links. While these may technically provide internet access, they can produce connectivity signals that differ from what the app expects for general online behavior.

The transport filter ensures:
	•	Consistency between isCurrentlyConnected() and the NetworkRequest
	•	Avoidance of false positives from validated but non-standard transports
	•	Clear, explicit definition of supported uplink types (Wi-Fi, Cellular, Ethernet)

That said, I’m happy to simplify this if the project prefers to rely solely on NET_CAPABILITY_VALIDATED and treat all validated transports as acceptable. Please let me know which direction you’d prefer, and I can update the implementation accordingly.

PR #2038, opened 2026-01-15
feat: updates `@DevicePreviews` to use the Devices object constants

telpirion:
**What I have done and why**

- The `@DevicePreviews` annotation didn't show different form factors; they were always portrait. I changed the annotation so that it uses the constants from the [`Devices`](https://developer.android.com/reference/kotlin/androidx/compose/ui/tooling/preview/Devices#device()) object.

This is the existing behavior for previews (from InterestsScreen): 

image

### DIFF
--- a/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/DevicePreviews.kt
+++ b/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/DevicePreviews.kt
@@ -16,14 +16,15 @@
 
 package com.google.samples.apps.nowinandroid.core.ui
 
+import androidx.compose.ui.tooling.preview.Devices
 import androidx.compose.ui.tooling.preview.Preview
 
 /**
  * Multipreview annotation that represents various device sizes. Add this annotation to a composable
  * to render various devices.
  */
-@Preview(name = "phone", device = "spec:shape=Normal,width=360,height=640,unit=dp,dpi=480")
-@Preview(name = "landscape", device = "spec:shape=Normal,width=640,height=360,unit=dp,dpi=480")
-@Preview(name = "foldable", device = "spec:shape=Normal,width=673,height=841,unit=dp,dpi=480")
-@Preview(name = "tablet", device = "spec:shape=Normal,width=1280,height=800,unit=dp,dpi=480")
+@Preview(name = "phone", device = Devices.PHONE, showBackground = true)
+@Preview(name = "foldable", device = Devices.FOLDABLE, showBackground = true)
+@Preview(name = "tablet", device = Devices.TABLET, showBackground = true)
+@Preview(name = "desktop", device = Devices.DESKTOP, showBackground = true)
 annotation class DevicePreviews
--- a/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/DevicePreviews.kt
+++ b/core/ui/src/main/kotlin/com/google/samples/apps/nowinandroid/core/ui/DevicePreviews.kt
@@ -24,6 +24,7 @@ import androidx.compose.ui.tooling.preview.Preview
  * to render various devices.
  */
 @Preview(name = "phone", device = Devices.PHONE, showBackground = true)
+@Preview(name = "phone_in_landscape", widthDp = 891, heightDp = 411, showBackground = true)
 @Preview(name = "foldable", device = Devices.FOLDABLE, showBackground = true)
 @Preview(name = "tablet", device = Devices.TABLET, showBackground = true)
 @Preview(name = "desktop", device = Devices.DESKTOP, showBackground = true)


telpirion (commented):
Bad Code Assist, no donut

app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/NiaApp.kt
```diff
@@ -205,6 +218,9 @@ internal fun NiaApp(
                     ),
                 )
             },
+            // Fix for render issue: Scaffold crashes with "List is empty" when contentWindowInsets is set
```
dturner: Please add a link to the issue so we can remove this workaround when it's fixed.

app/src/testDemo/screenshots/compactWidth_compactHeight_showsNavigationBar.png
dturner, line 1: The news cards are an important part of the screenshots because they improve test fidelity - is there a reason why they've been removed from all the screenshots?

core/datastore-proto/build.gradle.kts
```diff
@@ -42,14 +42,6 @@ protobuf {
     }
 }
 
-androidComponents.beforeVariants {
```
dturner: Why has this been removed? Usually if it's not directly related to the PR it should go in a separate PR.

feature/foryou/impl/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/impl/navigation/ForYouEntryProvider.kt
```diff
@@ -16,17 +16,35 @@
...
 import com.google.samples.apps.nowinandroid.feature.foryou.impl.ForYouScreen
+import com.google.samples.apps.nowinandroid.feature.foryou.impl.OnboardingUiState
 import com.google.samples.apps.nowinandroid.feature.topic.api.navigation.navigateToTopic
 
 fun EntryProviderScope<NavKey>.forYouEntry(navigator: Navigator) {
     entry<ForYouNavKey> {
-        ForYouScreen(
-            onTopicClick = navigator::navigateToTopic,
-        )
+        if (LocalInspectionMode.current) {
```
dturner: This pollutes the production code with code that is only used for previews. Are there any alternatives to doing this? Maybe have a different `entryProvider` that is used for previews instead?

gradle/libs.versions.toml
```diff
@@ -106,6 +106,7 @@ androidx-metrics = { group = "androidx.metrics", name = "metrics-performance", v
 androidx-navigation-testing = { group = "androidx.navigation", name = "navigation-testing", version.ref = "androidxNavigation" }
 androidx-navigation3-runtime = { group = "androidx.navigation3", name = "navigation3-runtime", version.ref = "androidxNavigation3" }
 androidx-navigation3-ui = { group = "androidx.navigation3", name = "navigation3-ui", version.ref = "androidxNavigation3" }
+androidx-navigationevent-compose = { group = "androidx.navigationevent", name = "navigationevent-compose", version = "1.0.1" }
```
dturner: Please follow the convention for all other dependencies by moving the version number into a variable at the top of the file.

feature/foryou/impl/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/impl/ForYouScreen.kt
```diff
@@ -618,3 +618,24 @@ fun ForYouScreenPopulatedAndLoading(
...
 }
+
+@DevicePreviews
+@Composable
+fun ForYouScreenNotPopulatedOnboarding() {
+    NiaTheme {
+        ForYouScreen(
+            isSyncing = true,
+            onboardingUiState = OnboardingUiState.Shown(
+                topics = emptyList(),
```
dturner: If there is a justification for adding this preview, using `topics = emptyList()` here means this is not a realistic preview of the `ForYouScreen` when the onboarding UI is shown. There's code from the other previews that can be copied here to load the topics.

feature/foryou/impl/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/impl/ForYouScreen.kt
```diff
@@ -618,3 +618,24 @@ fun ForYouScreenPopulatedAndLoading(
         )
     }
 }
+
+@DevicePreviews
+@Composable
+fun ForYouScreenNotPopulatedOnboarding() {
```
dturner: Why add this? While it is a valid state, it's very unlikely to ever happen in the app. Both topics and news resources load at the same time and the other previews cover the Loading and Shown states.

dturner (changes requested):
Thanks for the PR. Made a few suggestions and have a few questions.

app/src/testDemo/screenshots/compactWidth_compactHeight_showsNavigationBar.png
telpirion: That change wasn't intentional, I assure you. Perhaps one of the GitHub bots made this change? I'll look into it.

core/datastore-proto/build.gradle.kts
telpirion: This change was made by Gemini. I believe it was done to address a build error. I'll see if I can revert this without breaking the PR.

feature/foryou/impl/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/impl/ForYouScreen.kt
telpirion: This state happened to me when I ran the app.

feature/foryou/impl/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/impl/ForYouScreen.kt
telpirion: As mentioned previously, it's realistic in the sense that it's a state I encountered. Specifically when running on a tablet.

feature/foryou/impl/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/impl/ForYouScreen.kt
dturner: That's unexpected and good to know, thanks. Let's keep it then.

feature/foryou/impl/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/impl/ForYouScreen.kt
dturner: You encountered a state where the onboarding was shown but there were no topics in it? If so, that's a bug, and we shouldn't provide previews for buggy states. 

The correct state to preview should be that the onboarding has topics in it.

core/datastore-proto/build.gradle.kts
dturner: This change has already been made on `main` in https://github.com/android/nowinandroid/pull/2051. You can revert the changes made to this file and rebase.

app/src/testDemo/screenshots/compactWidth_compactHeight_showsNavigationBar.png
dturner: No worries. There is a [step in the build workflow](https://github.com/android/nowinandroid/blob/main/.github/workflows/Build.yaml#L122) that generates new screenshots based on the changes in your PR. This is so that you, the PR owner, can visually verify that the changes you've made have the desired effect, or allows you to spot when you've changed something that inadvertently changes the UI. 

In this case, it sounds like the latter so you need to fix the news resources being displayed. The current screenshots will be removed from the PR after you push the code changes and the build workflow completes.

telpirion:
I removed all of the extra changes as they weren't pertinent to the primary purpose of this PR. Please take another look.

dturner:
@telpirion Please could you help me understand the form factor previews a bit better?

> The @DevicePreviews annotation didn't show different form factors; 

Original code: 

```
@Preview(name = "phone", device = "spec:shape=Normal,width=360,height=640,unit=dp,dpi=480")
@Preview(name = "landscape", device = "spec:shape=Normal,width=640,height=360,unit=dp,dpi=480")
@Preview(name = "foldable", device = "spec:shape=Normal,width=673,height=841,unit=dp,dpi=480")
@Preview(name = "tablet", device = "spec:shape=Normal,width=1280,height=800,unit=dp,dpi=480")
annotation class DevicePreviews
```

Genuine question, aren't these different form factors? Is there something to do with the density being the same that makes them lack form factor diversity? 

> they were always portrait. 

The "landscape" and "tablet" specs both have width > height which makes them landscape - am I missing something? Or did you just mean that the desktop form factor wasn't present? 

Did you intend to remove the "landscape" preview? Should we have a preview for the "phone in landscape" form factor? 

Sorry for all the questions. Screenshot testing and previews for adaptive apps is a hot topic at the moment and I'd like to make sure I understand the space correctly and that we're providing best practices here.

One final thought, should we update the `DefaultTestDevices` (used for screenshot tests) to match the device previews?

dturner (approved):
Thanks for doing this.

telpirion:
🎆

PR #2041, opened 2026-01-16
Update README files that contained old svg references

SimonMarquis:


dturner (approved):
Thanks @SimonMarquis ❤️

PR #2043, opened 2026-01-16
Implement settings feature as a Nav3 dialog screen

SimonMarquis:
This makes the `:feature:settings` behave like any other feature with api/impl split.


core/navigation/src/main/kotlin/com/google/samples/apps/nowinandroid/core/navigation/NavigationState.kt
```diff
@@ -75,6 +75,11 @@ class NavigationState(
 
     @get:VisibleForTesting
     val currentKey: NavKey by derivedStateOf { currentSubStack.last() }
+
+    /**
+     * Current [NavKey] but ignores [DialogNavKey]s that could be added on top.
+     */
+    val currentKeyIgnoringDialogs: NavKey by derivedStateOf { currentSubStack.last { it !is DialogNavKey } }
```
SimonMarquis, line 82: @dturner I'm open to suggestions on how to properly support this.

PR #2054, opened 2026-01-27
Finalize bump agp 9.0 (enable newDsl)

Jaehwa-Noh:
**What I have done and why**

Finally enable newDsl that bump AGP 9.0 totally.

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -54,7 +54,7 @@ graph TB
     :core:data[data]:::android-library
     :core:database[database]:::android-library
     :core:datastore[datastore]:::android-library
-    :core:datastore-proto[datastore-proto]:::android-library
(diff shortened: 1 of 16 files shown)


Jaehwa-Noh:
@liutikas PTAL

liutikas:
This is an ugly workaround. Instead we just need to wait for protobuf plugin to get fixed.

Jaehwa-Noh:
@liutikas Is there any reason datastore-protobuf to be Android library?

Does protobuf need android dependency?

liutikas:
> @liutikas Is there any reason datastore-protobuf to be Android library?

After another pass, you are right, this can just be JVM.

Jaehwa-Noh:
@liutikas Thank you to review it again.

PR #2062, opened 2026-02-11
Use `org.jetbrains.kotlin.jvm` instead of `kotlin-dsl`

martinbonnin:
Opening this as reference for anyone curious what's going on under the `` `kotlin-dsl` `` hood. 

Feel free to merge or discard this PR. 

For more context, you can also read [this blog post](https://mbonnin.net/2025-07-10_the_case_against_kotlin_dsl/).

The immediate benefits in this repo aren't huge. The main one is to be able to use [newer tooling without having to wait](https://blog.alllex.me/posts/2023-11-03-liberal-library-tooling/) for Gradle to integrate newer versions of Kotlin. As of today, this effectively allows to use KGP `2.3` in `build-logic` instead of KGP `2.2`. 

Also, since [precompiled scripts plugins are not used anymore](https://github.com/android/nowinandroid/pull/38), this removes that code from build logic and makes it a tiny bit slimmer (I didn't notice any real performance improvement).

If there is interest, I can also make a follow up PR to make build-logic even slimmer and removes `sam-with-receiver`, `assignment` and `gradleKotlinDsl()`. That will also make the Kotlin code in `build-logic` closer to everyday Kotlin.

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -14,26 +14,29 @@
  * limitations under the License.
  */
 
-import org.jetbrains.kotlin.gradle.dsl.JvmTarget
-
 plugins {
-    `kotlin-dsl`
+    alias(libs.plugins.kotlin.jvm)
     alias(libs.plugins.android.lint)
+    alias(libs.plugins.tapmoc)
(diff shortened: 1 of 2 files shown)


liutikas:
It doesn't seem super useful here, so i'm not sure if this is work adding.

PR #2100, opened 2026-04-19
Update README graphs with correct datastore-proto plugin

FletchMcKee:
**What I have done and why**

Ran `./gradlew graphUpdate` so that [Check Graphs](https://github.com/android/nowinandroid/actions/runs/24601257469/job/71940220590) can pass. The datastore-proto module uses jvm-library but is referenced as an android-library in two of the READMEs.


dturner (approved):
Big thanks for doing this. The build has been broken because of this for a while and I haven't had a chance to look into it.

PR #2138, opened 2026-08-31
Fix baseline profile generation

liutikas:
Before this change baseline profiles where not being generated at all because
android { buildTypes { release {
            baselineProfile.automaticGenerationDuringBuild = true
}}}

is overwritten by
baselineProfile {
    // Don't build on every iteration of a full assemble.
    // Instead enable generation directly for the release build variant.
    automaticGenerationDuringBuild = false
}

Instead, correctly set it for the release variant

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -53,8 +53,6 @@ android {
             // who clones the code to sign and run the release variant, use the debug signing key.
             // TODO: Abstract the signing configuration to a separate file to avoid hardcoding this.
             signingConfig = signingConfigs.named("debug").get()
-            // Ensure Baseline Profile is fresh for release builds.
-            baselineProfile.automaticGenerationDuringBuild = true
         }
     }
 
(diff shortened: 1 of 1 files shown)


liutikas:
Note, this makes it runs, but it seems that GitHub actions fail as we cannot run the emulator. @keyboardsurfer do you have an idea what needs to change to fix that?

liutikas:
> Note, this makes it runs, but it seems that GitHub actions fail as we cannot run the emulator. @keyboardsurfer do you have an idea what needs to change to fix that?

```
Execution failed for task ':benchmarks:pixel6Api33Setup'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.ManagedDeviceInstrumentationTestSetupTask$ManagedDeviceSetupRunnable
   > com.android.build.gradle.internal.AvdSnapshotHandler$EmulatorStartFailedException: Unable to start Android emulator for dev33_default_x86_64_Pixel_6. Error message from emulator process = [ERROR        | gpuChoiceBasedOnGpuOptions: Selected GPU option 'auto-no-window' is not valid, switching to 'auto' mode.
gradle/actions: Writing build results to /home/runner/work/_temp/.gradle-actions/build-results/__run_6-1788208518773.json
     ERROR        | x86_64 emulation currently requires hardware acceleration!]
```

